PR #7, opened 2016-06-10
Transactions from the future are not included in blocks

asolovyov:


Tolsi:
Is this relevant?

asolovyov:
Please check with Sasha - initially it was his req. And apart from this, another reason to add such a restriction is the current implementation of tx uniqueness check - if you accept a tx from a far future you will have to wait that moment to send next tx.

Tolsi:
@sasha35625 @catena2w

bcdev-:
"wait that moment" - but that applies only to SimplePayment transactions, and these are deprecated.

Tolsi:
see wavesplatform/Scorex#179

PR #54, opened 2016-10-07
Updated scrypto to 1.2.0-RC2.

bcdev-:
A prerequisite for p2p communication encryption.
Please also accept https://github.com/wavesplatform/Scorex/pull/129 since they are dependent upon each other!


bcdev-:
It says that compilation failed. _This is incorrect!_ These two commits are simply dependent upon each other, that's why compilation fails. I cannot make a pull request for two repos at the same time.

So please also accept https://github.com/wavesplatform/Scorex/pull/129 since they are dependent upon each other.

PR #61, opened 2016-10-17
Simple unencrypted message transaction.

bcdev-:
Just a Simple Message transaction, pretty similar to NXT.
This is only the core protocol part [the one that requires a hard fork], the API is very basic will be refined later.


bcdev-:
I'll do the complete implementation [with embedded encryption] and then I'll push everything at once. :-)

PR #142, opened 2017-02-14
T329 new configuration

alexeykiselev:


phearnot:
I suggest we use `=`, which is easier to type, instead of a colon.

PR #145, opened 2017-02-16
T370 start with new config

alexeykiselev:


src/main/scala/com/wavesplatform/Application.scala
```diff
@@ -93,8 +95,16 @@ class Application(as: ActorSystem, wavesSettings: WavesSettings) extends {
 object Application extends ScorexLogging {
   def main(args: Array[String]): Unit = {
     log.info("Starting...")
-    val config: Config = ConfigFactory.load()
-    val settings = WavesSettings.fromConfig(config)
+
+    val maybeConfigurationFilename = args.headOption
+    val config = if (maybeConfigurationFilename.isDefined && new File(maybeConfigurationFilename.get).exists) {
+      log.info(s"Additional configuration will be loaded from ${maybeConfigurationFilename.get}")
+      ConfigFactory.parseFile(new File(maybeConfigurationFilename.get)).withFallback(ConfigFactory.load())
```
phearnot:
```scala
val userConfig = for {
  cf <- args.headOption
  f = new File(cf)
  if f.exists
} yield ConfigFactory.parseFile(f)

val config = userConfig.fold(ConfigFactory.load()) { c => c.withFallback(ConfigFactory.load()) }
```

src/main/scala/com/wavesplatform/Application.scala
```diff
@@ -20,6 +20,7 @@ import scorex.utils.ScorexLogging
 import scorex.waves.http.{DebugApiRoute, WavesApiRoute}
 import scorex.waves.transaction.WavesTransactionModule
 
+import scala.Option.option2Iterable
```
phearnot: Explicitly importing this one is unnecessary — implicit conversions are searched for in a companion object by default.

src/main/scala/com/wavesplatform/Application.scala
```diff
@@ -16,8 +19,8 @@ import scorex.network.{TransactionalMessagesRepo, UnconfirmedPoolSynchronizer}
 import scorex.utils.ScorexLogging
 import scorex.waves.http.{DebugApiRoute, WavesApiRoute}
 import scorex.waves.transaction.WavesTransactionModule
-import BlockchainSettingsExtension._
 
+import scala.Option.option2Iterable
```
phearnot: There's no need to import this explicitly — implicit conversions are searched for in a source type's companion object by default.

alexeykiselev: I think it was imported automatically. I've made a new commit to remove it.

PR #153, opened 2017-02-21
Exchange transaction hardfork param

koloale:


src/test/scala/com/wavesplatform/matcher/settings/BlockchainSettingsSpecification.scala
```diff
@@ -26,6 +26,7 @@ class BlockchainSettingsSpecification extends FlatSpec with Matchers {
         |        allow-burn-transaction-after: 8
         |        require-payment-unique-id-after: 9
         |        allow-lease-transaction-after: 10
+        |        allow-exchange-transaction-after: 11
```
alexeykiselev: You have to test new values in test :)

koloale: Added check in test

PR #162, opened 2017-02-22
T412 Allow empty strings in asset ids and attachments

phearnot:


alexeykiselev:
Test failed:
- transfer transaction *** FAILED *** (56 milliseconds)
TestFailedException was thrown during property evaluation.
Message: expected {"error":101,"message":"invalid signature"}, but instead got {"error":113,"message":"overflow error"}
Location: (AssetsBroadcastRouteSpec.scala:85)

PR #172, opened 2017-03-02
T439: pre validation utx sorting was added

Tolsi:


src/main/scala/scorex/transaction/SimpleTransactionModule.scala
```diff
@@ -98,7 +98,8 @@ class SimpleTransactionModule(hardForkParams: ChainParameters)(implicit val sett
     val txs = utxStorage.all()
     val notExpired = txs.filter { tx => (currentTime - tx.timestamp).millis <= MaxTimeForUnconfirmed }
     val notFromFuture = notExpired.filter { tx => (tx.timestamp - currentTime).millis <= MaxTimeDrift }
-    val valid = blockStorage.state.validate(notFromFuture, blockTime = currentTime)
```
ismagin: Have you considered `UnconfirmedTransactionsDatabaseImpl.transactions` to be a `Sorted` structure so adding/removing wouldn't require explicit resorting?

PR #192, opened 2017-03-15
T506: issuance/burn transactions apply and rollback fixes

Tolsi:


src/main/scala/scorex/transaction/state/database/blockchain/AssetsExtendedState.scala
```diff
@@ -51,17 +51,17 @@ class AssetsExtendedState(storage: StateStorageI with AssetsExtendedStateStorage
     val assetAtHeight = s"$asset@$height"
     val assetAtTransaction = s"$asset@$transaction"
 
-    if (!isIssueExists(assetId) ||
+    if (!(
```
ismagin: It looks overcomplicated plus I'm not sure `reissuable` should even be there.
Is that the logic needed?
```
if(!issueExists(assetId) throw new Exception("Related Issue tx not exists")
if(!isReissuable(assetId) throw new Exception("Issue not reissuable")
```
?

Tolsi: This is a bad sign, we only thanks to him caught a few bugs

PR #201, opened 2017-03-22
T461 integration tests

phearnot:


src/it/scala/com/wavesplatform/it/AliasTransactionSpec.scala
```diff
@@ -0,0 +1,47 @@
...
+
+    val createAliasFee = feeSettings(TransactionType.CreateAliasTransaction.id).find(_.asset == "WAVES").get.fee
+    val transferFee = feeSettings(TransactionType.TransferTransaction.id).find(_.asset == "WAVES").get.fee
+
+    val transferResult = for {
+      fb <- creator.lastBlock
+      t <- creator.createAlias(creator.address, "TEST-ALIAS", createAliasFee)
+      b <- creator.findBlock(_.transactions.exists(_.id == t.id), fb.height)
+      _ <- sender.waitForHeight(b.height)
+      t <- sender.transfer(sender.address, s"alias:${sender.settings.blockchainSettings.addressSchemeCharacter}:TEST_ALIAS", 1000000, transferFee)
```
ismagin: "TEST-ALIAS" != "TEST_ALIAS" :)

src/it/scala/com/wavesplatform/it/ValidChainGenerationSpec.scala
```diff
@@ -0,0 +1,44 @@
...
+import scala.concurrent.duration._
+import scala.concurrent.{Await, Future}
+import scala.util.Random
+
+class ValidChainGenerationSpec(allNodes: Seq[Node]) extends FreeSpec with ScalaFutures with IntegrationPatience
+  with Matchers with Eventually {
+
+  "Generate 30 blocks and synchronise" in {
+    val addressToNode = allNodes.map(n => n.address -> n).toMap
+    var balances = Await.result(Future.traverse(allNodes)(n => n.balance(n.address))
```
ismagin: Let's not use `var` in integration tests at all.

src/it/scala/com/wavesplatform/it/ValidChainGenerationSpec.scala
```diff
@@ -0,0 +1,44 @@
...
+
+    def makeTransfer(): Future[Node.Transaction] = {
+      val shuffledAddresses = Random.shuffle(addressToNode.keys)
+      val sourceAddress = shuffledAddresses.head
+      val sourceNode = addressToNode(sourceAddress)
+      val targetAddress = Random.shuffle(shuffledAddresses.tail).head
+      val fee = 100000
+      val amount = (Random.nextDouble() * 1e-3 * (balances(sourceAddress) - fee)).toLong
+
+      balances += sourceAddress -> (balances(sourceAddress) - amount)
```
ismagin: As the test is written now, it takes too much attention to understand, what is it actually doing.
The busniess logic of transfer is interleaved with the building transfer itself and some state of some `var`.

I understand this might be some "special" complex test, but if we want this to be a sort of reference implementation, we need to simplify it -- or at least try less mutable approach.


Regarding the random, itself, I'd give another shot to generators since they do compose.

src/it/scala/com/wavesplatform/it/Docker.scala
```diff
@@ -0,0 +1,124 @@
...
+
+  private def renderProperties(p: Properties) = p.asScala.map { case (k, v) => s"-D$k=$v" } mkString " "
+
+  private def extractHostPort(m: JMap[String, JList[PortBinding]], containerPort: String) =
+    m.get(s"$containerPort/tcp").get(0).hostPort().toInt
+
+  val DefaultConfigTemplate = ConfigFactory.parseResources("template.conf")
+  val NodeConfigs = ConfigFactory.parseResources("nodes.conf")
+
+  def apply(suiteConfig: Config = ConfigFactory.empty): Docker = new Docker {
```
ismagin: Is anonymous implementation worth it?

src/it/resources/nodes.conf
```diff
@@ -0,0 +1,81 @@
+nodes = [{
+  waves {
```
ismagin: Isn't this too much to be consistent? 
I mean wallet.seed produces priv+pub key, pub-key produces address.
We can easily get to incosistent state if we make things non-algebraic.

As (I guess) we need some of the params just for the reference, we can make them comments.

src/it/scala/com/wavesplatform/it/Node.scala
```diff
@@ -0,0 +1,152 @@
...
+      def retrying(timeout: Timeout): Unit = future.onComplete {
+        case Success(v) if cond(v) => p.success(v)
+        case _ => try {
+          timer.newTimeout(retrying, interval.toMillis, MILLISECONDS)
+        } catch {
+          case NonFatal(e) => p.failure(e)
+        }
+      }
+
+      retrying(null)
```
ismagin: `null` :(

src/it/scala/com/wavesplatform/it/Node.scala
```diff
@@ -0,0 +1,155 @@
...
+              log.debug(s"Loaded ${blocks.length} blocks, no match found. Next range: [$newFrom, ${newFrom + 99}]")
+              reschedule(newFrom, (newFrom + 99).min(to))
+            }
+        }
+      case Failure(t) =>
+        log.debug("Error loading blocks", t)
+        reschedule(_from, _to)
+    }
+
+    retrying(from, from + 99)(null)
```
ismagin: Is `timeout` param even used in the function?
I vote to ban all `null`s from any scala code.

src/it/scala/com/wavesplatform/it/ValidChainGenerationSpec.scala
```diff
@@ -0,0 +1,44 @@
...
+      balances += targetAddress -> (balances(targetAddress) + amount)
+
+      sourceNode.transfer(sourceAddress, targetAddress, amount, fee)
+    }
+
+    Await.result(Future.traverse(1 to 1000)(_ => makeTransfer()), 1.minute)
```
ismagin: This is a very small test. `Await.result(Future.traverse` appears 3 times here, it's already a pattern to make a syncronous helper method. \
Just saying :)

src/it/scala/com/wavesplatform/it/Node.scala
phearnot: It's a trick to use a function where an instance implementing a functional [TimerTask](https://netty.io/4.0/api/io/netty/util/TimerTask.html) interface is required. In this very case, null is like a seed for a (semi-)recursive function.

src/it/resources/nodes.conf
phearnot: Unfortunately, this can not easily be fixed because of how `chainId` is implemented.

PR #220, opened 2017-03-31
T525 Node API Specification

phearnot:


src/main/scala/com/wavesplatform/Shutdownable.scala
```diff
@@ -0,0 +1,5 @@
+package com.wavesplatform
+
+trait Shutdownable {
```
ismagin: Do we ever use this trait?

phearnot: The only thing Node API route needs from an application is a `shutdown()` method. Extracting this method into an interface and making `NodeAPIRoute` depend on `Shutdownable` rather than `RunnableApplication` makes testing so much easier. Pushing `shutdown()` to `Application` didn't seem like a good idea to me.

PR #224, opened 2017-04-03
T442 rewrite state

ismagin:
Points left to do:
- ~~Review new tests cover everything deleted tests did~~
- Consider ReadWriteLock(http://vlkan.com/blog/post/2015/09/09/enforce-locking/) on `BlockchainUpdaterImpl`, probably after `History` is cleaned up
- Rewrite ~~`MatcherActorSpecification`~~, `OrderMatchTransactionSpecification` -- hopefully @koloale will participate
- Parameterize in-mem-diff size
- 2-step in-mem diff to speed-up
- drop `State` by dropping file(when separated, probably after `History` is cleaned up)
- don't use `TestSettings.ENABLED` for settings-dependent tests
- Drop/rework `AssetAcc` in `Block`


src/main/scala/com/wavesplatform/state2/Diff.scala
```diff
@@ -0,0 +1,71 @@
...
+
+case class BlockDiff(txsDiff: Diff, heightDiff: Int, effectiveBalanceSnapshots: Seq[EffectiveBalanceSnapshot])
+
+object BlockDiff {
+
+  implicit class BlockDiffExt(bd: BlockDiff) {
+    lazy val maxPaymentTransactionTimestamp: Map[Account, Long] = bd.txsDiff.transactions.toList
+      .collect({ case (_, (_, ptx: PaymentTransaction, _)) => ptx.sender.toAccount -> ptx.timestamp })
+      .groupBy(_._1)
+      .map { case (acc, list) => acc -> list.map(_._2).max }
```
phearnot:
```scala
txsDiff.transactions.values
  .collect { case (_, ptx: PaymentTransaction, _) => ptx.sender.toAccount -> ptx.timestamp }
  .groupBy(_._1)
  .mapValues(_.max)
```

src/main/scala/com/wavesplatform/state2/reader/StateReader.scala
```diff
@@ -0,0 +1,124 @@
...
+          case None => Left(AliasNotExists(a))
+          case Some(acc) => Right(acc)
+        }
+      }
+    }
+
+    def findPreviousExchangeTxs(order: Order): Set[ExchangeTransaction] =
+      s.findPreviousExchangeTxs(EqByteArray(order.id))
+
+    def included(signature: Array[Byte]): Option[Int] = s.transactionInfo(EqByteArray(signature)).map(_._1)
```
ismagin: This function is part of `implicit class StateReaderExt(s: StateReader)`.
This prevented compilation errors at places where old `State` with the exact same signature was used. Now it's `StateReader` with this extension. Most of these extension methods are not needed at all and can be inlined, I haven't done this to prevent merge conflicts in case someone decides to rely on these in `master` branch.

src/main/scala/com/wavesplatform/state2/reader/StateReader.scala
```diff
@@ -0,0 +1,124 @@
...
+
+    def accountTransactions(account: Account, limit: Int): Seq[_ <: Transaction] =
```
ismagin: Paging is awesome, but, most probably, we won't do it before merge

src/main/scala/com/wavesplatform/state2/reader/StateReader.scala
```diff
@@ -0,0 +1,124 @@
...
+      s.accountTransactionIds(account).take(limit).flatMap(s.transactionInfo).map(_._2)
+
+    def balance(account: Account): Long = s.accountPortfolio(account).balance
```
ismagin: We have such method in HTTP API, if we want to change this, we need to discuss or whatever. 
We can do it in the next step

src/main/scala/com/wavesplatform/state2/reader/StateReader.scala
```diff
@@ -0,0 +1,124 @@
...
+
+    def assetBalance(account: AssetAcc): Long = {
+      val accountPortfolio = s.accountPortfolio(account.account)
+      account.assetId match {
+        case Some(assetId) => accountPortfolio.assets.getOrElse(EqByteArray(assetId), 0)
+        case None => accountPortfolio.balance
+      }
+    }
+
+    def getAccountBalance(account: Account): Map[AssetId, (Long, Boolean, Long, IssueTransaction)] =
```
ismagin: This is also extension method. This one provides `IssueTransaction`.

src/main/scala/com/wavesplatform/state2/reader/StateReaderImpl.scala
```diff
@@ -0,0 +1,82 @@
...
+      .map(_.toSeq)
+      .getOrElse(Seq.empty)
+      .map(EqByteArray)
+  }
+
+  override def nonEmptyAccounts: Seq[Account] =
+    p.portfolios
+      .keySet()
+      .asScala
+      .map(b => Account.fromBytes(b).right.get)
```
ismagin: I don't think I can to much about it.
Yes, in this particular case it's possible to flatten/whatever, but the code generally relies on data stored being correct. This includes deserialising txs,accs, other 'smart-constructed' entities from state.

PR #231, opened 2017-04-06
T618: lease cancel json extended with lease transaction json

Tolsi:


src/main/scala/scorex/api/http/TransactionsApiRoute.scala
```diff
@@ -73,6 +74,14 @@ case class TransactionsApiRoute(
...
   @ApiOperation(value = "Unconfirmed", notes = "Get list of unconfirmed transactions", httpMethod = "GET")
   def unconfirmed: Route = (path("unconfirmed") & get) {
-    complete(JsArray(transactionModule.unconfirmedTxs.map(_.json)))
+    complete(JsArray(transactionModule.unconfirmedTxs.map(txToExtendedJson)))
+  }
+
+  private def txToExtendedJson(tx: Transaction): JsObject = {
+    tx match {
+      case leaseCancel: LeaseCancelTransaction =>
+        leaseCancel.json ++ JsObject(Map("lease" -> state.findTransaction[LeaseTransaction](leaseCancel.leaseId).map(_.json).getOrElse(JsNull)))
```
ismagin: Will this `lease`(or better `__lease`) be as a subnode of LeaseCancelTransaction json, or will it be at the same level?

having LeaseCancelTranscation referencing non-existing LeaseTransaction is a serious problem, probably we should be more explicit that `getOrElse(JsNull)`.
We do unsafe gets in more critical places like resolving IssueTx w/o defaulting it to null, let's discuss our approach

phearnot: `JsObject(Map(...` -> Json.obj("lease" ->...`

PR #232, opened 2017-04-06
T586 address balances api

Tolsi:


src/main/scala/scorex/api/http/AddressApiRoute.scala
```diff
@@ -121,6 +122,15 @@ case class AddressApiRoute(settings: RestAPISettings, wallet: Wallet, state: Sta
     complete(balanceJson(address, 0))
   }
 
+  @Path("/balance/details/{address}")
+  @ApiOperation(value = "Details for balance", notes = "Account's balances", httpMethod = "GET")
+  @ApiImplicitParams(Array(
+    new ApiImplicitParam(name = "address", value = "Address", required = true, dataType = "string", paramType = "path")
+  ))
+  def balanceDetails: Route = (path("balance" / "details" / Segment) & get) { address =>
+    complete(balancesDetailsJson(address)(transactionModule))
```
ismagin: So first you make transaction module `implicit` parameter and then pass it explicitly

src/main/scala/scorex/api/http/AddressApiRoute.scala
```diff
@@ -228,6 +238,19 @@ case class AddressApiRoute(settings: RestAPISettings, wallet: Wallet, state: Sta
       .getOrElse(InvalidAddress)
   }
 
+  private def balancesDetailsJson(address: String)(implicit transactionModule: TransactionModule): ToResponseMarshallable = {
```
ismagin: please make this method of tansaction module with sig : `String=>BalanceDetails`, than invoke it here. it should be marshalled automatically

ismagin: The idea of this class is to provide routing, business logic is best to be kept elsewhere

src/test/scala/com/wavesplatform/http/AddressRouteSpec.scala
```diff
@@ -136,6 +153,7 @@ class AddressRouteSpec extends RouteSpec("/addresses") with PathMockFactory with
   routePath("/verifyText/{address}") in testVerify("verifyText", false)
   routePath("/verify/{address}") in testVerify("verify", true)
 
+  // todo add balance details test
```
phearnot: So there's no actual test, right?

src/test/scala/com/wavesplatform/http/AddressRouteSpec.scala
```diff
@@ -35,7 +34,25 @@ class AddressRouteSpec extends RouteSpec("/addresses") with PathMockFactory with
...
+    m
+  }
+
+  private val transactionModule = {
+    val m = mock[TransactionModule]
+    (m.blockStorage _).expects().returning(blockStorage).anyNumberOfTimes()
+    m
+  }
+
+  private val route = AddressApiRoute(restAPISettings, wallet, state, consensusModule, transactionModule).route
```
phearnot: You should not set expectations at class level unless all the tests require them. Also why is block storage mock here? If the route itself requires block storage, make the route depend on it explicitly, not implicitly via TransactionModule.

src/main/scala/scorex/api/http/AddressApiRoute.scala
```diff
@@ -213,8 +238,17 @@ case class AddressApiRoute(settings: RestAPISettings, wallet: Wallet, state: Sta
...
+  }
+
+  private def balancesDetailsJson(account: Account): BalanceDetails = {
+    val portfolio = state.accountPortfolio(account)
+    BalanceDetails(
+      account.address,
+      portfolio.balance,
+      consensusModule.generatingBalance(account, state.height)(transactionModule),
+      portfolio.balance - portfolio.leaseInfo.leaseOut,
+      state.effectiveBalance(account))
```
ismagin: jsut FYI, `portfolio already contains all the info, it's plain `lazy val effectiveBalance`

PR #234, opened 2017-04-10
orderbook_serialization_issue

koloale:


src/main/scala/scorex/transaction/assets/exchange/Order.scala
```diff
@@ -167,7 +166,7 @@ case class Order(@ApiModelProperty(dataType = "java.lang.String") senderPublicKe
     }
   }
```
ismagin: please fix equals too

src/main/scala/com/wavesplatform/matcher/model/EventSerializers.scala
```diff
@@ -76,14 +76,14 @@ object MatcherSerializer {
...
-      JsSuccess(TreeMap.empty[Price, Level[BuyLimitOrder]] ++ a)
+      JsSuccess(TreeMap.empty[Price, Level[BuyLimitOrder]](OrderBook.bidsOrdering) ++ a)
     }
   }
 
   implicit val sellOrderTreeMapReads = new Reads[TreeMap[Price, Level[SellLimitOrder]]] {
     override def reads(jv: JsValue): JsResult[TreeMap[Price, Level[SellLimitOrder]]] = {
       val a = jv.as[Map[String, Level[SellLimitOrder]]].map { case(k, v) => (k.toLong, v) }
-      JsSuccess(TreeMap.empty[Price, Level[SellLimitOrder]] ++ a)
+      JsSuccess(TreeMap.empty[Price, Level[SellLimitOrder]](OrderBook.asksOrdering) ++ a)
```
phearnot: TreeMap is serialized as a JSON object, which is [an unordered set of name/value pairs](http://www.json.org/). Explicitly specifying ordering isn't going to fix an issue. If you need to guarantee level ordering, you probably need to serialize bids/asks as an array.

phearnot: Disregard that.

src/main/scala/scorex/transaction/assets/exchange/Order.scala
koloale: What's wrong with equals? Currently, fields, which are used in equals also used to create idStr hence hashcode

PR #235, opened 2017-04-11
T635: unifying to stringRepr for alias-address mappings in StoredState

ismagin:


ismagin:
@phearnot when one sees toString, he has no idea whether it's implemented properly, can/cannot be  changed. For me, it has a semantics of repl/debug output. `stringRepr` is more of a cannonical string representation across the system. 

Other than that, if you have strong desire to switch to `toString`, I wouldn't mind.

PR #248, opened 2017-04-20
T442 rewrite state -- Separate Blockchain from State

ismagin:
Separates blockchain from state and checkpoints. This allows to drop state by deleting a file. Upon start, state can be recalculated based on existing blockchain.


alexeykiselev:
In Matcher tests where is an error:
[info] MatcherAPISpecification:
java.lang.NullPointerException
	at scala.collection.mutable.ArrayOps$ofByte$.length$extension(ArrayOps.scala:211)
	at scala.collection.mutable.ArrayOps$ofByte.length(ArrayOps.scala:211)
	at scala.collection.IndexedSeqOptimized.isEmpty(IndexedSeqOptimized.scala:26)
Is it OK?

src/main/scala/com/wavesplatform/history/BlockStorageImpl.scala
```diff
@@ -1,29 +1,33 @@
...
 class BlockStorageImpl(settings: BlockchainSettings) extends BlockStorage {
 
   import BlockStorageImpl._
 
-  val database: MVStore = createMVStore(settings.file)
-  val h: History = new StoredBlockchain(database)
-  val mVStorePrimitiveImpl = new MVStorePrimitiveImpl(database)
-  val rw = new StateWriterImpl(mVStorePrimitiveImpl)
-  val updater = new BlockchainUpdaterImpl(rw, settings.functionalitySettings, h)
+  val hdatabase: MVStore = createMVStore(settings.blockchainFile)
```
alexeykiselev: h-, b- and c-something sounds like a lazy naming...

src/main/scala/com/wavesplatform/history/HistoryWriterImpl.scala
```diff
@@ -0,0 +1,82 @@
...
+  override def discardBlock(): Unit = {
+    require(height() > 1, "Chain is empty or contains genesis block only, can't make rollback")
+    val h = height()
+    blocksCache.invalidate(h)
+    storage.blockBodyByHeight.remove(h)
+    val vOpt = Option(storage.blockIdByHeight.remove(h))
+    vOpt.map(v => storage.heightByBlockId.remove(v))
+  }
+
+  override def blockAt(height: Int): Option[Block] = {
```
phearnot:
```scala
Exception.allCatch.opt(blockCache.get(height))
```

src/main/scala/com/wavesplatform/history/HistoryWriterImpl.scala
```diff
@@ -0,0 +1,82 @@
...
+import com.google.common.util.concurrent.UncheckedExecutionException
+import scorex.account.Account
+import scorex.block.Block
+import scorex.block.Block.BlockId
+import scorex.transaction.History.BlockchainScore
+import scorex.transaction.ValidationError.CustomError
+import scorex.transaction.{History, HistoryWriter, ValidationError}
+import scorex.utils.ScorexLogging
+
+class HistoryWriterImpl(storage: HistoryStorage) extends History with HistoryWriter with ScorexLogging {
```
phearnot: Three different approaches to error handling are used in a single class: returning Either, returning an option with no clue and just throwing an exception. This doesn't look neat to me.

src/main/scala/com/wavesplatform/history/HistoryWriterImpl.scala
```diff
@@ -0,0 +1,82 @@
...
+  override def appendBlock(block: Block): Either[ValidationError, Unit] = {
+    if ((height() == 0) || (this.lastBlock.uniqueId sameElements block.reference)) {
+      val h = height() + 1
+      storage.blockBodyByHeight.put(h, block.bytes)
+      storage.scoreByHeight.put(h, score() + block.blockScore)
+      storage.blockIdByHeight.put(h, block.uniqueId)
+      storage.heightByBlockId.put(block.uniqueId, h)
+      Right(())
+    } else {
+      Left(CustomError(s"Appending block ${block.json} which parent is not last block in blockchain"))
```
phearnot: With a full block of 100 transactions, this error message will be too long to be meaningful. I suggest mentioning just block id and maybe block parent's id.

src/main/scala/com/wavesplatform/history/HistoryWriterImpl.scala
```diff
@@ -0,0 +1,82 @@
...
+import scorex.transaction.ValidationError.CustomError
+import scorex.transaction.{History, HistoryWriter, ValidationError}
+import scorex.utils.ScorexLogging
+
+class HistoryWriterImpl(storage: HistoryStorage) extends History with HistoryWriter with ScorexLogging {
+
+  private val BlocksCacheSizeLimit: Int = 1000
+  private val blocksCache = CacheBuilder.newBuilder()
+    .maximumSize(BlocksCacheSizeLimit)
+    .build[Integer, Block](new CacheLoader[Integer, Block]() {
```
phearnot:
```scala
  .build[Int, Block](CacheLoader.from { height =>
    Block.parseBytes(storage.blockBodyByHeight.get(height)).get
  })
```
Also please check other places where you create cache loaders, if any

src/main/scala/com/wavesplatform/history/HistoryWriterImpl.scala
```diff
@@ -0,0 +1,82 @@
...
+
+  override def height(): Int = storage.blockIdByHeight.size()
+
+  override def score(): BlockchainScore = if (height() > 0) storage.scoreByHeight.get(height()) else 0
+
+  override def scoreOf(id: BlockId): BlockchainScore = heightOf(id).map(storage.scoreByHeight.get(_)).getOrElse(0)
+
+  override def heightOf(blockSignature: Array[Byte]): Option[Int] = Option(storage.heightByBlockId.get(blockSignature))
+
+  override def generatedBy(account: Account, from: Int, to: Int): Seq[Block] = {
```
phearnot:
```scala
for {
  h <- from to to
  block <- blockAt(h)
  if block.signerData.generator.address.equals(account.address)
} yield block
```
Also no braces needed around a single expression

src/main/scala/com/wavesplatform/history/HistoryWriterImpl.scala
```diff
@@ -0,0 +1,82 @@
...
+    (from to to).toStream.flatMap {
+      h =>
+        for {
+          block <- blockAt(h)
+          if block.signerData.generator.address.equals(account.address)
+        } yield block
+    }
+  }
+
+  override def toString: String = ((1 to height()) map {
```
phearnot:
```scala
(for {
  h <- 1 to height()
  block <- blockAt(h)
} yield "...").mkString("\n")

File: src/main/scala/com/wavesplatform/history/CheckpointStorage.scala [hunk 8]
```diff
@@ -0,0 +1,35 @@
...
+  def commit(): Unit
+}
+
+class MVStoreCheckpointStorage(db: MVStore) extends CheckpointStorage {
+  val checkpoint: MVMap[Int, (Seq[(Int, Array[Byte])], Array[Byte])] = db.openMap("checkpoint", new LogMVMapBuilder[Int, (Seq[(Int, Array[Byte])], Array[Byte])])
+
+  def commit(): Unit = db.commit()
+}
+
+class CheckpointServiceImpl(storage: CheckpointStorage) extends CheckpointService {
```
phearnot: There doesn't seem to be any other implementations of `CheckpointStorage` except for `MVStoreCheckpointStorage`, and there are no tests where `CheckpointServiceImpl`'s functionality is tested with a fake instance of `CheckpointStorage`. My point is, instead of three entities, why not have just one
```scala
class CheckpointServiceImpl(db: MVStore) {
```

src/main/scala/com/wavesplatform/settings/BlockchainSettings.scala
```diff
@@ -156,10 +156,15 @@ object BlockchainSettings {
...
         val genesisSettings = GenesisSettings.fromConfig(config)
-
         (addressSchemeCharacter, functionalitySettings, genesisSettings)
     }
 
-    BlockchainSettings(file, addressSchemeCharacter, functionalitySettings, genesisSettings)
+    BlockchainSettings(
+      blockchainFile = config.as[String](s"$configPath.blockchain-file"),
+      stateFile = config.as[String](s"$configPath.state-file"),
+      checkpointFile = config.as[String](s"$configPath.checkpoint-file"),
```
phearnot: Ficus supports custom value extractors, so how about creating a type class and change this to
```scala
blockchainFile = config.as[Path](s"$configPath.blockchain-file")
```

src/main/scala/com/wavesplatform/state2/StateWriter.scala
```diff
@@ -96,4 +114,19 @@ class StateWriterImpl(p: JavaMapStorage) extends StateReaderImpl(p) with StateWr
...
+object StateWriterImpl extends ScorexLogging {
+
+  private def withTime[R](f: => R): (R, Long) = {
+    val t0 = System.currentTimeMillis()
+    val r: R = f
+    val t1 = System.currentTimeMillis()
+    (r, t1 - t0)
+  }
+
+  def measurePersist[F[_] <: TraversableOnce[_], A](s: String)(fa: F[A])(f: F[A] => Unit): Unit = {
```
phearnot: Looks like the time has finally come to make use of [Metrics](http://metrics.dropwizard.io/3.2.2/)!

src/main/scala/com/wavesplatform/state2/diffs/BlockDiffer.scala
```diff
@@ -38,7 +38,7 @@ object BlockDiffer extends ScorexLogging {
         })
     }
     lazy val feeDiff = Monoid[Diff].combineAll(accountPortfolioFeesMap.map { case (acc, p) =>
-      Diff(portfolios = Map(acc -> p))
+      new Diff(Map.empty, Map(acc -> p), Map.empty, Map.empty, Map.empty, Map.empty, Seq.empty)
```
phearnot: I suggest adding default empty values for all of the `Diff` constructor arguments.

src/main/scala/scorex/api/http/BlocksApiRoute.scala
```diff
@@ -79,8 +80,14 @@ case class BlocksApiRoute(settings: RestAPISettings, checkpointsSettings: Checkp
     new ApiImplicitParam(name = "signature", value = "Base58-encoded signature", required = true, dataType = "string", paramType = "path")
   ))
   def heightEncoded: Route = (path("height" / Segment) & get) { encodedSignature =>
-    withBlock(history, encodedSignature) { block =>
-      complete(Json.obj("height" -> history.heightOf(block)))
+    if (encodedSignature.length > TransactionParser.SignatureStringLength)
```
phearnot: We should probably come up with a better utility method for extracting signatures from request URIs.

src/main/scala/scorex/transaction/History.scala
```diff
@@ -7,125 +7,80 @@ import scorex.crypto.encode.Base58
...
-    *
-    * Append block to a chain, based on it's reference
-    *
-    * @param block - block to append
-    * @return Blocks to process in state
-    */
-  def appendBlock(block: Block): Either[ValidationError, Unit]
+  def setCheckpoint(checkpoint: Option[Checkpoint])
 
+  def getCheckpoint: Option[Checkpoint]
```
phearnot: The methods could be named just `get`/`set` for brevity.

src/main/scala/com/wavesplatform/history/HistoryWriterImpl.scala
ismagin: `Either` for `appendBlock` and `Option` for `blockAt` are natural as to the semantics of what they are doing. Makeing `blockAt` `Either` doesn't make any sense.
Regarding exceptions, these are merely results of using cacheloader(not refactored at this point).
They do not get exposed so the usage is safe.

Anyway if you have an idea how to easily drop exceptions here(the problem is with sig `def load(height: Integer): Block`, please advice. We could return `null` in cacheloader on a miss, and then make `Option(..)` at `blockAt`, but that doesn't really change a lot.

src/main/scala/com/wavesplatform/history/HistoryWriterImpl.scala
ismagin: ok

src/main/scala/com/wavesplatform/history/HistoryWriterImpl.scala
ismagin: ok

src/main/scala/com/wavesplatform/history/HistoryWriterImpl.scala
ismagin: ok

src/main/scala/com/wavesplatform/history/CheckpointStorage.scala
ismagin: The idea behind the special trait is to separate the way we hande data from the way we actually store it. Unfortunately this didn't work out for testing due to mvstore specifics.  But in case we migrate from mvstore, we'll have to switch Storages without(idealistically) touching *ServiceImpl. I'd prefer to keep this separation.

src/main/scala/com/wavesplatform/history/HistoryWriterImpl.scala
ismagin: ok

src/main/scala/com/wavesplatform/history/HistoryWriterImpl.scala
ismagin: I'll drop that at all, we don't use it

src/main/scala/com/wavesplatform/settings/BlockchainSettings.scala
ismagin: That just might be out of scope of PR, but I don't mind changing that. Can you implement the change?

src/main/scala/com/wavesplatform/state2/StateWriter.scala
ismagin: Shall we wait for one more case before introducing a dependency? I'd also consider more scala idiomatic libraries if such exist.

src/main/scala/com/wavesplatform/state2/diffs/BlockDiffer.scala
ismagin: ..and it does exist! Decided not to create more so noone gets confused with too specific overloadings.

src/main/scala/scorex/transaction/History.scala
ismagin: ok

src/main/scala/scorex/api/http/BlocksApiRoute.scala
ismagin: probably :)

PR #257, opened 2017-04-24
Apply fee diffs first, then transactions diffs

Tolsi:


src/main/scala/com/wavesplatform/state2/diffs/BlockDiffer.scala
```diff
@@ -17,32 +17,32 @@ object BlockDiffer extends ScorexLogging {
...
+    val feeDiff = Monoid[Diff].combineAll(accountPortfolioFeesMap.map { case (acc, p) =>
       Diff(portfolios = Map(acc -> p))
     })
 
+    val txsDiffEi = block.transactionData.foldLeft(rightDiff(feeDiff)) { case (ei, tx) => ei match {
+      case Left(error) => Left(error)
+      case Right(diff) =>
+        txDiffer(new CompositeStateReader(s, diff.asBlockDiff), tx)
+          .map(newDiff => diff.combine(newDiff))
+      }
```
phearnot: Why not just use `ei.map(diff => ...)` here?

src/main/scala/com/wavesplatform/state2/diffs/BlockDiffer.scala
```diff
@@ -17,32 +17,32 @@ object BlockDiffer extends ScorexLogging {
 
   val emptyDiff: Diff = Monoid[Diff].empty
```
ismagin: These two lines are now unsed, aren't they?

src/main/scala/com/wavesplatform/state2/diffs/BlockDiffer.scala
ismagin: what we actually need here is `foldM`, because we don't want to traverse all the collection after first error, yet we do it.

PR #258, opened 2017-04-25
Strange Release string was removed from application name and settings.

alexeykiselev:


phearnot:
Closing in favour of #259

PR #276, opened 2017-05-03
T715: state writer should increase height when applying block diff

Tolsi:


src/test/scala/com/wavesplatform/state2/StateWriterSpec.scala
```diff
@@ -0,0 +1,28 @@
+package com.wavesplatform.state2
+
+import org.h2.mvstore.MVStore
+import org.scalacheck.Gen
+import org.scalatest.prop.GeneratorDrivenPropertyChecks
+import org.scalatest.{Matchers, Outcome, fixture}
+
+class StateWriterSpec extends fixture.FunSuite with Matchers with GeneratorDrivenPropertyChecks {
+  val stateHeight = 100
```
ismagin: any reason to use this?

src/test/scala/com/wavesplatform/state2/StateWriterSpec.scala
```diff
@@ -0,0 +1,28 @@
...
+
+  override type FixtureParam = StateStorage
+
+  override protected def withFixture(test: OneArgTest): Outcome = {
```
ismagin: please drop the fixture from the test

PR #295, opened 2017-05-11
T442 sync read/write access to history and state

ismagin:
Further state syncronization, now State & History are both syncronized over 1 ReentrantReadWriteLock


ismagin:
IT tests passed twice locally.
At jenkins,

```
[info] - Generate 30 blocks and synchronise *** FAILED ***
[info]   'all' inspection failed, because: 
[info]     at index 3, 73 did not equal 72 (ValidChainGenerationSpec.scala:63) 
[info]   in ArrayBuffer(72, 72, 72, 73) (ValidChainGenerationSpec.scala:63)
```

PR #302, opened 2017-05-15
T508 rollback api

Tolsi:


src/main/scala/scorex/waves/http/DebugApiRoute.scala
```diff
@@ -66,6 +66,26 @@ case class DebugApiRoute(settings: RestAPISettings, wallet: Wallet, stateReader:
...
+  @Path("/rollback")
+  @ApiOperation(value = "Rollback to block", notes = "Removes all blocks after height", httpMethod = "GET")
+  @ApiResponses(Array(
+    new ApiResponse(code = 200, message = "200 if success, 404 if there are no block at this height")
+  ))
+  def rollback: Route = withAuth {
+    (path("rollback") & post) {
+      decodeRequest {
+        entity(as[String]) { rollbackTo =>
+          history.blockAt(rollbackTo.toInt) match {
```
ismagin: `entity(as[String]).toInt` seems unsafe, why not just `entity.as[Int]` ?

Tolsi: this does not work, I tried it

PR #303, opened 2017-05-16
T442: safesum to work with asset balance

ismagin:


src/main/scala/com/wavesplatform/state2/Portfolio.scala
```diff
@@ -18,6 +18,6 @@ object Portfolio {
     = Portfolio(
       balance = safeSum(older.balance, newer.balance),
       leaseInfo = Monoid.combine(older.leaseInfo, newer.leaseInfo),
-      assets = older.assets.combine(newer.assets))
+      assets = Monoid.combine(older.assets, newer.assets))
   }
 }
```
phearnot: Please add a missing line at the end of file.

PR #308, opened 2017-05-16
stop containers on shutdown hook

Tolsi:


src/it/scala/com/wavesplatform/it/Docker.scala
```diff
@@ -39,14 +40,22 @@ class Docker(suiteConfig: Config = ConfigFactory.empty) extends AutoCloseable wi
   private val client = DefaultDockerClient.fromEnv().build()
   private val timer = new HashedWheelTimer()
   private var nodes = Map.empty[String, NodeInfo]
+  private val isStopped = new AtomicBoolean(false)
 
   timer.start()
 
+  sys.addShutdownHook {
+    if (!isStopped.get()) {
```
phearnot: What's the point of having an `AtomicBoolean` and not using `compareAndSet`?

src/it/scala/com/wavesplatform/it/Docker.scala
```diff
@@ -115,6 +124,7 @@ class Docker(suiteConfig: Config = ConfigFactory.empty) extends AutoCloseable wi
     client.removeNetwork(wavesNetwork.id())
     client.close()
     http.close()
+    isStopped.set(true)
```
phearnot: The whole method should be wrapped in 
```scala
if (isStopped.compareAndSet(false, true)) {
   ...
}
```

PR #321, opened 2017-05-23
Transactions API was degraded.

alexeykiselev:


src/main/scala/scorex/api/http/TransactionsApiRoute.scala
```diff
@@ -51,7 +51,7 @@ case class TransactionsApiRoute(
               path(Segment) { limitStr =>
                 Exception.allCatch.opt(limitStr.toInt) match {
                   case Some(limit) if limit > 0 && limit <= MaxTransactionsPerRequest =>
-                    complete(JsArray(state.accountTransactions(a, limit).map(txToExtendedJson)))
+                    complete(Json.arr(JsArray(state.accountTransactions(a, limit).map(txToExtendedJson))))
```
Tolsi: Why using `Json.arr` and `JsArray` for the same?

alexeykiselev: Because it's impossible to reproduce old behavior using JsArray. It takes sequence and maps it, so it won't take other JsArray as param.

PR #323, opened 2017-05-24
T442: state version added

ismagin:


src/main/scala/com/wavesplatform/state2/StateStorage.scala
```diff
@@ -45,12 +51,28 @@ class StateStorage private(db: MVStore) {
 
 object StateStorage {
 
+  private val VERSION = 1
```
alexeykiselev: Maybe better put version to Constants object?

ismagin: it would be easier to forget about in on development as well as PR

PR #344, opened 2017-06-08
Netty network

phearnot:


src/main/scala/com/wavesplatform/Application.scala
```diff
@@ -39,80 +38,92 @@ import scorex.waves.http.{DebugApiRoute, WavesApiRoute}
...
+    val apiRoutes = Seq(
+      BlocksApiRoute(settings.restAPISettings, settings.checkpointsSettings, history, checkpoint => network.writeToLocalChannel(checkpoint)),
+      TransactionsApiRoute(settings.restAPISettings, stateReader, history, utxStorage),
+      NxtConsensusApiRoute(settings.restAPISettings, stateReader, history, settings.blockchainSettings.functionalitySettings),
+      WalletApiRoute(settings.restAPISettings, wallet),
+      PaymentApiRoute(settings.restAPISettings, wallet, newTransactionHandler, time),
+      UtilsApiRoute(settings.restAPISettings),
+      PeersApiRoute(settings.restAPISettings, network.connect, peerDatabase, establishedConnections),
+      AddressApiRoute(settings.restAPISettings, wallet, stateReader, settings.blockchainSettings.functionalitySettings),
+      DebugApiRoute(settings.restAPISettings, wallet, stateReader, history, peerDatabase, blockId => network.writeToLocalChannel(RollbackTo(blockId))),
```
ismagin: Using lambda-senders instead of real dependencies surely make code less verbose, but also less explicit. 
It is disturbing that you write to local channel instead of just invoking rollbackTo on blockchainUpdater. If it's necessary, it might indicate a poor abstraction, let's discuss this

src/main/scala/com/wavesplatform/Application.scala
```diff
@@ -39,80 +38,92 @@ import scorex.waves.http.{DebugApiRoute, WavesApiRoute}
...
+      establishedConnections)
+
+    val apiRoutes = Seq(
+      BlocksApiRoute(settings.restAPISettings, settings.checkpointsSettings, history, checkpoint => network.writeToLocalChannel(checkpoint)),
+      TransactionsApiRoute(settings.restAPISettings, stateReader, history, utxStorage),
+      NxtConsensusApiRoute(settings.restAPISettings, stateReader, history, settings.blockchainSettings.functionalitySettings),
+      WalletApiRoute(settings.restAPISettings, wallet),
+      PaymentApiRoute(settings.restAPISettings, wallet, newTransactionHandler, time),
+      UtilsApiRoute(settings.restAPISettings),
+      PeersApiRoute(settings.restAPISettings, network.connect, peerDatabase, establishedConnections),
```
ismagin: Please pass the real dependency

src/main/scala/com/wavesplatform/network/Coordinator.scala
```diff
@@ -193,134 +145,87 @@ class Coordinator(protected override val networkControllerRef: ActorRef, blockch
...
+    } else {
+      ctx.writeAndFlush(BlockchainUpdated)
+    }
   }
 
-
-  private def str(block: Block) = {
-    if (log.logger.isTraceEnabled) block.json
-    else block.uniqueId + ", parent " + block.reference
+  override def channelRead(ctx: ChannelHandlerContext, msg: AnyRef) = msg match {
```
ismagin: `Coordinator` is about business logic, `channelRead` is about implementation detail, also non-type-sage, they should be separated and `Coordinator`  needs to be referencened in channel reader.

src/main/scala/com/wavesplatform/network/Coordinator.scala
```diff
@@ -1,162 +1,114 @@
...
+    blockchainUpdater: BlockchainUpdater,
+    time: Time,
+    stateReader: StateReader,
+    utxStorage: UnconfirmedTransactionsStorage,
+    settings: BlockchainSettings,
+    checkpointPublicKey: String,
+    miner: Miner,
+    blacklist: Channel => Unit,
+    broadcast: (AnyRef, Option[Channel]) => Unit)
+  extends ChannelInboundHandlerAdapter with ScorexLogging {
```
ismagin: Again, non-transparent references

src/main/scala/com/wavesplatform/mining/Miner.scala
```diff
@@ -0,0 +1,88 @@
...
+
+/* This class is NOT SYNCHRONIZED AT ALL, it is meant to be used from Coordinator only! */
+class Miner(
+    history: History,
+    state: StateReader,
+    utx: UnconfirmedTransactionsStorage,
+    privateKeyAccounts: => Seq[PrivateKeyAccount],
+    time: Time,
+    bcs: BlockchainSettings,
+    blockHandler: Block => Unit) extends ScorexLogging {
```
ismagin: ...

src/main/scala/com/wavesplatform/network/Coordinator.scala
```diff
@@ -1,162 +1,114 @@
...
 import scala.util.control.NonFatal
-import com.wavesplatform.state2._
-
-class Coordinator(protected override val networkControllerRef: ActorRef, blockchainSynchronizer: ActorRef, blockGenerator: ActorRef,
-                  peerManager: ActorRef, scoreObserver: ActorRef, blockchainUpdater: BlockchainUpdater, time: Time,
-                  utxStorage: UnconfirmedTransactionsStorage,
-                  history: History, stateReader: StateReader, checkpoints: CheckpointService, settings: WavesSettings) extends ViewSynchronizer with ScorexLogging {
 
+@Sharable
+class Coordinator(
```
ismagin: We might argue whether `ExtensionLoader` is a protocol detail or not, but Coordinator for sure isn't. Its business logic needs to be clear from any `netty` implementation details

src/main/scala/com/wavesplatform/network/NetworkServer.scala
```diff
@@ -0,0 +1,223 @@
...
+                allChannels.remove(closeFuture.channel())
+                channels.remove(remoteAddress, closeFuture.channel())
+              }
+              allChannels.add(connFuture.channel())
+            }
+          }
+        }.channel()
+    })
+
+  def writeToLocalChannel(message: AnyRef): Unit = localClientChannel.writeAndFlush(message)
```
ismagin: In order to communicate with itself, the application uses netty. This can't be a good sign. 
Probably we can extract some other layer of abstraction.

src/main/scala/com/wavesplatform/network/DiscardingHandler.scala
```diff
@@ -0,0 +1,32 @@
...
+  private var blockchainIsOutdated = true
+
+  override def channelRead(ctx: ChannelHandlerContext, msg: AnyRef) = msg match {
+    case RawBytes(code, _) if code == TransactionMessageSpec.messageCode && blockchainIsOutdated =>
+      log.trace(s"${id(ctx)} Discarding incoming message $code")
+    case _ => super.channelRead(ctx, msg)
+  }
+
+  override def write(ctx: ChannelHandlerContext, msg: scala.Any, promise: ChannelPromise) = msg match {
+    case BlockchainExpired =>
```
ismagin: I do not recall  we have such network messages.
What I see here is a message passing interface between application components.
If we have to do this, literally ANY other way is better -- actors, services, plain event bus(in comparison to non-type-safe-network-library-driven one)

src/main/scala/com/wavesplatform/network/UtxPoolSynchronizer.scala
```diff
@@ -0,0 +1,25 @@
...
+import com.wavesplatform.state2.diffs.TransactionDiffer.TransactionValidationError
+import io.netty.channel.ChannelHandler.Sharable
+import io.netty.channel.{ChannelHandlerContext, ChannelInboundHandlerAdapter}
+import scorex.transaction.{NewTransactionHandler, Transaction}
+import scorex.utils.ScorexLogging
+
+@Sharable
+class UtxPoolSynchronizer(handler: NewTransactionHandler, network: Network)
+  extends ChannelInboundHandlerAdapter with ScorexLogging {
+  override def channelRead(ctx: ChannelHandlerContext, msg: AnyRef) = msg match {
```
ismagin: Business logic heavily mixed into nework implementation detail.

PR #353, opened 2017-06-15
Network tx integration tests

Tolsi:


Tolsi:
I added only the integration test, they work, you can see it

src/it/scala/com/wavesplatform/it/network/client/BasicMessagesRepo.scala
```diff
@@ -0,0 +1,155 @@
+package com.wavesplatform.it.network.client
```
ismagin: is this a copy-paste of `scorex.network.message.BasicMessagesRepo`?

src/it/scala/com/wavesplatform/it/transactions/TransferTransactionSpecification.scala
```diff
@@ -52,6 +60,79 @@ class TransferTransactionSpecification(override val allNodes: Seq[Node]) extends
     Await.result(f, 1 minute)
   }
 
+  test("invalid signed waves transfer should not be in UTX or blockchain") {
```
Tolsi: you can't create tx id from request for example

src/it/scala/com/wavesplatform/it/Node.scala
```diff
@@ -158,12 +173,25 @@ class Node (config: Config, val nodeInfo: NodeInfo, client: AsyncHttpClient, tim
...
+  def signedTransfer(transfer: SignedTransferRequest): Future[Transaction] =
+    postJson("/assets/broadcast/transfer", transfer).as[Transaction]
+
   def createAlias(targetAddress: String, alias: String, fee: Long): Future[Transaction] =
     postJson("/alias/create", CreateAliasRequest(targetAddress, alias, fee)).as[Transaction]
 
   def rollback(to: Long): Future[Unit] =
     post("/debug/rollback", to.toString).map(_ => ())
 
+  def isTransactionNotExists(txId: String): Future[Unit] =
```
Tolsi: how?

src/it/scala/com/wavesplatform/it/Node.scala
```diff
@@ -112,6 +121,12 @@ class Node (config: Config, val nodeInfo: NodeInfo, client: AsyncHttpClient, tim
 
   def balance(address: String): Future[Balance] = get(s"/addresses/balance/$address").as[Balance]
 
+  def secureTransactionInfo(txId: String): Future[Option[Transaction]] = transactionInfo(txId).transform {
+    case Success(tx) => Success(Some(tx))
```
ismagin: `findTransaction` would fit better, everybody understands what `find(..): Option[..]` does

src/it/scala/com/wavesplatform/it/Node.scala
ismagin: `ensureTxDoesntExist`, `transactionDoesNotExist`

src/it/scala/com/wavesplatform/it/transactions/TransferTransactionSpecification.scala
ismagin: Still no need for this code,  just create a transaction, then extract everything you need to `txId` and `invalidByTsSignedRequest`, this will reduce up the code up to 50 lines

src/it/scala/com/wavesplatform/it/network/client/Handshake.scala
```diff
@@ -0,0 +1,55 @@
+package com.wavesplatform.it.network.client
```
ismagin: Agreed to replace it with real network code when it's stable

PR #360, opened 2017-06-19
Shutdown, MVStore and block signature validation

phearnot:


src/main/scala/com/wavesplatform/network/UtxPoolSynchronizer.scala
```diff
@@ -2,12 +2,12 @@ package com.wavesplatform.network
...
 import com.wavesplatform.state2.diffs.TransactionDiffer.TransactionValidationError
 import io.netty.channel.ChannelHandler.Sharable
-import io.netty.channel.{ChannelHandlerContext, ChannelInboundHandlerAdapter}
+import io.netty.channel.{Channel, ChannelHandlerContext, ChannelInboundHandlerAdapter}
 import scorex.transaction.{NewTransactionHandler, Transaction}
 import scorex.utils.ScorexLogging
 
 @Sharable
-class UtxPoolSynchronizer(handler: NewTransactionHandler, network: Network)
+class UtxPoolSynchronizer(handler: NewTransactionHandler, broadcast: (AnyRef, Option[Channel]) => Unit)
```
ismagin: Please restore explicit dependency

phearnot: `Network` used to have two methods, the second one being [`requestExtendsion`](https://github.com/wavesplatform/Waves/pull/360/commits/90d65bc6438ed923a75babb96923442d5f72f41e#diff-6310bdb4f3377f310c6f85116e66dd6cL64), which was not used anywhere, actually. While can see the point in explicitly passing complex interfaces like `HistoryReader`, I see no point keeping a SAM interface here.

PR #363, opened 2017-06-20
Failure to establish initial connections was fixed

alexeykiselev:
Bug then broadcast of a new block was fixed


src/main/scala/com/wavesplatform/Coordinator.scala
```diff
@@ -134,6 +142,13 @@ class Coordinator(
       Left(GenericError("Checkpoint already applied"))
     }
 
+  def processRollback(blockId: ByteStr): Either[ValidationError, BigInt] = {
```
ismagin: Wouldn't `.right.map(_=>history.score())` do the job? if score is needed, it can be moved deeper into `blockchainUpdater.removeAfter`

PR #365, opened 2017-06-21
Debug state hash calculates reproducible account portfolios hash

Tolsi:


src/main/scala/com/wavesplatform/state2/reader/StateReader.scala
```diff
@@ -28,6 +27,8 @@ trait StateReader extends Synchronized {
 
   def height: Int
 
+  def accountPortfoliosHash: Int
```
ismagin: please make this extension method

PR #368, opened 2017-06-26
Transaction generator it test

Tolsi:


generator/src/main/scala/com.wavesplatform.generator/TestnetTransactionsGeneratorTest.scala
```diff
@@ -0,0 +1,46 @@
...
+import scorex.transaction.TransactionParser
+import scorex.transaction.TransactionParser.{TransactionType => TT}
+import scorex.transaction.assets.exchange.AssetPair
+import scorex.utils.LoggerFacade
+
+import scala.concurrent.Await
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.duration._
+
+object TestnetTransactionsGeneratorTest extends App {
```
ismagin: Please rename to `TransactionsGeneratorApp` or similar

PR #371, opened 2017-06-27
Rewritten UtxPool

phearnot:


src/main/scala/com/wavesplatform/UtxPool.scala
```diff
@@ -0,0 +1,90 @@
...
+    TransactionDiffer.apply(fs, time.correctedTime(), stateReader.height)(stateReader, t)
+      .map(_ => t)
+
+  private def collectValidTransactions(currentTs: Long): Seq[Transaction] = {
+    val differ = TransactionDiffer.apply(fs, currentTs, stateReader.height) _
+
+    val (invalidTxs, validTxs, _) = transactions.asScala
+      .values.toSeq
+      .sorted(TransactionsOrdering.InUTXPool)
+      .foldLeft((Seq.newBuilder[ByteStr], ArrayBuffer.empty[Transaction], Monoid[Diff].empty)) {
```
ismagin: Let's return to immutable things here, `prepend` for performance and then `reverse`(if needed, looks like it's not), as it has always been

src/main/scala/com/wavesplatform/UtxPool.scala
```diff
@@ -0,0 +1,90 @@
...
+              (invalid, valid, diff)
+          }
+        case (r, _) => r
+      }
+
+    invalidTxs.result().foreach(transactions.remove)
+    validTxs.sorted(TransactionsOrdering.InBlock)
+  }
+
+  def putIfNew(tx: Transaction, source: Option[Channel] = None): Either[ValidationError, Transaction] = {
```
ismagin: All usages of this show that it's either one-arg call or `None`, so this signature is either an error or an obsolete.

src/test/scala/com/wavesplatform/matcher/market/OrderBookActorSpecification.scala
```diff
@@ -90,10 +90,10 @@ class OrderBookActorSpecification extends TestKit(ActorSystem("MatcherTest"))
     val functionalitySettings = stub[FunctionalitySettings]
 
     actor = system.actorOf(Props(new OrderBookActor(pair, orderHistoryRef, storedState,
-      wallet, settings, history, functionalitySettings, new EmbeddedChannel()) with RestartableActor {
-      override def validate(orderMatch: ExchangeTransaction): Either[ValidationError, SignedTransaction] = Right(orderMatch)
-
-      override def sendToNetwork(tx: SignedTransaction): Either[ValidationError, SignedTransaction] = Right(tx)
+      wallet, stub[UtxPool], settings, history, functionalitySettings) with RestartableActor {
+//      override def validate(orderMatch: ExchangeTransaction): Either[ValidationError, SignedTransaction] = Right(orderMatch)
```
ismagin: please clenup

src/test/scala/com/wavesplatform/matcher/market/OrderBookActorSpecification.scala
```diff
@@ -276,13 +276,13 @@ class OrderBookActorSpecification extends TestKit(ActorSystem("MatcherTest"))
...
       actor = system.actorOf(Props(new OrderBookActor(pair, orderHistoryRef, storedState,
-        wallet, settings, history, functionalitySettings, new EmbeddedChannel()) with RestartableActor {
-        override def validate(orderMatch: ExchangeTransaction): Either[ValidationError, SignedTransaction] = {
-          if (orderMatch.buyOrder == ord2) Left(ValidationError.GenericError("test"))
-          else Right(orderMatch)
-        }
-
-        override def sendToNetwork(tx: SignedTransaction): Either[ValidationError, SignedTransaction] = Right(tx)
+        wallet, stub[UtxPool], settings, history, functionalitySettings) with RestartableActor {
+//        override def validate(orderMatch: ExchangeTransaction): Either[ValidationError, SignedTransaction] = {
```
ismagin: please cleanup

src/test/scala/scorex/transaction/state/database/UtxPoolSpecification.scala
```diff
@@ -0,0 +1,54 @@
...
+
+      val db: UtxPool = new UtxPool(stub[ChannelGroup], stub[Time],
+        stub[StateReader], new FeeCalculator(FeesSettings(Map.empty)), FunctionalitySettings.TESTNET, smallSizedUTXSettings)
+
+
+      db.putIfNew(newTx(1)) shouldBe 'right
+
+      db.all() should have size 1
+
+      db.putIfNew(newTx(2)) shouldBe 'left
```
ismagin: I suggest we use `should produce("error text")`. This would ensure we get expected error, not some mock-related one or whatnot.

src/main/scala/com/wavesplatform/UtxPool.scala
```diff
@@ -0,0 +1,90 @@
...
+      _ = transactions.put(t.id, t)
+      _ = allChannels.broadcast(t, source)
+    } yield t
+  }
+
+  def remove(tx: Transaction): Unit = transactions.remove(tx.id)
+
+  def all(): Seq[Transaction] = transactions.values.asScala.toSeq
+
+  def getBySignature(signature: ByteStr): Option[Transaction] = Option(transactions.get(signature))
```
ismagin: never used

src/main/scala/com/wavesplatform/UtxPool.scala
```diff
@@ -0,0 +1,90 @@
...
+
+class UtxPool(
+    allChannels: ChannelGroup,
+    time: Time,
+    stateReader: StateReader,
+    feeCalculator: FeeCalculator,
+    fs: FunctionalitySettings,
+    utxSettings: UtxSettings) extends ScorexLogging {
+
+  private val transactions = new ConcurrentHashMap[ByteStr, Transaction](utxSettings.maxSize / 2)
```
ismagin: `scala.concurrent.TrieMap` works too, no need for `.asScala`

src/main/scala/com/wavesplatform/network/UtxPoolSynchronizer.scala
```diff
@@ -1,34 +1,27 @@
...
+  override def channelRead(ctx: ChannelHandlerContext, msg: AnyRef) = msg match {
+    case t: Transaction => utx.putIfNew(t) match {
+      case Left(TransactionValidationError(_, e)) =>
+        log.debug(s"${id(ctx)} Error processing transaction ${t.id}: $e")
       case Left(e) =>
         log.debug(s"${id(ctx)} Error processing transaction ${t.id}: $e")
       case Right(_) =>
         log.debug(s"${id(ctx)} Added transaction ${t.id} to UTX pool")
-        allChannels.broadcast(RawBytes(TransactionalMessagesRepo.TransactionMessageSpec.messageCode, t.bytes), remoteSource)
+        allChannels.broadcast(RawBytes(TransactionalMessagesRepo.TransactionMessageSpec.messageCode, t.bytes), Some(ctx.channel()))
```
ismagin: (!)looks like tx is broadcasted twice: from `utx.putIfNew(t)` and here

src/test/scala/scorex/transaction/state/database/UtxPoolSpecification.scala
```diff
@@ -0,0 +1,54 @@
...
+        stub[StateReader], new FeeCalculator(FeesSettings(Map.empty)), FunctionalitySettings.TESTNET, UtxSettings(10, 1.minute))
+
+      db.all() shouldBe empty
+
+      db.putIfNew(newTx(1)) shouldBe 'left
+      db.all() shouldBe empty
+    }
+  }
+
+  private def newTx(id: Long) = GenesisTransaction.create(PublicKeyAccount(Array.fill(32)(0: Byte)), id, 4598723454L).right.get
```
ismagin: is it id? looks like it's amount

src/main/scala/com/wavesplatform/UtxPool.scala
phearnot: `Seq.newBuilder` does exactly that: prepends new head to a cached list, and then reverses it when `result()` is called. You just don't need to do it yourself. As for `ArrayByffer` of transactions, I need to check its length, so using `Seq` builder here and reverse and then traverse it just to count elements seems wasteful.

src/main/scala/com/wavesplatform/UtxPool.scala
phearnot: In `removeExpired`, I use `ConcurrentHashMap`'s `entrySet().removeIf(tx => ...)`, which is quite handy.

src/test/scala/scorex/transaction/state/database/UtxPoolSpecification.scala
phearnot: No idea, this was the part of an old test.

src/main/scala/com/wavesplatform/UtxPool.scala
ismagin: ok

src/main/scala/com/wavesplatform/UtxPool.scala
ismagin: The goal is to use immutable structures where possible.

src/main/scala/com/wavesplatform/Application.scala
```diff
@@ -58,13 +57,14 @@ class Application(val actorSystem: ActorSystem, val settings: WavesSettings) ext
...
 
     val peerDatabase = new PeerDatabaseImpl(settings.networkSettings)
     val establishedConnections = new ConcurrentHashMap[Channel, PeerInfo]
     val allChannels = new DefaultChannelGroup(GlobalEventExecutor.INSTANCE)
 
+    val utxStorage = new UtxPool(allChannels,
+      time, stateReader, feeCalculator, settings.blockchainSettings.functionalitySettings, settings.utxSettings)
+
     val network = new NetworkServer(
       settings.blockchainSettings.addressSchemeCharacter,
```
ismagin: please strip down first 4 dependencies to just `settings`

ismagin:
Currently merged into NG, not sure if we still need this?

src/test/scala/com/wavesplatform/UtxPoolSpecification.scala
```diff
@@ -0,0 +1,118 @@
...
+    "does not add new transactions when full" in utxTest(UtxSettings(1, 5.seconds)) { (txs, utx, _) =>
+      expectBroadcast(txs.head)
+      utx.putIfNew(txs.head) shouldBe 'right
+      all(txs.tail.map(t => utx.putIfNew(t))) shouldBe 'left
+    }
+
+    "does not add the same transaction twice" in utxTest() { (txs, utx, _) =>
+      expectBroadcast(txs.head)
+      utx.putIfNew(txs.head) shouldBe 'right
+      utx.putIfNew(txs.head) shouldEqual Left(GenericError(s"Transaction ${txs.head.id} already in the pool"))
```
ismagin: Consider `should produde("already in the pool") as less strict and easier to assert

PR #376, opened 2017-06-29
Synchronization process tweaks

alexeykiselev:
Log messages improvement


src/main/scala/com/wavesplatform/mining/Miner.scala
```diff
@@ -110,7 +110,7 @@ class Miner(
           })
 
           thisAttempt
-        }
+        } else null
```
ismagin: :O

PR #386, opened 2017-07-10
Utx pool diff aggregation fix

Tolsi:


src/test/scala/com/wavesplatform/UtxPoolSpecification.scala
```diff
@@ -125,5 +142,18 @@ class UtxPoolSpecification extends FreeSpec
       utx.packUnconfirmed() shouldBe 'empty
       utx.all() shouldBe 'empty
     }
+
+    "evicts one of invalid together transactions when packUnconfirmed is called" in forAll(onlyOneValidFromManyTransfers) { case (utx, time, txs, offset) =>
```
phearnot: "evicts one of *mutually invalid* transactions when packUnconfirmed is called"

src/test/scala/com/wavesplatform/UtxPoolSpecification.scala
```diff
@@ -50,6 +51,22 @@ class UtxPoolSpecification extends FreeSpec
...
+    (sender, senderBalance, state) <- stateGen
+    recipient <- accountGen
+    fee <- chooseNum(1, (senderBalance * 0.01).toLong)
+    offset <- chooseNum(1000L, 2000L)
+  } yield {
+    val time = new TestTime(System.currentTimeMillis())
+    val utx = new UtxPool(group, time, state, calculator, FunctionalitySettings.TESTNET, UtxSettings(10, 10.minutes))
+    val amountPart = (senderBalance - fee) / 2 - fee
+    val txs = Seq(
+      PaymentTransaction.create(sender, recipient, amountPart, fee, time.getTimestamp() + 1).right.get,
```
phearnot: Check out `TestTime`: it should emit unique timestamps from `getTimestamp`. If it does not, fix it there, not in this test. I also suggest using Transfer transaction.

src/test/scala/com/wavesplatform/UtxPoolSpecification.scala
```diff
@@ -125,5 +142,18 @@ class UtxPoolSpecification extends FreeSpec
...
+    "evicts one of invalid together transactions when packUnconfirmed is called" in forAll(onlyOneValidFromManyTransfers) { case (utx, time, txs, offset) =>
+      all(txs.map { t =>
+        expectBroadcast(t)
+        utx.putIfNew(t)
+      }) shouldBe 'right
+      utx.all().size shouldEqual txs.size
+
+      time.advance(offset)
+
+      utx.packUnconfirmed().size shouldBe 2
```
ismagin: 2 are packed and 2 are left? why?

src/test/scala/com/wavesplatform/UtxPoolSpecification.scala
```diff
@@ -50,6 +51,22 @@ class UtxPoolSpecification extends FreeSpec
   } yield TransferTransaction.create(None, sender, recipient, amount, time.getTimestamp(), None, fee, Array.empty[Byte]).right.get)
     .label("transferTransaction")
 
+  private def onlyOneValidFromManyTransfers = (for {
+    (sender, senderBalance, state) <- stateGen
+    recipient <- accountGen
+    fee <- chooseNum(1, (senderBalance * 0.01).toLong)
+    offset <- chooseNum(1000L, 2000L)
+  } yield {
+    val time = new TestTime(System.currentTimeMillis())
```
phearnot: Looks like `TestTime` is always initialized to current time in this test, so it's probably a good idea to change default constructor. Please also check other tests where `TestTime` is used.

src/test/scala/com/wavesplatform/UtxPoolSpecification.scala
```diff
@@ -125,5 +140,18 @@ class UtxPoolSpecification extends FreeSpec
       utx.packUnconfirmed() shouldBe 'empty
       utx.all() shouldBe 'empty
     }
+
+    "evicts one of invalid together transactions when packUnconfirmed is called" in forAll(twoOutOfManyValidPayments) { case (utx, time, txs, offset) =>
```
phearnot: "evicts one of *mutually invalid* transactions when packUnconfirmed is called"

PR #404, opened 2017-07-17
Don't add peer to list of candidates twice, ignore unsolicited peers

phearnot:


src/main/scala/com/wavesplatform/network/AttributeKeys.scala
```diff
@@ -7,4 +7,5 @@ import io.netty.util.AttributeKey
 object AttributeKeys {
   val NodeName = AttributeKey.newInstance[String]("node-name")
   val DeclaredAddress = AttributeKey.newInstance[InetSocketAddress]("declared-address")
+  val PeersRequested = AttributeKey.newInstance[Boolean]("peers-requested")
```
ismagin: `NodeName` is not used
`DeclaredAddress` and `PeersRequested` are used only in PeerSynchronizer, we can achieve the same just making them `var`s and dropping `Shareable`

PR #405, opened 2017-07-17
Node 52 do not broadcast txs on rollback

Tolsi:


src/main/scala/com/wavesplatform/UtxPool.scala
```diff
@@ -38,15 +38,15 @@ class UtxPool(
   private def removeExpired(currentTs: Long): Unit =
     transactions.entrySet().removeIf(tx => (currentTs - tx.getValue.timestamp).millis > utxSettings.maxTransactionAge)
 
-  def putIfNew(tx: Transaction, source: Option[Channel] = None): Either[ValidationError, Transaction] =
+  def putIfNew(tx: Transaction, source: Option[Channel] = None, broadcast: Boolean = true): Either[ValidationError, Transaction] =
```
ismagin: `source` parameter doesn't make any sense when `broadcast` is false
Not only that, but the whole `putIfNew` signature now takes responsibility and details on what should happen after.
I suugest we remove `allChannels` from `UtxPool` and just the broadcasting later

src/test/scala/com/wavesplatform/matcher/market/OrderBookActorSpecification.scala
```diff
@@ -274,7 +274,7 @@ class OrderBookActorSpecification extends TestKit(ActorSystem("MatcherTest"))
       val ord3 = sell(pair, 100, 10*Order.PriceConstant)
 
       val pool = stub[UtxPool]
-      (pool.putIfNew _).when(*, *).onCall { (tx, _) =>
+      (pool.putIfNew _).when(*, *, *).onCall { (tx, _, _) =>
```
ismagin: if we do as suggested, you won't need to write tests :)

ismagin:
Don't merge until tests are green, some serialization error is still there

ismagin:
Ok, just wait for test to become green

PR #426, opened 2017-07-26
Alias validation agains new rules

alexeykiselev:


src/main/scala/scorex/account/Alias.scala
```diff
@@ -21,20 +21,23 @@ object Alias {
...
   private def buildAlias(networkByte: Byte, name: String): Either[ValidationError, Alias] = {
 
     case class AliasImpl(networkByte: Byte, name: String) extends Alias
 
-    if (name contains "\n") {
-      Left(GenericError("Alias cannot be multiline"))
-    } else if (name.trim() != name) {
-      Left(GenericError("Alias cannot contain leading or trailing whitespaces"))
-    } else if (!(MinLength to MaxLength contains name.length))
+    if (!(MinLength to MaxLength contains name.length))
```
phearnot: Why not just `name.length < MinLength || MaxLength <  name.length`?

PR #442, opened 2017-08-08
NODE-73: Get account balance after application of transactions in UTX pool

vsuharnikov:
You can check a portfolio (including the balance) through a debug API-method:
```
/debug/portfolios/{address}?considerUnspent=true|false
```
, but don't forget to provide an API key ;)

* considerUnspent=*true* - including transactions from UTX pool;
* considerUnspent=*false* - without it.

### DIFF
--- a/src/it/scala/com/wavesplatform/it/api/NodeApi.scala
+++ b/src/it/scala/com/wavesplatform/it/api/NodeApi.scala
@@ -5,6 +5,7 @@ import java.util.concurrent.TimeoutException
 
 import com.wavesplatform.it.util._
 import com.wavesplatform.matcher.api.CancelOrderRequest
+import com.wavesplatform.state2.Portfolio
 import io.netty.util.{HashedWheelTimer, Timer}
 import org.asynchttpclient.Dsl.{get => _get, post => _post}
 import org.asynchttpclient._
@@ -12,12 +13,14 @@ import org.asynchttpclient.util.HttpConstants
 import org.slf4j.LoggerFactory
 import play.api.libs.json.Json.{parse, stringify, toJson}
 import play.api.libs.json._
+import scorex.account.Address
 import scorex.api.http.alias.CreateAliasRequest
 import scorex.api.http.assets._
 import scorex.api.http.leasing.{LeaseCancelRequest, LeaseRequest}
 import scorex.transaction.assets.exchange.Order
 import scorex.utils.{LoggerFacade, ScorexLogging}
 import scorex.waves.http.RollbackParams
+import scorex.waves.http.DebugApiRoute.portfolioFormat
 
 import scala.compat.java8.FutureConverters._
 import scala.concurrent.ExecutionContext.Implicits.global
@@ -108,6 +111,12 @@ trait NodeApi {
     case Failure(UnexpectedStatusCodeException(_, r)) if r.getStatusCode == 404 => Success(None)
     case Failure(ex) => Failure(ex)
   }, tOpt => tOpt.exists(_.id == txId), 1.second).map(_.get)
+
+  def waitForUtxIncreased(fromSize: Int): Future[Int] = waitFor[Int](
+    utxSize,
+    _ > fromSize,
+    100.millis
+  )
 
   def waitForHeight(expectedHeight: Int): Future[Int] = waitFor[Int](height, h => h >= expectedHeight, 1.second)
 
@@ -249,6 +258,13 @@ trait NodeApi {
   def waitForDebugInfoAt(height: Long): Future[DebugInfo] = waitFor[DebugInfo](get("/debug/info").as[DebugInfo], _.stateHeight >= height, 1.seconds)
 
   def debugStateAt(height: Long): Future[Map[String, Long]] = get(s"/debug/stateWaves/$height").as[Map[String, Long]]
+
+  def debugPortfoliosFor(address: Address) = retrying {
+    _get(s"http://$restAddress:$nodeRestPort/debug/portfolios/${address.address}?considerUnspent=true")
+    .setHeader("api_key", "integration-test-rest-api")
+    .build()
+  }.as[Portfolio]
+
 }
 
 object NodeApi extends ScorexLogging {
--- a/src/it/scala/com/wavesplatform/it/api/NodeApi.scala
+++ b/src/it/scala/com/wavesplatform/it/api/NodeApi.scala
@@ -59,6 +59,12 @@ trait NodeApi {
   def get(path: String, f: RequestBuilder => RequestBuilder = identity): Future[Response] =
     retrying(f(_get(s"http://$restAddress:$nodeRestPort$path")).build())
 
+  def getDebug(path: String, f: RequestBuilder => RequestBuilder = identity): Future[Response] = retrying {
+    _get(s"http://$restAddress:$nodeRestPort$path")
+      .setHeader("api_key", "integration-test-rest-api")
+      .build()
+  }
+
   def post(url: String, port: Int, path: String, f: RequestBuilder => RequestBuilder = identity): Future[Response] =
     retrying(f(
       _post(s"$url:$port$path").setHeader("api_key", "integration-test-rest-api")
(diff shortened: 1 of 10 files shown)


src/it/scala/com/wavesplatform/it/api/NodeApi.scala
```diff
@@ -56,6 +59,12 @@ trait NodeApi {
   def get(path: String, f: RequestBuilder => RequestBuilder = identity): Future[Response] =
     retrying(f(_get(s"http://$restAddress:$nodeRestPort$path")).build())
 
+  def getDebug(path: String, f: RequestBuilder => RequestBuilder = identity): Future[Response] = retrying {
```
ismagin: `getWithApiKey` sounds more reasonable

src/main/scala/com/wavesplatform/UtxPool.scala
```diff
@@ -2,22 +2,25 @@ package com.wavesplatform
...
-import scorex.transaction.{FeeCalculator, History, Transaction, ValidationError}
+import scorex.transaction._
 import scorex.utils.{ScorexLogging, Time}
 
 import scala.collection.JavaConverters._
 import scala.concurrent.duration._
 import scala.util.{Left, Right}
 
-
+/**
```
ismagin: Please remove unnnessary comments

src/main/scala/com/wavesplatform/UtxPool.scala
```diff
@@ -26,21 +29,55 @@ class UtxPool(
...
+
   private val transactions = new ConcurrentHashMap[ByteStr, Transaction]
+
   private lazy val knownTransactions = CacheBuilder.newBuilder()
     .maximumSize(utxSettings.maxSize * 2)
     .build[ByteStr, Either[ValidationError, Transaction]]()
 
-  private def removeExpired(currentTs: Long): Unit =
-    transactions.entrySet().removeIf(tx => (currentTs - tx.getValue.timestamp).millis > utxSettings.maxTransactionAge)
+  /**
```
ismagin: The best comment is readable code :)

vsuharnikov:
```scala
private val txPessimisticPortfolios = new ConcurrentHashMap[ByteStr, Portfolios]
```
I know, but it is hard to name this variable, i did my best. The comments here is for clearing things. 

If someone has an idea about naming, you're welcome :)

src/main/scala/com/wavesplatform/state2/Portfolio.scala
```diff
@@ -7,6 +7,20 @@ import cats.Monoid
...
   lazy val spendableBalance: Long = balance - leaseInfo.leaseOut
+
+  /**
+    * The Portfolio with only the withdraw part
+    */
+  def pessimistic: Portfolio = Portfolio(
+    balance = Math.min(balance, 0),
+    leaseInfo = LeaseInfo(
+      leaseIn = 0,
+      leaseOut = Math.min(leaseInfo.leaseOut, 0)
```
ismagin: Shouldn't this be `Math.max(leaseInfo.leaseOut, 0)`
If you lease something, it's positive(should be included), if you get your lease back, it's negative(should not be included)

src/main/scala/com/wavesplatform/UtxPool.scala
```diff
@@ -26,21 +26,55 @@ class UtxPool(
...
       Left(GenericError("Transaction pool size limit is reached"))
     } else knownTransactions.get(tx.id, () => {
       val validationResult = for {
         _ <- feeCalculator.enoughFee(tx)
-        _ <- TransactionDiffer.apply(fs, history.lastBlock.map(_.timestamp), time.correctedTime(), stateReader.height)(stateReader, tx)
+        diff <- TransactionDiffer(fs, history.lastBlock.map(_.timestamp), time.correctedTime(), stateReader.height)(stateReader, tx)
+        _ = {
+          val nonEmptyPessimisticPortfolios = diff.portfolios
+            .mapValues(_.pessimistic)
+            .filterNot {
```
ismagin: Combining with empty does nothing, please simplify

src/main/scala/com/wavesplatform/UtxPool.scala
```diff
@@ -55,6 +89,23 @@ class UtxPool(
     }
   }
 
+  def portfolio(addr: Address): Portfolio = {
```
ismagin: Here you reduce ALL transaction diffs on every request.
Condider `Map[Address,List[TransactionId]]` to be able to fold only nessessary txs.

src/main/scala/com/wavesplatform/state2/Portfolio.scala
```diff
@@ -7,6 +7,20 @@ import cats.Monoid
...
+    ),
+    assets = assets.filter { case (_, v) => v < 0 }
+  )
+
+  def isEmpty: Boolean = this == Monoid.empty[Portfolio]
```
ismagin: Not needed

src/main/scala/scorex/waves/http/DebugApiRoute.scala
```diff
@@ -74,6 +76,40 @@ case class DebugApiRoute(settings: RestAPISettings,
...
+      value = "Taking into account pessimistic transactions from UTX pool",
+      required = false,
+      dataType = "boolean",
+      paramType = "query",
+      defaultValue = "true"
+    )
+  ))
+  @ApiResponses(Array(new ApiResponse(code = 200, message = "Json portfolio")))
+  def portfolios: Route = path("portfolios" / Segment) { (rawAddress) =>
+    (get & parameter('considerUnspent.as[Boolean])) { (considerUnspent) =>
```
ismagin: Not sure if anyone needs it, but shouldn't we also include `effectiveBalance` and the needed `spendableBalance`  to the response?

src/test/scala/com/wavesplatform/UtxPoolSpecification.scala
```diff
@@ -49,6 +49,12 @@ class UtxPoolSpecification extends FreeSpec
   } yield TransferTransaction.create(None, sender, recipient, amount, time.getTimestamp(), None, fee, Array.empty[Byte]).right.get)
     .label("transferTransaction")
 
+  private def transferWithRecipient(sender: PrivateKeyAccount, recipient: PrivateKeyAccount, maxAmount: Long, time: Time) = (for {
```
ismagin: for `recipient`, `PublicKeyAccount` is enough.

src/it/scala/com/wavesplatform/it/transactions/debug/DebugPortfoliosSpecification.scala
```diff
@@ -0,0 +1,56 @@
...
+    val firstAddr = Address.fromString(firstAddress).right.get
+
+    val f = for {
+      _ <- assertBalances(firstAddress, 100.waves, 100.waves)
+      _ <- assertBalances(secondAddress, 100.waves, 100.waves)
+
+      portfolioBefore <- sender.debugPortfoliosFor(firstAddr, considerUnspent = true)
+      utxSizeBefore <- sender.utxSize
+
+      _ <- sender.payment(firstAddress, secondAddress, 5.waves, fee = 5.waves)
```
ismagin: Consider  `LeaseTransaction` here. This way we'll check `leaseOut component as well.

src/main/scala/com/wavesplatform/UtxPool.scala
vsuharnikov: With this approach we should go over all addresses in `UtxPool.removeExpired`:
```scala
      expiredTxs.foreach { tx =>
        transactions.remove(tx.id)
        txPessimisticPortfolios.remove(tx.id)
      }
```
but yes, I agree, that `removeExpired` could be executed rarely comparing with `portfolio` .

src/main/scala/com/wavesplatform/UtxPool.scala
vsuharnikov: > Combining with empty does nothing

That's why we filter out all empty portfolios :)
```scala
            .filterNot {
              case (_, portfolio) => portfolio.isEmpty
            }
```

src/main/scala/com/wavesplatform/state2/Portfolio.scala
vsuharnikov: Please, see above

src/main/scala/scorex/waves/http/DebugApiRoute.scala
vsuharnikov: I think, QAs should make this calculation by themselves, it is not too hard. If we really need this, i'd rather create a special API case class and make a [de]serializers for it.

src/it/scala/com/wavesplatform/it/transactions/debug/DebugPortfoliosSpecification.scala
vsuharnikov: I added two tests for `Portfolio.pessimistic` in `PortfolioTest` to check `leaseOut` and other fields. I think, this is enough.

vsuharnikov:
Thanks for helpful comments! Answered on some questions. Also please take a look onto `UtxPool`, a lot of changes are here.

PR #446, opened 2017-08-10
NODE-93: do not rebroadcast existing tx

ismagin:


src/main/scala/com/wavesplatform/matcher/market/OrderBookActor.scala
```diff
@@ -233,7 +233,10 @@ class OrderBookActor(assetPair: AssetPair,
         None
 
       case event@OrderExecuted(o, c) =>
-        createTransaction(o, c).flatMap(utx.putIfNew(_)) match {
+        (for {
+          tx <- createTransaction(o, c)
+          _ <- utx.putIfNew(tx)
+        } yield tx) match {
```
koloale: Do these changes relate to the current task?

ismagin: yes they do

PR #450, opened 2017-08-13
Application and node names bytes representation length fixed

alexeykiselev:


alexeykiselev:
Application name can't be a long string because we get it from Constants plus one char. I've added node name validation while reading it from the configuration file.

PR #452, opened 2017-08-14
NODE-100 Metrics

vsuharnikov:


.gitignore
```diff
@@ -35,3 +35,4 @@ Dockerfile
 
 package
 !src/package
+native
```
vsuharnikov: Is added by kamon, so we should ignore it in the repository

PR #459, opened 2017-08-16
NODE-104 StackOverflow error in UtxPool and other possible places

vsuharnikov:


src/main/scala/com/wavesplatform/state2/Diff.scala
```diff
@@ -54,8 +54,12 @@ case class Diff(transactions: Map[ByteStr, (Int, Transaction, Set[Address])],
       m.combine(Map(acc -> set))
     }
     groupedByAcc
```
phearnot: To me, this looks like the only possible cause of `StackOverflowError`. I see no point in changing other occurrences, because `mapValues` result is never cached.

phearnot: Never mind, apparently this is just the debug handle.

vsuharnikov:
@phearnot The problems with `filterKeys`/`mapValues` are:
1. They have a bad name
2. They have a bad signature

Also there is a `view` in collections, that creates a lazy version of a given collection. So, you did not expect, that `filterKeys`/`mapValues` work lazy. 

So from my point of view, it is better to not use this methods.

PR #469, opened 2017-08-28
NODE-124: Measure forks frequency and microblock receiving lag

vsuharnikov:


src/main/scala/com/wavesplatform/network/MicroBlockSynchronizer.scala
```diff
@@ -14,6 +18,8 @@ import scala.concurrent.duration.FiniteDuration
 class MicroBlockSynchronizer(settings: Settings, history: NgHistory)
   extends ChannelInboundHandlerAdapter with ScorexLogging {
 
+  private val microBlockCreationTime = new ConcurrentHashMap[ByteStr, Long]()
```
ismagin: doesn't get cleaned if poluted from malicious node

src/main/scala/com/wavesplatform/network/messages.scala
```diff
@@ -17,7 +17,7 @@ case class GetBlock(signature: ByteStr) extends Message
 case class LocalScoreChanged(newLocalScore: History.BlockchainScore) extends Message
 case class RawBytes(code: Byte, data: Array[Byte]) extends Message
 case class BlockForged(block: Block) extends Message
-case class MicroBlockInv(totalBlockSig: ByteStr, prevBlockSig: ByteStr)  extends Message
+case class MicroBlockInv(totalBlockSig: ByteStr, prevBlockSig: ByteStr, created: Long = System.currentTimeMillis()) extends Message
```
ismagin: No reason to use default values

src/main/scala/scorex/transaction/NgHistoryWriter.scala
```diff
@@ -17,7 +18,7 @@ trait NgHistoryWriter extends HistoryWriter with NgHistory {
 
   def bestLiquidBlock(): Option[Block]
 
-  def forgeBlock(id: BlockId): Option[(Block, DiscardedTransactions)]
+  def forgeBlock(id: BlockId): Option[(Block, Int, DiscardedTransactions)]
```
ismagin: It's better to use `Option[(Block, DiscardedMicroblocks)]`

src/main/scala/com/wavesplatform/job/ParallelJobPoolAbstractDecorator.scala
```diff
@@ -0,0 +1,10 @@
+package com.wavesplatform.job
+
+abstract class ParallelJobPoolAbstractDecorator[ItemT, GroupIdT, JobT <: Runnable](orig: ParallelJobPool[ItemT, GroupIdT, JobT])
```
ismagin: Don't forget to remove this on merge

src/main/scala/com/wavesplatform/network/messages.scala
vsuharnikov: There is no reason to write `System.currentTimeMillis()` everywhere

vsuharnikov:
The attackers should send the data a long time :) But ok, i'll fix it

src/main/scala/com/wavesplatform/network/MicroBlockSynchronizer.scala
vsuharnikov: Ok

src/main/scala/scorex/transaction/NgHistoryWriter.scala
vsuharnikov: Ok

src/main/scala/com/wavesplatform/job/ParallelJobPoolAbstractDecorator.scala
vsuharnikov: Done :)

PR #474, opened 2017-08-29
NODE-128 Generator sends transactions to multiple nodes

vsuharnikov:


vsuharnikov:
@ismagin So, what's with this PR?

PR #478, opened 2017-08-31
Adding `fees` section :boom:

sigwo:
Adding this section for adding very popular tokens being used for all different types of transactions.


ismagin:
We are not adding any particular asset as default acceptable fee for mainnet

PR #486, opened 2017-09-05
NODE-143

ebceu4:


src/main/scala/com/wavesplatform/mining/Miner.scala
```diff
@@ -99,17 +100,20 @@ class Miner(
...
       Right(None)
-    } else {
+    }
+    else {
       log.trace(s"Accumulated ${unconfirmed.size} txs for microblock")
       (for {
         fullAndMicro <- measureSuccessful(microBlockBuildTimeStats, {
-          val signedBlock = Block.buildAndSign(version = 3,
+
+          val eitherBlock = Block.buildAndSign(version = 3,
```
ismagin: for {
signedBlock <- Block.buildAndSign
microBlock <- MicroBlock.buildAndSign
} yield (signedBlock, microBlock)

PR #491, opened 2017-09-08
NODE-153 new block version 3 from ng

ebceu4:


src/main/scala/scorex/block/Block.scala
```diff
@@ -90,32 +90,42 @@ case class Block(timestamp: Long, version: Byte, reference: ByteStr, signerData:
...
+    if (bytes.isEmpty) {
+      Seq.empty
+    } else {
+      val v: (Array[Byte], Int) = version match {
+        case 1 | 2 => (bytes.tail, bytes.head) //  127 max, won't work properly wif greater
+        case 3 =>
+          // https://stackoverflow.com/a/18247942/288091
+          val high = if (bytes(1) >= 0) bytes(1) else 256 + bytes(1)
+          val low = if (bytes(0) >= 0) bytes(0) else 256 + bytes(0)
+          val amount = low | (high << 8)
```
vsuharnikov: Why little endian? We already use `com.google.common.primitives.Ints/Shorts/...` (e.g. see `scorex.serialization`) which works with big endian.

PR #494, opened 2017-09-12
Node 154 supported changes data field

ebceu4:


src/main/scala/scorex/block/Block.scala
```diff
@@ -18,15 +18,21 @@ import scorex.utils.ScorexLogging
...
 
-case class Block(timestamp: Long, version: Byte, reference: ByteStr, signerData: SignerData,
-                 consensusData: NxtLikeConsensusBlockData, transactionData: Seq[Transaction]) extends Signed {
+case class Block(timestamp: Long,
+                 version: Byte,
+                 reference: ByteStr,
+                 signerData: SignerData,
+                 consensusData: NxtLikeConsensusBlockData,
+                 transactionData: Seq[Transaction],
+                 supportedFeaturesIds: Seq[Short] = Seq.empty) extends Signed {
```
ismagin: Please don't use default values in domain case classes

src/test/scala/scorex/lagonaki/unit/BlockSpecification.scala
```diff
@@ -13,27 +14,54 @@ import scala.util.Random
 
 class BlockSpecification extends FunSuite with Matchers with MockFactory {
 
-  test(" block with txs bytes/parse roundtrip") {
+  val reference: Array[Byte] = Array.fill(Block.BlockIdLength)(Random.nextInt(100).toByte)
```
ismagin: This has a lot of magic numbers, please rewrite it using generators so that things would be more random

src/main/scala/scorex/block/Block.scala
```diff
@@ -172,13 +194,15 @@ object Block extends ScorexLogging {
...
                    transactionData: Seq[Transaction],
-                   signer: PrivateKeyAccount): Either[GenericError, Block] = (for {
-    _ <- Either.cond(transactionData.size <= MaxTransactionsPerBlockVer3, (), s"too many transactions in Block: allowed: $MaxTransactionsPerBlockVer3, actual: ${transactionData.size}")
+                   signer: PrivateKeyAccount,
+                   supportedFeaturesIds: Seq[Short] = Seq.empty): Either[GenericError, Block] = (for {
+    _ <- Either.cond(transactionData.size <= MaxTransactionsPerBlockVer3, (), s"Too many transactions in Block: allowed: $MaxTransactionsPerBlockVer3, actual: ${transactionData.size}")
     _ <- Either.cond(reference.arr.length == SignatureLength, (), "Incorrect reference")
     _ <- Either.cond(consensusData.generationSignature.length == GeneratorSignatureLength, (), "Incorrect consensusData.generationSignature")
     _ <- Either.cond(signer.publicKey.length == KeyLength, (), "Incorrect signer.publicKey")
+    _ <- Either.cond(version > 2 || supportedFeaturesIds.isEmpty, (), s"Block version $version could not contain supported feature flags")
```
ismagin: I strongly suggest we move this from run-time either to compile-time error
Wouldn't need to write test, understanding and parsing would be more clear
```
st Version
co Version1 extends Version
co Version2 extends Version
cc Version3(features: Seq[Short]) extends Version
```

src/main/scala/scorex/block/Block.scala
alexeykiselev: May be we should use **Set**[Short] here?

src/main/scala/scorex/block/Block.scala
alexeykiselev: We should check somewhere that the length of features set is less than maximum allowed value. Maximum allowed features should be considered itself. I suggest 64.

ismagin:
`feautureFlags: Seq[Short]` is very raw; let's stick to initial idea of `SupportedChangesDataField` with its way to serialize depending on version passed(as in `TransactionDataField`). 
This will reduce `if (version > 2)` and create less mess in the future

PR #498, opened 2017-09-15
NODE-156 update block creation process

ebceu4:


src/main/scala/com/wavesplatform/features/FeatureStatus.scala
```diff
@@ -0,0 +1,13 @@
+package com.wavesplatform.features
+
+sealed case class FeatureStatus(status: Byte){
```
ismagin: I was able to create `FeatureStatus(5:Byte)` which makes no sense.
please rewrite to 
`sealed trait FeatureStatus`, `case class`es implementing and 
```
object FeatureStatus {
def apply(b: Byte) : Either[...,FeatureStatus]
}
```

PR #500, opened 2017-09-15
NODE-172: extract ng-state to separate entity

ismagin:
All ng-related state moved to `NgState` class.


src/main/scala/com/wavesplatform/state2/BlockchainUpdaterImpl.scala
```diff
@@ -76,112 +72,153 @@ class BlockchainUpdaterImpl private(persisted: StateWriter with StateReader,
...
-          settings, currentPersistedBlocksState, ngHistoryWriter.lastBlock, block)).map { case ((newBlockDiff, discraded)) =>
-          liquidBlockCandidatesDiff.set(Map(block.uniqueId -> newBlockDiff))
-          discraded
+        lazy val logDetails = s"The referenced block(${block.reference})" +
+          s" ${if (historyWriter.contains(block.reference)) "exits, it's not last persisted" else "doesn't exist"}"
+        historyWriter.lastBlock match {
+          case Some(lastInner) if lastInner.uniqueId != block.reference =>
+            Left(BlockAppendError(s"References incorrect or non-existing block: " + logDetails, block))
+          case _ =>
+            historyWriter.lastBlock.map(_.uniqueId) match {
```
vsuharnikov: Looks strange, because we already in the None branch. So, we fall to
```scala
case _ => BlockDiffer.fromBlock(settings,
```
here. What do you think?

src/main/scala/com/wavesplatform/state2/BlockchainUpdaterImpl.scala
```diff
@@ -76,112 +72,153 @@ class BlockchainUpdaterImpl private(persisted: StateWriter with StateReader,
...
+              case Some(lastId) if lastId != block.reference => Left(GenericError(s"Parent ${block.reference} of new liquid block ${block.uniqueId}" +
+                s" does not match last block $lastId"))
+              case _ => BlockDiffer.fromBlock(settings, currentPersistedBlocksState, historyWriter.lastBlock, block).map((_, Seq.empty[Transaction]))
+            }
+        }
+      case Some(ng) if !ng.contains(block.reference) =>
+        Left(BlockAppendError(s"References incorrect or non-existing block", block))
+      case Some(ng) =>
+        val referencedLiquidDiff = ng.diffs.get(block.reference).get
```
vsuharnikov: You can use just `ng.diff(block.reference)` here. But what if there is no such element? We'll get an exception and it will be forwarded to upper levels. Is that correct?

ismagin: Actually we won't, because `.diffs` and `.micros` are in sync, if we didn't go to ` Some(ng) if !ng.contains(block.reference)`, then both values of `ng.diffs` and next line of `ng.forgeBlock(block.reference).get` would be safe.

src/main/scala/scorex/transaction/NgHistoryReader.scala
```diff
@@ -0,0 +1,76 @@
...
+import scorex.block.{Block, MicroBlock}
+import scorex.transaction.History.BlockchainScore
+import scorex.utils.ScorexLogging
+
+class NgHistoryReader(ngState: () => Option[NgState], inner: History) extends History with NgHistory with DebugNgHistory with ScorexLogging with Instrumented {
+
+  override def synchronizationToken: ReentrantReadWriteLock = inner.synchronizationToken
+
+  override def height(): Int = read { implicit l =>
+    inner.height() + ngState().map(_ => 1).getOrElse(0)
```
vsuharnikov: Minor: `ngState().size` ? :)

src/main/scala/scorex/transaction/NgHistoryReader.scala
```diff
@@ -0,0 +1,76 @@
...
+
+  override def blockBytes(height: Int): Option[Array[Byte]] = read { implicit l =>
+    inner.blockBytes(height).orElse(if (height == inner.height() + 1) ngState().map(_.bestLiquidBlock.bytes) else None)
+  }
+
+  override def scoreOf(blockId: BlockId): Option[BlockchainScore] = read { implicit l =>
+    inner.scoreOf(blockId)
+      .orElse(ngState() match {
+        case Some(ng) => Some(inner.score() + ng.base.blockScore)
+        case _ => None
```
vsuharnikov: Hmm.. `ngState().map { ng => inner.score() + ng.base.blockScore }` ?

src/main/scala/scorex/transaction/NgHistoryReader.scala
```diff
@@ -0,0 +1,76 @@
...
+      })
+  }
+
+  override def heightOf(blockId: BlockId): Option[Int] = read { implicit l =>
+    lazy val innerHeight = inner.height()
+    inner.heightOf(blockId).orElse(ngState() match {
+      case Some(ng) if ng.contains(blockId) => Some(innerHeight + 1)
+      case _ => None
+    })
```
vsuharnikov: I doesn't think that the version below is better, but keep in mind we can write something like:
```scala
ngState().filter(_.contains(blockId)).map(_ => innerHeight + 1)
```
;)

src/main/scala/com/wavesplatform/state2/BlockchainUpdaterImpl.scala
ismagin: Will fix `.get(..).get`

src/main/scala/com/wavesplatform/state2/BlockchainUpdaterImpl.scala
ismagin: We are in branch(ng=none, lastblock=none_or_matches_by_id), but you're right, will simplify

src/main/scala/com/wavesplatform/state2/NgState.scala
```diff
@@ -0,0 +1,68 @@
...
+
+case class NgState private(base: Block, diffs: Map[BlockId, BlockDiff], micros: List[MicroBlock]) {
+
+  lazy val lastMicroTotalSig: Option[ByteStr] =
+    micros.headOption.map(_.totalResBlockSig)
+
+  lazy val bestLiquidBlockId: AssetId =
+    lastMicroTotalSig.getOrElse(base.uniqueId)
+
+  lazy val transactions: Seq[Transaction] = base.transactionData ++ micros.map(_.transactionData).reverse.flatten
```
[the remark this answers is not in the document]
ismagin: discussed

src/main/scala/scorex/transaction/NgHistoryReader.scala
ismagin: fixed

src/main/scala/scorex/transaction/NgHistoryReader.scala
ismagin: should be `case Some(ng) if ng.contains(blockId) => Some(inner.score() + ng.base.blockScore)`

PR #509, opened 2017-09-19
NODE-170-fork-parameter-to-activate-feature-approval

ebceu4:


src/main/scala/com/wavesplatform/settings/BlockchainSettings.scala
```diff
@@ -89,11 +93,12 @@ object FunctionalitySettings {
...
+    new FunctionalitySettings(featureCheckBlocksPeriod, blocksForFeatureActivation, allowTemporaryNegativeUntil, allowInvalidPaymentTransactionsByTimestamp, requireSortedTransactionsAfter, generationBalanceDepthFrom50To1000AfterHeight, minimalGeneratingBalanceAfter, allowTransactionsFromFutureUntil, allowUnissuedAssetsUntil, allowBurnTransactionAfter, allowLeaseTransactionAfter, requirePaymentUniqueIdAfter, allowExchangeTransactionAfter, allowInvalidReissueInSameBlockUntilTimestamp, allowCreatealiasTransactionAfter, allowMultipleLeaseCancelTransactionUntilTimestamp, resetEffectiveBalancesAtHeight, allowLeasedBalanceTransferUntil, blockVersion3After)
```
ismagin: please move validation to case class itself:
```
scala> case class X(i:Int) { require(i>0) }
defined class X

scala> X(1)
res0: X = X(1)

scala> X(-1)
java.lang.IllegalArgumentException: requirement failed
  at scala.Predef$.require(Predef.scala:264)
  ... 30 elided
```

src/main/scala/com/wavesplatform/general/BlockVersionResolver.scala
```diff
@@ -0,0 +1,12 @@
+package com.wavesplatform.general
```
ismagin: This is overengineered, we only need 1-line function, so make it function and call where needed

PR #510, opened 2017-09-20
NODE-187 WideTransactionGenerator can limit accounts

vsuharnikov:


generator/src/main/scala/com.wavesplatform.generator/WideTransactionGenerator.scala
```diff
@@ -30,4 +34,12 @@ class WideTransactionGenerator(val accounts: Seq[PrivateKeyAccount]) extends Tra
         if (maybeTransaction.isRight) txs :+ maybeTransaction.right.get else txs
     }
   }
+
```
ismagin: Stream? :)

generator/src/main/scala/com.wavesplatform.generator/WideTransactionGenerator.scala
```diff
@@ -30,4 +34,12 @@ class WideTransactionGenerator(val accounts: Seq[PrivateKeyAccount]) extends Tra
...
+  def randomContinually[A](orig: Seq[A]): Iterator[A] = new Iterator[A] {
```
ismagin: Stream? :)

PR #515, opened 2017-09-21
NODE-189 Generator can make increasing load

vsuharnikov:
Awaits #510 

* The `DynamicWideTransactionGenerator` for making an increasing load;
* A configuration in the Generator is combined with parameters;
* Settings for each generators;
* Fixed IDEA issues with the Generator;
* The full information about mode in Generator is printed at start.


generator/src/main/scala/com.wavesplatform.generator/GeneratorSettings.scala
```diff
@@ -1,77 +1,46 @@
...
-import java.io.File
 import java.net.InetSocketAddress
 
-import com.google.common.base.CaseFormat
-import com.typesafe.config.{Config, ConfigFactory}
-import com.wavesplatform.settings.loadConfig
-import net.ceedubs.ficus.Ficus._
-import net.ceedubs.ficus.readers.ValueReader
-import org.slf4j.LoggerFactory
+import cats.Show
```
ismagin: 👍

PR #523, opened 2017-09-25
NODE-198 Generator should reconnect when the connection was lost

vsuharnikov:


generator/src/main/scala/com.wavesplatform.generator/Worker.scala
```diff
@@ -0,0 +1,107 @@
...
+import scala.concurrent.{ExecutionContext, Future, blocking}
+import scala.util.control.NonFatal
+
+class Worker(settings: Settings,
+             sender: NetworkSender,
+             node: InetSocketAddress,
+             generator: TransactionGenerator)
+            (implicit ec: ExecutionContext) extends ScorexLogging {
+
+  private type Result[T] = EitherT[Future, (Int, Throwable), T]
```
ismagin: 👍

PR #531, opened 2017-09-26
NODE-205: Ng via voting introduced

ismagin:


src/main/scala/com/wavesplatform/state2/diffs/BlockDiffer.scala
```diff
@@ -17,19 +18,22 @@ object BlockDiffer extends ScorexLogging with Instrumented {
 
   def right(diff: Diff): Either[ValidationError, Diff] = Right(diff)
 
-  def fromBlock(settings: FunctionalitySettings, s: StateReader, maybePrevBlock: Option[Block], block: Block): Either[ValidationError, BlockDiff] = {
+  def fromBlock(settings: FunctionalitySettings, featureProvider: FeatureProvider, s: StateReader, maybePrevBlock: Option[Block], block: Block): Either[ValidationError, BlockDiff] = {
     val blockSigner = block.signerData.generator.toAddress
     val stateHeight = s.height
 
+    // height switch is next after activation
+    val ng4060switchHeight = featureProvider.activationHeight(BlockchainFeatures.NG).getOrElse(Int.MaxValue)
```
ismagin: if you are farming a block with `h = ng4060switchHeight`, you get old way, because this means `stateHeight = ng4060switchHeight-1`

if you are farming a block with `h = ng4060switchHeight +1`, it's 40%-block:
stateHeight = ng4060switchHeight , `None, None` go next

for any next block it's 60% of prev block + 40 from each tx

PR #536, opened 2017-09-27
NODE-217 Client prematurely adds a server's channel to the allChannels list

vsuharnikov:


vsuharnikov:
Thanks to @phearnot , made it simpler and safer

PR #539, opened 2017-09-28
Node 215 better last block

ismagin:


src/main/scala/com/wavesplatform/Coordinator.scala
```diff
@@ -87,6 +88,7 @@ object Coordinator extends ScorexLogging with Instrumented {
                          settings: WavesSettings, featureProvider: FeatureProvider)
                         (newBlock: Block, local: Boolean): Either[ValidationError, BigInt] = measureSuccessful(blockProcessingTimeStats, {
     val newScore = for {
+      _ <- Either.cond(history.heightOf(newBlock.reference).exists(_ >= history.height() - 1), (), GenericError("Can process either new top block or current top block's competitor"))
```
vsuharnikov: What if we received the same block? I think we should ignore it and this situation is not rare :)

src/main/scala/com/wavesplatform/Coordinator.scala
```diff
@@ -103,14 +105,23 @@ object Coordinator extends ScorexLogging with Instrumented {
     _ <- blockchainUpdater.processMicroBlock(microBlock)
   } yield utxStorage.removeAll(microBlock.transactionData))
 
+  private def validateEffectiveBalance(fp: FeatureProvider, fs: FunctionalitySettings, block: Block, h: Int)(effectiveBalance: Long): Either[String, Long] =
```
vsuharnikov: `h` of what? Current? Could you give it a more concrete name?

PR #544, opened 2017-10-01
NODE-213 Container logs are written to the files

vsuharnikov:


ismagin:
Why remove `implicit class HistorgramExt(h: Histogram)` ?

PR #549, opened 2017-10-04
NODE-185 Massive transactions test

vsuharnikov:
* Contains various changes in Generator and metrics;
* Contains a new devnet's setup.

### DIFF
--- a/src/main/scala/com/wavesplatform/Coordinator.scala
+++ b/src/main/scala/com/wavesplatform/Coordinator.scala
@@ -39,7 +39,9 @@ object Coordinator extends ScorexLogging with Instrumented {
               b -> appendBlock(
                 checkpoint, history, blockchainUpdater, stateReader, utxStorage, time, settings.blockchainSettings,
                 featureProvider
-              )(b, local = false, extension = true)
+              )(b, local = false).right.map { baseHeight =>
+                BlockStats.applied(b, BlockStats.Source.Ext, baseHeight)
+              }
             }
             .collectFirst { case (b, Left(e)) => b -> e }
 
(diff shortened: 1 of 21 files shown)


src/main/scala/com/wavesplatform/Coordinator.scala
```diff
@@ -129,8 +137,8 @@ object Coordinator extends ScorexLogging with Instrumented {
     height = history.height()
     discardedTxs <- blockchainUpdater.processBlock(block)
   } yield {
-    if (local) BlockStats.mined(block, height) else BlockStats.applied(block, height)
-    TxsInBlockchainStats.record(block.transactionData.size - discardedTxs.size)
+    if (local) BlockStats.mined(block, height)
```
ismagin: we have 2 booleans which come from 2 sources:
1) miner: local = true, extension = false
2) recieved: local = false, extension = false
3) extension: local = false, extension = true

I think those who call this method are able to handle each case by themselves w/o propogating its info down here, the callee can just `.right.map` and do the logging

src/main/scala/com/wavesplatform/Coordinator.scala
```diff
@@ -119,21 +133,19 @@ object Coordinator extends ScorexLogging with Instrumented {
   private def appendBlock(checkpoint: CheckpointService, history: History, blockchainUpdater: BlockchainUpdater,
                           stateReader: StateReader, utxStorage: UtxPool, time: Time, settings: BlockchainSettings,
                           featureProvider: FeatureProvider)
-                         (block: Block, local: Boolean): Either[ValidationError, Unit] = for {
+                         (block: Block, local: Boolean): Either[ValidationError, Int] = for {
```
ismagin: `local` not needed

PR #555, opened 2017-10-04
Saner logging

phearnot:


src/main/scala/com/wavesplatform/state2/BlockchainUpdaterImpl.scala
```diff
@@ -178,31 +176,36 @@ class BlockchainUpdaterImpl private(persisted: StateWriter with StateReader,
...
-            if (bestLiquidState.height != height) {
-              val persistedPlusBottomHeight = persisted.height + bottomMemoryDiff().heightDiff
-              if (height > persistedPlusBottomHeight) {
-                val newTopDiff = unsafeDiffByRange(composite(persisted, () => bottomMemoryDiff()), persistedPlusBottomHeight + 1, height + 1)
-                topMemoryDiff.set(newTopDiff)
-              } else {
-                topMemoryDiff.set(BlockDiff.empty)
-                if (height < persistedPlusBottomHeight)
-                  bottomMemoryDiff.set(unsafeDiffByRange(persisted, persisted.height + 1, height + 1))
+          if (height < historyWriter.height()) {
```
ismagin: If `historyWriter.heightOf(blockId)` matches `Some(height)`, we actually rollback ng-block if it exists. So technically even if `height == historyWriter.height()` this might be a rollback(at least log-wise).

PR #557, opened 2017-10-05
Node 180, 219 it activation protocol

ebceu4:


src/it/resources/template.conf
```diff
@@ -28,9 +28,7 @@ waves {
         allow-leased-balance-transfer-until = 0
         enable-microblocks-after-height = 0
         block-version-3-after = 0
-        pre-activated-features = {
-          2 = 0
-        }
+        pre-activated-features = {}
```
ismagin: NG should be enabled in it-tests, please restore

src/it/scala/com/wavesplatform/it/ActivationTestSuite.scala
```diff
@@ -0,0 +1,17 @@
+package com.wavesplatform.it
+
+import scala.collection.JavaConverters._
+import scala.concurrent.duration._
+import org.scalatest.{BeforeAndAfterAll, FreeSpec, Matchers}
+
+import scala.concurrent.Await
+
+class ActivationTestSuite extends FreeSpec with Matchers with BeforeAndAfterAll with IntegrationNodesInitializationAndStopping {
```
ismagin: What is the point of this test?

src/it/scala/com/wavesplatform/it/ActivationTestSuite.scala
```diff
@@ -0,0 +1,17 @@
...
+  override val docker = new Docker()
+  override val nodes = Docker.NodeConfigs.getConfigList("nodes").asScala.take(2).map(docker.startNode)
+
+  "api consuming example" in {
+    Await.result(nodes.head.waitForHeight(5), 5.minute)
```
ismagin: Use `Future.traverse` to await for all nodes to get to some state

src/main/scala/com/wavesplatform/Application.scala
```diff
@@ -90,7 +91,8 @@ class Application(val actorSystem: ActorSystem, val settings: WavesSettings) ext
       DebugApiRoute(settings.restAPISettings, wallet, stateReader, history, peerDatabase, establishedConnections, blockchainUpdater, allChannels, utxStorage, blockchainDebugInfo, miner),
       WavesApiRoute(settings.restAPISettings, wallet, utxStorage, allChannels, time),
       AssetsApiRoute(settings.restAPISettings, wallet, utxStorage, allChannels, stateReader, time),
-      NodeApiRoute(settings.restAPISettings, () => this.shutdown()),
+      NodeApiRoute(settings.restAPISettings,() => this.shutdown()),
+      ActivationApiRoute(settings.restAPISettings, settings.blockchainSettings.functionalitySettings, settings.featuresSettings, history, featureProvider),
```
ismagin: might as well provide the whole `settings`

waves-nopeers.conf
```diff
@@ -0,0 +1,149 @@
+# Waves node settins
```
ismagin: There must be a better way to broadcast override only `known-peers`. What is this used for?

src/it/scala/com/wavesplatform/it/ActivationTestSuite.scala
ebceu4: api consuming example

src/main/scala/com/wavesplatform/Application.scala
ebceu4: might not ;)

src/it/scala/com/wavesplatform/it/ActivationTestSuite.scala
ebceu4: If you want to test particular node, what's the point to wait for others?

PR #567, opened 2017-10-09
NODE-224: Back merge from 0.7.8

alexeykiselev:


src/main/scala/com/wavesplatform/network/CoordinatorHandler.scala
```diff
@@ -38,88 +38,97 @@ class CoordinatorHandler(checkpointService: CheckpointService,
...
-        val r: Either[Any, Any] = Signed.validateSignatures(m) match {
-          case Right(_) =>
-            Coordinator.processMicroBlock(checkpointService, history, blockchainUpdater, utxStorage)(m)
-              .map { _ =>
-                allChannels.broadcast(MicroBlockInv(m.totalResBlockSig, m.prevResBlockSig), Some(ctx.channel()))
-              }
-          case Left(err) =>
-            peerDatabase.blacklistAndClose(ctx.channel(), err.toString)
-            Left(())
+  private def updateMicroBlockStats(m: MicroBlock): Either[_, Option[Unit]] = processMicroBlock(m) match {
```
ismagin: `updateMicroBlockStats` shouldn't `processMicroBlock`;
is there a way to make it back to `.left.map(..).right.map(..)`?

src/main/scala/com/wavesplatform/network/CoordinatorHandler.scala
```diff
@@ -38,88 +38,97 @@ class CoordinatorHandler(checkpointService: CheckpointService,
...
+        log.debug(success)
+        maybeNewScore.foreach(r)
+      case Success(Left(ve)) =>
+        log.warn(s"$errorPrefix: $ve")
+        peerDatabase.blacklistAndClose(src, s"$errorPrefix: $ve")
+      case Failure(t) => rethrow(errorPrefix, t)
+    }
+  }
+
+  private def rethrow(msg: String, failure: Throwable) = throw new Exception(msg, failure.getCause)
```
ismagin: why not inlined?

PR #568, opened 2017-10-09
NODE-234 storage versioning

ebceu4:


src/main/scala/com/wavesplatform/history/HistoryWriterImpl.scala
```diff
@@ -19,14 +19,15 @@ import scala.util.Try
 
 class HistoryWriterImpl private(file: Option[File], val synchronizationToken: ReentrantReadWriteLock,
                                 functionalitySettings: FunctionalitySettings, featuresSettings: FeaturesSettings)
-  extends History with FeatureProvider with ScorexLogging {
+  extends VariablesStorage(createMVStore(file)) with VersionableStorage with History with FeatureProvider with ScorexLogging {
+
+  override protected val Version: Option[Int] = None
```
ismagin: Are you sure this should be `None`?

src/main/scala/com/wavesplatform/utils/package.scala
```diff
@@ -41,10 +42,10 @@ package object utils extends ScorexLogging {
     }
   }
 
-  def createWithStore[A <: AutoCloseable](storeFile: Option[File], f: => A, pred: A => Boolean, deleteExisting: Boolean = false): Try[A] = Try {
+  def createWithStore[A <: AutoCloseable with VersionableStorage](storeFile: Option[File], f: => A, pred: A => Boolean = (_: A) => true, deleteExisting: Boolean = false): Try[A] = Try {
```
ismagin: don't use defaults

src/main/scala/com/wavesplatform/state2/VersionableStorage.scala
```diff
@@ -0,0 +1,20 @@
+package com.wavesplatform.state2
+
+trait VersionableStorage {
+  this: VariablesStorage =>
+
+  protected val Version: Option[Int]
+  private val stateVersion = "stateVersion"
+
+  def isVersionValid: Boolean =
```
ismagin: What is the point of making it a part of class and all the ineritance?
make it a function and pass it as predicate.

src/main/scala/com/wavesplatform/history/HistoryWriterImpl.scala
ismagin: Please move it to companion object

PR #577, opened 2017-10-11
NODE-248 Remember microblock inv

vsuharnikov:


src/main/scala/com/wavesplatform/network/MicroBlockSynchronizer.scala
```diff
@@ -11,27 +12,36 @@ import io.netty.channel._
...
+  lastBlockIdEvents.foreach { lastBlockSig =>
+    tryDownloadNext(lastBlockSig).runAsync
+  }
 
   private def alreadyRequested(microBlockSig: MicroBlockSignature): Boolean = Option(awaitingMicroBlocks.getIfPresent(microBlockSig)).isDefined
 
   private def alreadyProcessed(microBlockSig: MicroBlockSignature): Boolean = Option(successfullyReceivedMicroBlocks.getIfPresent(microBlockSig)).isDefined
 
-  def requestMicroBlockTask(microblockInv: MicroBlockInv, attemptsAllowed: Int): Task[Unit] = Task {
+  private def requestMicroBlockTask(microblockInv: MicroBlockInv, attemptsAllowed: Int): Task[Unit] = Task.unit.flatMap { _ =>
```
ismagin: I think you can replace `Task.unit.flatMap { _ =>` with just `{`  :)

PR #580, opened 2017-10-11
NODE-236: version stored in companion object

ismagin:


ismagin:
Not neccessary, hopefully we'll get new storage soon

PR #589, opened 2017-10-12
Node 247 feature check to block validation

ebceu4:


src/test/scala/com/wavesplatform/history/BlockchainUpdaterMicroblockSunnyDayTest.scala
```diff
@@ -142,7 +142,7 @@ class BlockchainUpdaterMicroblockSunnyDayTest extends PropSpec with PropertyChec
       da.blockchainUpdater.processBlock(block3a).explicitGet()
 
 
-      val db = domain(MicroblocksActivatedAt0WavesSettings, EmptyFeaturesSettings)
+      val db = domain(MicroblocksActivatedAt0WavesSettings, AutoShutdownFeatureSettings)
```
ismagin: since you're here, please clean up the second parameter

src/main/scala/com/wavesplatform/state2/BlockchainUpdaterImpl.scala
```diff
@@ -120,7 +120,25 @@ class BlockchainUpdaterImpl private(persisted: StateWriter with StateReader,
...
+      (newBlockDiff, discarded) = ng
+    } yield {
+      ngState.set(Some(NgState(block, newBlockDiff, 0L, featuresApprovedWithBlock(block))))
+      historyReader.lastBlockId().foreach(lastBlockId.onNext)
+      log.info(s"Block ${block.uniqueId} -> ${trim(block.reference)} appended. New height: ${height + 1}, transactions: ${block.transactionData.size})")
+      discarded
+    }
+  }
+
+  private def processNg(block: Block): Either[ValidationError, (BlockDiff, Seq[Transaction])] = write { implicit l =>
```
ismagin: Something like consensusValidation makes more sense.
Just a suggesstion, you can make it `lazy val blockConsensusValidation = ...`:

```
lazy val blockConsensusValidation = ngState() match {
   ...
}
 for {
      _ <- Either.cond(settings.featuresSettings.autoShutdownOnUnsupportedFeature && notImplementedFeatures.isEmpty, (), GenericError(s"UNIMPLEMENTED ${displayFeatures(notImplementedFeatures)} ACTIVATED ON BLOCKCHAIN, UPDATE THE NODE IMMEDIATELY"))
      newBlockDiffAndDiscardedTxs<- blockConsensusValidation 
      (newBlockDiff, discarded) = newBlockDiffAndDiscardedTxs
    }  ...
```

PR #595, opened 2017-10-13
Node 267 improve resolution of forks

ismagin:
Reapply dropped blocks if extension application has been unsuccessful


ismagin:
Still need to write tests

PR #599, opened 2017-10-13
Node 267 sync fork application 2

ismagin:
- De-synchronized UTX pool
- Introduced peaceful(`Right(None)`) declining of block if better version exists 
- `CoordinatorHandler` via monix executor to reduce entropy
- Check nodes are not deadlocked before stop

### DIFF
--- a/src/main/scala/scorex/utils/Synchronized.scala
+++ b/src/main/scala/scorex/utils/Synchronized.scala
@@ -74,7 +74,7 @@ trait Synchronized extends ScorexLogging {
   def read[T](body: ReadLock => T): T =
     synchronizeOperation(instanceReadLock)(body)
 
-  protected def write[T](body: WriteLock => T): T =
+  def write[T](body: WriteLock => T): T =
     synchronizeOperation(instanceReadWriteLock)(body)
 
   protected def synchronizeOperation[T, L <: TypedLock](lock: L)(body: L => T): T = {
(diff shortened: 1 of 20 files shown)


ismagin:
I don't like the idea to lock utx and state with one RWL, but currently it seems to be the easiest fix(contrary to rewriting UTX w/o lock)

ismagin:
I also had to introduce silent non-application of a block similar, which is not necessary, but it came up with re-application of unsynced version of fork.

src/test/scala/com/wavesplatform/UtxPoolSpecification.scala
```diff
@@ -48,6 +48,8 @@ class UtxPoolSpecification extends FreeSpec
     (state, history)
   }
 
+  def lock = new ReentrantReadWriteLock(true)
```
vsuharnikov: Looks like a non-used code :)

src/main/scala/scorex/utils/Synchronized.scala
```diff
@@ -89,7 +89,7 @@ trait Synchronized extends ScorexLogging {
     }
     finally {
       lock.unlock()
-//      log.trace(s"unlocked $lock")
+      //      log.trace(s"unlocked $lock")
```
alexeykiselev: maybe better remove

PR #615, opened 2017-10-19
NODE 293: Cleanup messages while creating address

ismagin:


src/main/scala/scorex/api/http/ApiError.scala
```diff
@@ -23,7 +23,7 @@ trait ApiError {
 
 object ApiError {
   def fromValidationError(e: ValidationError): ApiError = e match {
-    case ValidationError.InvalidAddress => InvalidAddress
+    case ValidationError.InvalidAddress(_) => InvalidAddress
```
phearnot: We should probably expose address validation error to API consumer

src/main/scala/com/wavesplatform/state2/BlockchainUpdaterImpl.scala
```diff
@@ -262,7 +262,7 @@ class BlockchainUpdaterImpl private(persisted: StateWriter with StateReader,
                 () => ng.bestLiquidDiff.copy(snapshots = Map.empty)),
                 historyWriter.lastBlock.map(_.timestamp), microBlock, ng.base.timestamp)
             } yield {
-              log.info(s"MicroBlock ${microBlock.totalResBlockSig.trim}~>${microBlock.prevResBlockSig.trim} appended, tx count: ${microBlock.transactionData.size}")
+              log.info(s"$microBlock appended")
```
phearnot: Decrease to debug maybe?

src/main/scala/scorex/account/Address.scala
```diff
@@ -62,25 +56,17 @@ object Address extends ScorexLogging {
   }
 
 
-  private def isByteArrayValid(addressBytes: Array[Byte]): Boolean = {
+  private def isByteArrayValid(addressBytes: Array[Byte]): Either[InvalidAddress, Unit] = {
```
phearnot: This should probably be renamed to `validateByteArray` since it no longer returns `Boolean`

src/main/scala/scorex/api/http/ApiError.scala
ismagin: Decided not to break API

src/main/scala/com/wavesplatform/state2/BlockchainUpdaterImpl.scala
ismagin: Will discuss later when do general logging cleanup

PR #617, opened 2017-10-19
Node 296 block in logs

ismagin:


src/main/scala/com/wavesplatform/history/HistoryWriterImpl.scala
```diff
@@ -78,7 +78,7 @@ class HistoryWriterImpl private(file: Option[File], val synchronizationToken: Re
 
         if (h % 100 == 0) db.compact(CompactFillRate, CompactMemorySize)
 
-        log.trace(s"Full Block(id=${block.uniqueId},txs_count=${block.transactionData.size}) persisted")
+        log.trace(s"Full Block $block(id=${block.uniqueId} persisted")
```
vsuharnikov: I think, we don't need `block.uniqueId` here, because it is already logged (from Block.toString):
```scala
s"Block(${signerData.signature.trim}…"
```

src/main/scala/com/wavesplatform/network/CoordinatorHandler.scala
```diff
@@ -83,18 +83,18 @@ class CoordinatorHandler(checkpointService: CheckpointService,
...
           allChannels.broadcast(BlockForged(b), Some(ctx.channel()))
         miner.scheduleMining()
         allChannels.broadcast(LocalScoreChanged(newScore))
       case Left(is: InvalidSignature) =>
-        peerDatabase.blacklistAndClose(ctx.channel(), s"Could not append block ${b.uniqueId}: $is")
+        peerDatabase.blacklistAndClose(ctx.channel(), s"Could not append block $b: $is")
       case Left(ve) =>
         BlockStats.declined(b, BlockStats.Source.Broadcast)
-        log.debug(s"Could not append block ${b.uniqueId}: $ve")
+        log.debug(s"Could not append block $b: $ve")
```
alexeykiselev: Trimmed ID is not useful in scenario when you copy-paste block's ID from log and lookup for it in the explorer
Should we change back to full ID, or simply output full ID in toString?

PR #626, opened 2017-10-24
NODE-300 A block should not be broadcasted to its owners

vsuharnikov:


vsuharnikov:
During issues with rollbacks

PR #629, opened 2017-10-24
Node 306 linearize ng space

ismagin:


src/main/scala/com/wavesplatform/state2/NgState.scala
```diff
@@ -1,38 +1,69 @@
...
+
+class NgState(val base: Block, val baseBlockDiff: BlockDiff, val acceptedFeatures: Set[Short]) {
+
+  private val MaxTotalDiffs = 3
 
+  private val microDiffs: MMap[BlockId, (BlockDiff, Long)] = MMap.empty
+  private val micros: MList[MicroBlock] = MList.empty // fresh head
+  private val totalBlockDiffCache = CacheBuilder.newBuilder()
+    .maximumSize(MaxTotalDiffs)
+    .expireAfterWrite(10, TimeUnit.MINUTES)
```
vsuharnikov: Just interesting, why 10 minutes? This period is much more than a distance between blocks.

src/main/scala/com/wavesplatform/state2/NgState.scala
```diff
@@ -1,38 +1,69 @@
...
-    micros.headOption.map(_.totalResBlockSig)
+  def microBlockIds: Seq[BlockId] = micros.map(_.totalResBlockSig).toList
+
+  private def diffFor(totalResBlockSig: BlockId): BlockDiff =
+    if (totalResBlockSig == base.uniqueId)
+      baseBlockDiff
+    else Option(totalBlockDiffCache.getIfPresent(totalResBlockSig)) match {
+      case Some(d) => d
+      case None =>
+        val prevResBlockSig = micros.find(_.totalResBlockSig == totalResBlockSig).get.prevResBlockSig
```
vsuharnikov: Can we miss a microblock with that `totalResBlockSig`?

ismagin: This is private method. if we fail with `get`, it's a programming error

PR #633, opened 2017-10-27
NODE-308 Node did not receive extensions

vsuharnikov:
Issue was appeared once on devnet

### DIFF
--- a/src/main/scala/com/wavesplatform/network/RemoteScoreObserver.scala
+++ b/src/main/scala/com/wavesplatform/network/RemoteScoreObserver.scala
@@ -11,105 +11,89 @@ import scorex.utils.ScorexLogging
 
 import scala.concurrent.duration.FiniteDuration
 
-
 @Sharable
 class RemoteScoreObserver(scoreTtl: FiniteDuration, lastSignatures: => Seq[ByteStr], initialLocalScore: BigInt)
   extends ChannelDuplexHandler with ScorexLogging {
 
-  private val pinnedChannel = new AtomicReference[Channel]()
-
-  @volatile
-  private var localScore = initialLocalScore
+  private type ScorePair = (Channel, BigInt)
 
   private val scores = new ConcurrentHashMap[Channel, BigInt]
 
-  private def channelWithHighestScore =
+  @volatile private var localScore = initialLocalScore
+  private val currentRequest = new AtomicReference[Option[ScorePair]](None)
+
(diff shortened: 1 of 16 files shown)


src/main/scala/com/wavesplatform/network/RemoteScoreObserver.scala
```diff
@@ -11,105 +11,96 @@ import scorex.utils.ScorexLogging
...
-        }
+  override def handlerAdded(ctx: ChannelHandlerContext): Unit = {
+    ctx.channel().closeFuture().addListener { channelFuture: ChannelFuture =>
+      val closedChannel = channelFuture.channel()
+      Option(scores.remove(closedChannel)).foreach { removedScore =>
+        log.debug(s"${id(ctx)} Closed, removing score $removedScore")
+      }
+
+      trySwitchToBestIf(ctx, s"Switching to second best channel, because ${id(closedChannel)} was closed") {
+        case Some((currChannel, _)) => currChannel == closedChannel
```
phearnot:
```scala
case Some((`closedChannel`, _)) => true
```

src/main/scala/com/wavesplatform/network/RemoteScoreObserver.scala
```diff
@@ -11,105 +11,96 @@ import scorex.utils.ScorexLogging
...
-
-  @volatile
-  private var localScore = initialLocalScore
+  private type ScorePair = (Channel, BigInt)
 
   private val scores = new ConcurrentHashMap[Channel, BigInt]
 
-  private def channelWithHighestScore =
+  @volatile private var localScore = initialLocalScore
+  private val currentRequest = new AtomicReference[Option[ScorePair]](None)
```
phearnot: Pinned channel score doesn't seem to be used anywhere, so why not just have a reference to `Option[Channel]`? That way the code may be simplified even further.

src/main/scala/com/wavesplatform/network/RemoteScoreObserver.scala
```diff
@@ -11,105 +11,96 @@ import scorex.utils.ScorexLogging
...
+
+  private def bestRemotePair: Option[ScorePair] = {
```
phearnot: Is there really a point in renaming this method?

vsuharnikov:
@phearnot Actually, a score was used in `trySwitchToBestIf`, in `.filter(_ == best)`. So I had to return ScorePair back.

PR #637, opened 2017-10-30
NODE-315 Matcher invalid\expired orders cleanup routine

ebceu4:


src/main/scala/com/wavesplatform/matcher/market/OrderBookActor.scala
```diff
@@ -28,6 +28,8 @@ import scala.concurrent.duration._
 
 import com.wavesplatform.network._
 
+case class OrderCleanup()
```
ismagin: case object

src/main/scala/com/wavesplatform/matcher/market/OrderHistoryActor.scala
```diff
@@ -114,6 +116,12 @@ class OrderHistoryActor(val settings: MatcherSettings, val utxPool: UtxPool, val
     }
   }
 
+  def ordersCleanup(orderBook: OrderBook, ts: Long): Either[GenericError, Seq[String]] = {
+    val result = orderBook.asks.values.++(orderBook.bids.values).flatten.filterNot(x => x.order.isValid(ts)).map(_.order.idStr).toSeq
```
koloale: Use validateNewOrder here

src/main/scala/com/wavesplatform/matcher/market/OrderBookActor.scala
```diff
@@ -60,6 +62,8 @@ class OrderBookActor(assetPair: AssetPair,
       onAddOrder(order)
     case cancel: CancelOrder =>
       onCancelOrder(cancel)
+    case OrderCleanup  =>
```
koloale: Where is this message sent? Should be scheduled in a configurable period of time (default 5 min)

src/main/scala/com/wavesplatform/matcher/market/OrderBookActor.scala
```diff
@@ -110,16 +117,23 @@ class OrderBookActor(assetPair: AssetPair,
     context.become(waitingValidation)
   }
 
-  def handleValidateCancelResult(res: Either[GenericError, CancelOrder]): Unit = {
+  def onOrderCleanup(orderBook: OrderBook): Unit = {
+    orderHistory ! OrdersCleanup(orderBook, NTP.correctedTime())
+    apiSender = None
+    cancellable = Some(context.system.scheduler.scheduleOnce(ValidationTimeout, self, ValidationTimeoutExceeded))
```
koloale: No need to handle timeout and become readonly

PR #640, opened 2017-10-31
NODE-313: Chunk size to depend on txs in blocks

ismagin:


src/main/scala/com/wavesplatform/state2/BlockchainUpdaterImpl.scala
```diff
@@ -64,16 +65,21 @@ class BlockchainUpdaterImpl private(persisted: StateWriter with SnapshotStateRea
...
-      val diffToBePersisted = unsafeDiffByRange(persisted, head, last)
-      persisted.applyBlockDiff(diffToBePersisted)
+    val notPersisted = historyWriter.height() - persisted.height
+    val inMemSize = Math.min(notPersisted, maxBlocksInMemory)
+    val persistUpTo = historyWriter.height() - inMemSize + 1
+
+    val persistFrom = persisted.height + 1
+    ranges(persistFrom, persistUpTo, rebuildByBlocks).foreach { case (head, last) =>
+      val diffs = unsafeDiffByRange(persisted, head, last)
+      log.debug(s"Diffs built for Range($persistFrom, $persistUpTo): ${diffs.map(_.txsDiff.transactions.size)}")
```
phearnot: `diffs` seems to always contain just one element. This should likely be `Diffs built for [$head, $last]: ${transactions.size}`

PR #641, opened 2017-11-01
Node discovery tool

ebceu4:


discovery/build.sbt
```diff
@@ -0,0 +1,3 @@
+libraryDependencies += "com.typesafe.akka" % "akka-http-experimental_2.11" % "2.4.11.2"
```
ismagin: we use 2.12, akka-http is not experimental, please update dependedncies
`"com.typesafe.akka" %% "akka-http" % "10.0.9"`

discovery/devnet.conf
```diff
@@ -0,0 +1,25 @@
+generator {
```
ismagin: why do we need this for discovery?

discovery/src/main/scala/com.wavesplatform.discovery/DiscoveryApp.scala
```diff
@@ -0,0 +1,55 @@
...
+        handleWebSocketMessages(Flow.fromSinkAndSource(sink, source))
+      }
+    }
+
+    (route, system.scheduler.schedule(FiniteDuration(0, TimeUnit.SECONDS), Settings.default.discoveryInterval, mainActor, MainActor.Discover))
+  }.reduce((a,b) => (a._1 ~ b._1, a._2.combine(b._2)))
+
+  val binding = Http().bindAndHandle(route, Settings.default.webSocketHost, Settings.default.webSocketPort)
+
+  println(s"Server is now online at http://${Settings.default.webSocketHost}:${Settings.default.webSocketPort}\nPress RETURN to stop...")
```
ismagin: use `log.info` instead of println
you'll have to adopt logback as in other projects

discovery/src/main/scala/com.wavesplatform.discovery/DiscoveryApp.scala
```diff
@@ -0,0 +1,55 @@
...
+  StdIn.readLine()
```
ismagin: :(

discovery/src/main/scala/com.wavesplatform.discovery/collections/ExpirationSet.scala
```diff
@@ -0,0 +1,26 @@
+package com.wavesplatform.discovery.collections
+
+class ExpirationSet[T](expirationTimeMilis: Long) extends scala.collection.mutable.Set[T]{
```
ismagin: you can use guava's `cache` like in `MicroblockSyncronizer`: 
`  private val nextInvs = cache[MicroBlockSignature, MicroBlockInv](settings.invCacheTimeout)`

discovery/src/main/scala/com.wavesplatform.discovery/network/LegacyFrameCodec.scala
```diff
@@ -0,0 +1,64 @@
+package com.wavesplatform.discovery.network
```
ismagin: is this copy-pasted?

discovery/src/main/scala/com.wavesplatform.discovery/network/MessageCodec.scala
```diff
@@ -0,0 +1,39 @@
+package com.wavesplatform.discovery.network
```
ismagin: I'm sure you don't need to encode all of this, either reiuse if it's not important, or clean this from Microblocks, blocks, etc

discovery/src/main/scala/com.wavesplatform.discovery/package.scala
```diff
@@ -0,0 +1,12 @@
+package com.wavesplatform
+
+import io.netty.channel.nio.NioEventLoopGroup
+
+import scala.concurrent.ExecutionContext
+
+package object discovery {
```
ismagin: why not define it where needed and not abuse implicits?

discovery/src/main/scala/com.wavesplatform.discovery/DiscoveryApp.scala
ismagin: `Await.result(system.whenTerminated, Duration.Inf)`

discovery/build.sbt
```diff
@@ -0,0 +1,3 @@
+libraryDependencies += "com.typesafe.akka" %% "akka-http" % "10.0.9"
+libraryDependencies += "com.github.pureconfig" %% "pureconfig" % "0.8.0"
+libraryDependencies += "com.typesafe.akka" % "akka-actor_2.10" % "2.2-M1"
```
ismagin: please bump this too

discovery/src/main/scala/com.wavesplatform.discovery/DiscoveryApp.scala
```diff
@@ -0,0 +1,58 @@
...
+      }
+    }
+
+    (route, system.scheduler.schedule(FiniteDuration(0, TimeUnit.SECONDS), Settings.default.discoveryInterval, mainActor, MainActor.Discover))
+  }.reduce((a,b) => (a._1 ~ b._1, a._2.combine(b._2)))
+
+  val binding = Http().bindAndHandle(route, Settings.default.webSocketHost, Settings.default.webSocketPort)
+
+  log.info(s"Server is now online at http://${Settings.default.webSocketHost}:${Settings.default.webSocketPort}\nPress RETURN to stop...")
+  StdIn.readLine()
```
ismagin: To await shutdown, use `Await.result(system.whenTerminated, Duration.Inf)`

ismagin:
Use shutdown hooks to properly close connections on `Ctrl-C`

PR #650, opened 2017-11-02
NODE-319 Node should not request an extension before a previous one is applied

vsuharnikov:
* A node doesn't break an extension loading when mining;
* A node doesn't mine a microblock during a block mining.

### DIFF
--- a/src/main/scala/com/wavesplatform/network/CoordinatorHandler.scala
+++ b/src/main/scala/com/wavesplatform/network/CoordinatorHandler.scala
@@ -44,9 +44,9 @@ class CoordinatorHandler(checkpointService: CheckpointService,
   private val processBlock = Coordinator.processSingleBlock(checkpointService, history, blockchainUpdater, time, stateReader, utxStorage, settings.blockchainSettings, featureProvider) _
   private val processMicroBlock = Coordinator.processMicroBlock(checkpointService, history, blockchainUpdater, utxStorage) _
 
-  private def scheduleMiningAndBroadcastScore(reason: LocalScoreChanged.Reason = LocalScoreChanged.Reason.Other)(score: BigInt): Unit = {
+  private def scheduleMiningAndBroadcastScore(breakExtLoading: Boolean)(score: BigInt): Unit = {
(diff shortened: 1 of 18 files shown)


src/main/scala/com/wavesplatform/network/messages.scala
```diff
@@ -36,4 +36,18 @@ object MicroBlockInv{
...
+
+object LocalScoreChanged {
+  sealed trait Reason extends Product with Serializable
+  object Reason {
+    val All: Set[Reason] = Set(ForkApplied, Rollback, Checkpoint, Other)
+
+    case object ForkApplied extends Reason
+    case object Rollback extends Reason
+    case object Checkpoint extends Reason
+    case object Other extends Reason
```
ismagin: Other = BlockApplied?

src/main/scala/com/wavesplatform/network/CoordinatorHandler.scala
```diff
@@ -43,9 +43,9 @@ class CoordinatorHandler(checkpointService: CheckpointService,
   private val processBlock = Coordinator.processSingleBlock(checkpointService, history, blockchainUpdater, time, stateReader, utxStorage, settings.blockchainSettings, featureProvider) _
   private val processMicroBlock = Coordinator.processMicroBlock(checkpointService, history, blockchainUpdater, utxStorage) _
 
-  private def scheduleMiningAndBroadcastScore(score: BigInt): Unit = {
+  private def scheduleMiningAndBroadcastScore(reason: LocalScoreChanged.Reason = LocalScoreChanged.Reason.Other)(score: BigInt): Unit = {
```
ismagin: What's the point of using  default value?

vsuharnikov: To not break extension loading

src/main/scala/com/wavesplatform/network/messages.scala
vsuharnikov: Yes, but I'll change it to flag

PR #651, opened 2017-11-02
NODE-321-Add-API-to-get-blocks-headers

ebceu4:


src/main/scala/scorex/block/Block.scala
```diff
@@ -59,6 +59,21 @@ case class Block(timestamp: Long,
         "blocksize" -> bytes.length
       )
 
+  lazy val jsonWithoutTransactionsPayload: JsObject =
```
ismagin: looks like copy-paste, why not just
 ```
def json(txsBodies: Boolean) =
 ... ++ 
 if (txsBodies) transactionField.json else Json.obj("transactions" -> transactionField.value.size) ++ 
... ++
...
```

ismagin:
yet another merge

PR #658, opened 2017-11-07
NODE-321 API to get blocks headers

ebceu4:


src/main/scala/scorex/api/http/BlocksApiRoute.scala
```diff
@@ -104,19 +104,25 @@ case class BlocksApiRoute(settings: RestAPISettings, checkpointsSettings: Checkp
...
-  private def at(height: Int, includeTransactions: Boolean): StandardRoute =
-    history.blockAt(height).map(_.json(includeTransactions)) match {
-    case Some(json) => complete(json + ("height" -> JsNumber(height)))
-    case None => complete(Json.obj("status" -> "error", "details" -> "No block for this height"))
+  private def at(height: Int, includeTransactions: Boolean): StandardRoute = {
+    (if (includeTransactions) {
+      history.blockAt(height).map(_.json)
+    }
+    else {
+      history.blockHeaderAndSizeAt(height).map(x => x._1.json(x._2))
```
ismagin: next time try to use pattern-matching where possible
```
map { case((header,size)) => ...
```

ebceu4: yes, my liege

PR #663, opened 2017-11-09
Node 325: Tasks for HistoryReplier

ismagin:


src/main/scala/com/wavesplatform/network/HistoryReplier.scala
```diff
@@ -6,70 +6,69 @@ import com.wavesplatform.settings.SynchronizationSettings
...
-      }
-
-    case _: Handshake =>
-      Future(blocking(history.score())).onComplete {
-        case Success(score) => ctx.writeAndFlush(LocalScoreChanged(score))
-        case Failure(e) => ctx.fireExceptionCaught(e)
-      }
+    case _: Handshake => Task {
+      ctx.writeAndFlush(LocalScoreChanged(history.score()))
+    }.runAsync
```
vsuharnikov: It could be done without a task.
1. We send a current score
2. We process a block and broadcast a new score.

With Task we may not respond in a given term and could be banned

src/main/scala/com/wavesplatform/network/HistoryReplier.scala
```diff
@@ -6,70 +6,69 @@ import com.wavesplatform.settings.SynchronizationSettings
...
 
-    case GetBlock(sig) =>
-      Future(knownBlocks.get(sig)).foreach { bytes =>
-        ctx.writeAndFlush(RawBytes(BlockMessageSpec.messageCode, bytes))
-      }
+    case GetBlock(sig) => Task {
+      Option(knownBlocks.get(sig)).foreach(bytes =>
+        ctx.writeAndFlush(RawBytes(BlockMessageSpec.messageCode, bytes)))
+    }.runAsync
```
vsuharnikov: `knownBlocks` is a concurrent data structure. Also we can't send a signature of block we didn't processed. So, the remote node should not ask it.

src/main/scala/com/wavesplatform/network/HistoryReplier.scala
vsuharnikov: nvm, `history.score` is a potentially long-running operation

PR #668, opened 2017-11-13
Don't discard properly loaded extension

phearnot:


src/test/scala/com/wavesplatform/BlockGen.scala
```diff
@@ -17,11 +17,21 @@ trait BlockGen extends TransactionGen {
...
+  def versionedBlockGen(txs: Seq[Transaction], signer: PrivateKeyAccount, version: Byte): Gen[Block] =
+    byteArrayGen(Block.BlockIdLength).flatMap(ref => versionedBlockGen(ByteStr(ref), txs, signer, version))
+
+  def versionedBlockGen(reference: ByteStr, txs: Seq[Transaction], signer: PrivateKeyAccount, version: Byte): Gen[Block] = for {
     baseTarget <- Gen.posNum[Long]
     generationSignature <- byteArrayGen(Block.GeneratorSignatureLength)
-  } yield Block.buildAndSign(version, txs.map(_.timestamp).max, ByteStr(reference), NxtLikeConsensusBlockData(baseTarget, ByteStr(generationSignature)), txs, signer, Set.empty).explicitGet()
+  } yield Block.buildAndSign(
+      version,
+      if (txs.isEmpty) System.currentTimeMillis() else  txs.map(_.timestamp).max,
```
ismagin: why `System.currentTimeMillis()` and not just  `Gen[Long]` timestamp?

PR #677, opened 2017-11-15
NODE-324

ebceu4:
Utx pool doesn't get cleaned

### DIFF
--- a/src/main/scala/com/wavesplatform/UtxPool.scala
+++ b/src/main/scala/com/wavesplatform/UtxPool.scala
@@ -4,7 +4,7 @@ import java.util.concurrent.ConcurrentHashMap
 
 import cats._
 import com.google.common.cache.CacheBuilder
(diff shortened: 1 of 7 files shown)


src/main/scala/com/wavesplatform/UtxPool.scala
```diff
@@ -27,10 +30,23 @@ class UtxPool(time: Time,
               history: History,
               feeCalculator: FeeCalculator,
               fs: FunctionalitySettings,
-              utxSettings: UtxSettings) extends ScorexLogging with Instrumented {
+              utxSettings: UtxSettings) extends ScorexLogging with Instrumented with AutoCloseable {
 
   private val transactions = new ConcurrentHashMap[ByteStr, Transaction]()
 
+  private implicit val scheduler: Scheduler = Scheduler.global
```
ismagin: Scheduler.singleThread

PR #681, opened 2017-11-20
NODE-348 It network improvement

vsuharnikov:
* Fixed several issues with networking in integration tests (updating ports after reconnect to the network, force reconnection between peers, etc.);
* Honest reporting of outgoing and incoming connections;
* Some tests are updated (simplified or do less work).

### DIFF
--- a/src/it/scala/com/wavesplatform/it/BlacklistTestSuite.scala
+++ b/src/it/scala/com/wavesplatform/it/BlacklistTestSuite.scala
@@ -7,7 +7,7 @@ import org.scalatest._
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future.traverse
 import scala.concurrent.duration._
-import scala.concurrent.{Await, Future}
+import scala.concurrent.Await
 
 class BlacklistTestSuite extends FreeSpec with Matchers with BeforeAndAfterAll with CancelAfterFailure
   with ReportingTestName with MultipleNodesApi {
@@ -26,7 +26,7 @@ class BlacklistTestSuite extends FreeSpec with Matchers with BeforeAndAfterAll w
 
   override protected def beforeAll(): Unit = {
     super.beforeAll()
-    Await.result(Future.traverse(nodes)(_.waitForPeers(nodes.size - 1)), 2.minute)
+    nodes.foreach(_.status) // Initialize
   }
 
   override protected def afterAll(): Unit = {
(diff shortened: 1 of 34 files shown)


src/it/scala/com/wavesplatform/it/transactions/AliasTransactionSuite.scala
```diff
@@ -62,7 +62,7 @@ class AliasTransactionSuite extends BaseTransactionSuite with TableDrivenPropert
       aliasTxId <- sender.createAlias(firstAddress, alias, aliasFee).map(_.id)
 
       _ <- waitForHeightAraiseAndTxPresent(aliasTxId, 1)
-      _ <- assertBadRequest(sender.createAlias(secondAddress, alias, aliasFee))
+      //_ <- assertBadRequest(sender.createAlias(secondAddress, alias, aliasFee))
```
ismagin: please uncomment or remove

src/it/scala/com/wavesplatform/it/BlacklistTestSuite.scala
```diff
@@ -26,7 +26,7 @@ class BlacklistTestSuite extends FreeSpec with Matchers with BeforeAndAfterAll w
 
   override protected def beforeAll(): Unit = {
     super.beforeAll()
-    Await.result(Future.traverse(nodes)(_.waitForPeers(nodes.size - 1)), 2.minute)
+    nodes.foreach(_.status) // Initialize
```
phearnot: Why no `Await` here?

src/it/scala/com/wavesplatform/it/Docker.scala
```diff
@@ -183,12 +248,18 @@ class Docker(suiteConfig: Config = ConfigFactory.empty,
...
         node.close()
         client.stopContainer(node.nodeInfo.containerId, 0)
-        client.disconnectFromNetwork(node.nodeInfo.containerId, wavesNetworkId)
+        disconnectFromNetwork(node)
         saveLog(node)
         client.removeContainer(node.nodeInfo.containerId, RemoveContainerParam.forceKill())
       }
 
-      client.removeNetwork(wavesNetworkId)
+      Try(client.removeNetwork(wavesNetwork.id)) match {
```
phearnot: Is `Try(...) match {` any better than plain old `try { ... } catch { ... }` in this case?

vsuharnikov: Nope, I'll fix it

src/it/scala/com/wavesplatform/it/BlacklistTestSuite.scala
vsuharnikov: I've replace it by the logging a number of nodes. This line is required to initialize the lazy variable only. So, we haven't to do it

PR #684, opened 2017-11-24
all balances freezed by the orders +itest

ebceu4:


src/main/scala/com/wavesplatform/matcher/market/OrderHistoryActor.scala
```diff
@@ -170,4 +174,14 @@ object OrderHistoryActor {
     val json: JsObject = JsObject(balances.map{ case (k, v) => (k, JsNumber(v)) })
     val code = StatusCodes.OK
   }
+
+  case class GetMatcherBalance(address: String, ts: Long) extends ExpirableOrderHistoryRequest
+  case class GetMatcherBalanceResponse(balances: Option[Map[String, Long]]) extends MatcherResponse {
+
+    //val json = Json.obj("status" -> "OrderCancelRejected", "message" -> message)
```
ismagin: please cleanup

PR #685, opened 2017-11-24
NODE-358 Blacklist nodes with known invalid blocks during the synchronization

vsuharnikov:


src/main/scala/com/wavesplatform/network/ExtensionSignaturesLoader.scala
```diff
@@ -8,14 +8,20 @@ import scorex.utils.ScorexLogging
...
+                                knownInvalidBlocks: ByteStr => Boolean)
   extends ChannelDuplexHandler with ScorexLogging {
 
   private var currentTimeout = Option.empty[ScheduledFuture[Unit]]
   private var lastKnownSignatures = Seq.empty[ByteStr]
 
   override def channelRead(ctx: ChannelHandlerContext, msg: AnyRef): Unit = msg match {
     case s: Signatures =>
+      if (s.signatures.contains(knownInvalidBlocks)) {
+        peerDatabase.blacklistAndClose(ctx.channel(), "Has known invalid block")
```
ismagin: please log $invalidBlockId

src/main/scala/com/wavesplatform/network/NetworkServer.scala
```diff
@@ -84,9 +85,25 @@ class NetworkServer(checkpointService: CheckpointService,
     settings.networkSettings.maxInboundConnections,
     settings.networkSettings.maxConnectionsPerHost)
 
+  private val knownInvalidBlocks = {
+    val residenceTime = 1.day
```
ismagin: should we move this to config?

src/main/scala/com/wavesplatform/Coordinator.scala
```diff
@@ -24,7 +24,7 @@ import scala.util.{Left, Right}
 object Coordinator extends ScorexLogging with Instrumented {
   def processFork(checkpoint: CheckpointService, history: History, blockchainUpdater: BlockchainUpdater,
                   stateReader: StateReader, utxStorage: UtxPool, time: Time, settings: WavesSettings,
-                  blockchainReadiness: AtomicBoolean, featureProvider: FeatureProvider)
+                  blockchainReadiness: AtomicBoolean, featureProvider: FeatureProvider, storeInvalidBlock: ByteStr => Unit)
```
ismagin: I'd use some
```
trait InvalidBlockStorage {
def store()
def exists()
}
```
This is more obvious when you read/support/further develop the functionality
It's more consistent  with existing  repositories like utx, peerdatabase, microblockOwners, etc

PR #688, opened 2017-11-27
Node 339 level db

alexeykiselev:


src/main/scala/com/wavesplatform/Application.scala
```diff
@@ -249,46 +246,63 @@ object Application extends ScorexLogging {
...
-          Kamon.shutdown()
-          Metrics.shutdown()
-          super.shutdown()
+        // Initialize global var with actual address scheme
+        AddressScheme.current = new AddressScheme {
+          override val chainId: Byte = settings.blockchainSettings.addressSchemeCharacter.toByte
         }
-      }.run()
+
+        log.info(s"${
```
ismagin: This weird formatting happens when you format incorrect with IDEA while syntax is incorrect in terms of `{` and `}`

src/main/scala/com/wavesplatform/history/StorageFactory.scala
```diff
@@ -1,44 +1,45 @@
...
+import org.iq80.leveldb.DB
 import scorex.transaction._
 import scorex.utils.{NTP, Time}
 
 import scala.util.{Success, Try}
 
 object StorageFactory {
 
-  type Storage = Coeval[(NgHistory with DebugNgHistory with AutoCloseable, FeatureProvider, AutoCloseable, StateReader, BlockchainUpdater, BlockchainDebugInfo)]
+  type Storage = Coeval[(NgHistory with DebugNgHistory, FeatureProvider, StateWriter, StateReader, BlockchainUpdater, BlockchainDebugInfo)]
```
alexeykiselev: no need to expose StateWriter

PR #703, opened 2017-11-30
NODE-366 Disable PaymentTransaction in block version 3

petermz:
Changes:
- Payment tx won't be allowed starting from block version 3
- Added a testcase to PaymentTransactionDiffTest.scala
- Removed obsolete API routes and corresponding tests

The essential fix is in TransactionDiffer.scala, other changes are mostly in tests


src/main/scala/com/wavesplatform/state2/diffs/TransactionDiffer.scala
```diff
@@ -24,7 +24,12 @@ object TransactionDiffer {
       t5 <- CommonValidation.disallowSendingGreaterThanBalance(s, settings, currentBlockTimestamp, t4)
       diff <- t5 match {
         case gtx: GenesisTransaction => GenesisTransactionDiff(currentBlockHeight)(gtx)
-        case ptx: PaymentTransaction => PaymentTransactionDiff(s, currentBlockHeight, settings, currentBlockTimestamp)(ptx)
+        case ptx: PaymentTransaction =>
```
ismagin: please move this `if` etc to `PaymentTransactionDiff` in order to keep things consistent

src/main/scala/com/wavesplatform/state2/diffs/TransactionDiffer.scala
```diff
@@ -24,7 +24,12 @@ object TransactionDiffer {
...
+          if (currentBlockHeight <= settings.blockVersion3AfterHeight) {
+            PaymentTransactionDiff(s, currentBlockHeight, settings, currentBlockTimestamp)(ptx)
+          } else {
+            Left(UnsupportedTransactionType)
```
ismagin: I think `GenericError(s"Payment transaction is deprecated after h=${settings.blockVersion3AfterHeight}")`  would give more info to users, at least, it would be easier to find the exact line where `Left` appeared

petermz:
Changes in this version:
- Restored legacy REST endpoints and tests
- Added new ApiError subclass, `DiscontinuedApi`

petermz:
In this commit:
- Removed PaymentRequest & SignedPaymentRequest classes
- Removed Swagger annotations from discontinued REST endpoints

PR #707, opened 2017-12-01
Increase priority of blacklist assets validation

koloale:


ismagin:
Stale request, 0.8 is out already

PR #708, opened 2017-12-04
NODE-379 Add "cancelled" flag to leasing transaction

petermz:
[Link to JIRA](https://wavesplatform.atlassian.net/browse/NODE-379)

Note that I've named the flag "canceled" (one L) because this is the preferred American English spelling according to [Grammarly](https://www.grammarly.com/blog/canceled-vs-cancelled/)

### DIFF
--- a/src/it/scala/com/wavesplatform/it/transactions/LeasingTransactionsSuite.scala
+++ b/src/it/scala/com/wavesplatform/it/transactions/LeasingTransactionsSuite.scala
@@ -2,6 +2,8 @@ package com.wavesplatform.it.transactions
 
 import com.wavesplatform.it.util._
 import org.scalatest.CancelAfterFailure
+import play.api.libs.json.Json
+import scorex.transaction.lease.LeaseTransaction
 
 import scala.concurrent.Await
 import scala.concurrent.Future.traverse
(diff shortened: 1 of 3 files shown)


src/main/scala/scorex/api/http/TransactionsApiRoute.scala
```diff
@@ -126,6 +126,8 @@ case class TransactionsApiRoute(
 
   private def txToExtendedJson(tx: Transaction): JsObject = {
     tx match {
+      case lease: LeaseTransaction =>
+        lease.json() ++ Json.obj("canceled" -> ! state().isLeaseActive(lease))
```
ismagin: negation is hard to read, I suggest `status = ACTIVE/CANCELLED`

src/it/scala/com/wavesplatform/it/transactions/LeasingTransactionsSuite.scala
```diff
@@ -139,4 +140,27 @@ class LeasingTransactionsSuite extends BaseTransactionSuite with CancelAfterFail
 
     Await.result(f, waitCompletion)
   }
+
+  test("canceled flag is set correctly for lease transactions") {
```
ismagin: You could just upgrade `lease cancellation reverts eff.b. changes` test with 1 assert

petermz:
Changes in this commit:
- Replaced boolean `canceled` flag with `status` with two values: `Active` and `Canceled`
- Merged the new test into an existing one

PR #741, opened 2017-12-11
Cleaned up fee settings for the obsolete payment transaction

petermz:
Removed payment transaction fee from fee settings, also cleaned up tests


petermz:
I've reverted all changes to tests, so only application.conf cleanup has left

PR #750, opened 2017-12-16
NODE-380 Add "height" value to transactions received from API by address

petermz:
https://wavesplatform.atlassian.net/browse/NODE-380
Changed SnapshotStateReader to return both transactions and their heights.

### DIFF
--- a/src/it/scala/com/wavesplatform/it/transactions/TransferTransactionSuite.scala
+++ b/src/it/scala/com/wavesplatform/it/transactions/TransferTransactionSuite.scala
@@ -1,7 +1,9 @@
 package com.wavesplatform.it.transactions
 
+import com.wavesplatform.it.api._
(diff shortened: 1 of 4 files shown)


src/it/scala/com/wavesplatform/it/transactions/TransferTransactionSuite.scala
```diff
@@ -206,4 +208,22 @@ class TransferTransactionSuite extends BaseTransactionSuite with CancelAfterFail
 
     Await.result(f, waitCompletion)
   }
+
+  // probably should be moved elsewhere, just don't see any suite that fits the purpose right now
```
ismagin: Please don't leave TODOs/FIXMEs of any kind, either do the improvement if it's needed  or remove the letter

PR #752, opened 2017-12-19
wallet-account-generation-test

ebceu4:


ismagin:
Are you sure we need this in future versions? I suggest we drop WalletObsolete completely in 0.9x

PR #755, opened 2017-12-21
NODE-410 Response to blocks API call without lock

petermz:
JIRA: https://wavesplatform.atlassian.net/browse/NODE-410

A few HistoryWriterImpl methods call atomic operations on MVMaps under a read lock. This is not needed because MVMap is thread safe.

I'm introducing a new method, `Synchronized.lockfree`, that provides the implicit lock parameter but does not acquire the lock.

### DIFF
--- a/src/main/scala/com/wavesplatform/history/HistoryWriterImpl.scala
+++ b/src/main/scala/com/wavesplatform/history/HistoryWriterImpl.scala
@@ -111,17 +111,17 @@ class HistoryWriterImpl private(file: Option[File], val synchronizationToken: Re
       .reverse
   }
 
-  override def height(): Int = read { implicit lock => blockIdByHeight().size() }
+  override def height(): Int = lockfree { implicit lock => blockIdByHeight().size() }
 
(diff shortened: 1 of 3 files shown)


src/main/scala/com/wavesplatform/history/HistoryWriterImpl.scala
```diff
@@ -111,17 +111,17 @@ class HistoryWriterImpl private(file: Option[File], val synchronizationToken: Re
...
 
-  override def height(): Int = read { implicit lock => blockIdByHeight().size() }
+  override def height(): Int = lockfree { implicit lock => blockIdByHeight().size() }
 
   override def scoreOf(id: ByteStr): Option[BlockchainScore] = read { implicit lock =>
     heightOf(id).map(scoreByHeight().get(_))
   }
 
-  override def heightOf(blockSignature: ByteStr): Option[Int] = read { implicit lock =>
+  override def heightOf(blockSignature: ByteStr): Option[Int] = lockfree { implicit lock =>
```
ismagin: `NgHistoryReader`, which is actually used everywhere still acquires `read` here and in other places

PR #759, opened 2017-12-25
NODE-438: utx cache to move to network message handler

ismagin:


src/main/scala/com/wavesplatform/network/MicroBlockSynchronizer.scala
```diff
@@ -70,7 +70,7 @@ object MicroBlockSynchronizer {
           peerDatabase.blacklistAndClose(ch, err.toString)
         case Right(_) =>
           microBlockOwners.get(totalSig, () => MSet.empty) += ch
-          nextInvs.get(prevSig, { () =>
+          nextInvs. get(prevSig, { () =>
```
Tolsi: ?

PR #761, opened 2017-12-25
NODE-426 Extensions loading optimisation

vsuharnikov:
Also fixes for tests.

Most important changes in:
* RxExtensionLoader
* RxScoreObserver

### DIFF
--- a/src/main/scala/com/wavesplatform/network/RxExtensionLoader.scala
+++ b/src/main/scala/com/wavesplatform/network/RxExtensionLoader.scala
@@ -61,7 +61,7 @@ object RxExtensionLoader extends ScorexLogging {
                   val ch = best.channel
                   log.debug(s"${id(ch)} Requesting extension signatures ${if (optimistic) "optimistically" else ""}, last ${knownSigs.length} are ${formatSignatures(knownSigs)}")
                   val blacklisting = scheduleBlacklist(ch, s"Timeout loading extension").runAsync
-                  Task(ch.writeAndFlush(GetSignatures(knownSigs))).logErr.runAsync
+                  ch.writeAndFlush(GetSignatures(knownSigs))
                   state.withLoaderState(LoaderState.ExpectingSignatures(ch, knownSigs, blacklisting))
                 case None =>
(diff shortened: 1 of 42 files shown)


src/main/scala/com/wavesplatform/network/RxExtensionLoader.scala
```diff
@@ -178,7 +177,7 @@ object RxExtensionLoader extends ScorexLogging {
         case ApplierState.Applying(maybeBuffer, applying) =>
           if (applying != extension) log.warn(s"Applied $extension doesn't match expected $applying")
           maybeBuffer match {
-            case None => syncNext(state.copy(applierState = ApplierState.Idle))
+            case None => state.copy(applierState = ApplierState.Idle)
```
ismagin: Why not sync next?

src/main/scala/com/wavesplatform/network/RxExtensionLoader.scala
```diff
@@ -162,11 +163,9 @@ object RxExtensionLoader extends ScorexLogging {
           log.error("Error applying extension", err)
           Left(GenericError(err))
         })
-        .flatMap { ar =>
-          Task {
-            s = onExtensionApplied(s, extensionBlocks, ch, ar)
-          }.executeOn(scheduler)
-        }.logErr.runAsync(scheduler)
+        .map { ar => s = onExtensionApplied(s, extensionBlocks, ch, ar) }
```
ismagin: When you do `map`, task gets executed in the same scheduler, so we do need flatmap here!

vsuharnikov: Already fixed with `asyncBoundary(scheduler)`, it's a wip :D

src/main/scala/com/wavesplatform/network/RxExtensionLoader.scala
vsuharnikov: because the local score comes to RxExtensionLoader after an extension applying, so `syncNext` will happen anyway. An error during an extension applying is handled by `onExtensionApplied`

PR #764, opened 2017-12-27
NODE-446 /node/status returns state height that lags behind blockchai…

petermz:
JIRA: https://wavesplatform.atlassian.net/browse/NODE-446

Instead of reading persisted state height from StateWriter, I'm now reading complete (persisted + in-memory) state height from SnapshotStateReader

### DIFF
--- a/src/main/scala/com/wavesplatform/http/NodeApiRoute.scala
+++ b/src/main/scala/com/wavesplatform/http/NodeApiRoute.scala
@@ -5,24 +5,17 @@ import javax.ws.rs.Path
 
 import akka.http.scaladsl.server.Route
 import com.wavesplatform.Shutdownable
-import com.wavesplatform.network.lastObserved
 import com.wavesplatform.settings.{Constants, RestAPISettings}
-import com.wavesplatform.state2.{StateReader, StateWriter}
 import io.swagger.annotations._
-import monix.eval.Coeval
-import monix.execution.Scheduler.Implicits.global
 import play.api.libs.json.Json
 import scorex.api.http.{ApiRoute, CommonApiFunctions}
-import scorex.transaction.{BlockchainUpdater, LastBlockInfo}
+import scorex.transaction.BlockchainDebugInfo
 import scorex.utils.ScorexLogging
 
 @Path("/node")
 @Api(value = "node")
-case class NodeApiRoute(settings: RestAPISettings, blockchainUpdater: BlockchainUpdater,
-                        stateReader: StateReader, stateWriter: StateWriter, application: Shutdownable)
+case class NodeApiRoute(settings: RestAPISettings, debugInfo: BlockchainDebugInfo, application: Shutdownable)
   extends ApiRoute with CommonApiFunctions with ScorexLogging {
-
-  private val lastHeight: Coeval[Option[LastBlockInfo]] = lastObserved(blockchainUpdater.lastBlockInfo)
 
   override lazy val route = pathPrefix("node") {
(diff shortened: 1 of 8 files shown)


src/main/scala/com/wavesplatform/state2/StateStorage.scala
```diff
@@ -17,9 +17,14 @@ class StateStorage private(file: Option[File]) extends VariablesStorage(createMV
 
   override protected val Version = 2
 
+  @volatile private[state2] var lastUpdated: Long = 0L
```
ismagin: Please clean this up, make a public getter and private setter, package-private vars is a mess

src/main/scala/com/wavesplatform/state2/StateWriter.scala
```diff
@@ -5,22 +5,18 @@ import java.util.concurrent.locks.ReentrantReadWriteLock
...
 import scorex.utils.ScorexLogging
-import scorex.utils.Synchronized.ReadLock
 
 trait StateWriter {
   def applyBlockDiff(blockDiff: BlockDiff): Unit
 
   def clear(): Unit
 
-  def status: Observable[Status]
+  def lastUpdated: Long
```
ismagin: is `Status` not used anymore?

src/main/scala/com/wavesplatform/state2/StateStorage.scala
```diff
@@ -17,9 +17,14 @@ class StateStorage private(file: Option[File]) extends VariablesStorage(createMV
...
+
   def getHeight: Int = getInt(heightKey).getOrElse(0)
 
-  def setHeight(i: Int): Unit = putInt(heightKey, i)
+  def setHeight(i: Int): Unit = {
+    putInt(heightKey, i)
+    lastUpdated = System.currentTimeMillis
```
ismagin: Why not use `Time`? It is not critical here, but it's best if we use single source of time everywhere

src/main/scala/com/wavesplatform/http/NodeApiRoute.scala
```diff
@@ -49,7 +49,7 @@ case class NodeApiRoute(settings: RestAPISettings, blockchainUpdater: Blockchain
   @ApiOperation(value = "Status", notes = "Get status of the running core", httpMethod = "GET")
   def status: Route = (get & path("status")) {
     val (bcHeight, bcTime) = lastHeight().map { case LastBlockInfo(_, h, _, _, t) => (h, t) }.getOrElse((0, 0L))
-    val (stHeight, stTime) = lastState().map { case StateWriter.Status(h, t) => (h, t) }.getOrElse((0, 0L))
+    val (stHeight, stTime) = (stateReader().height, stateWriter.lastUpdated)
```
ismagin: There's th `trait BlockchainDebugInfo`, please use it this stuff, BlockchainUpdater and StateWriter shouldn't be here at all

petermz:
In this commit:
- Moved stuff into BlockchainDebugInfo to avoid exposing StateReader/Writer
- Added history height and timestamps to StateDebugInfo (should I rename this class now?)
- Time is used everywhere instead of System.currentTimeMillis

petermz:
No, wait, that doesn't work because of locking =(

ismagin:
Looks like all you need to expose from not-yet-rebuilt state is height. Maybe you can just expose coeval instead of  accepting callback?

ismagin:
NODE-467 Stop exposing internal entities to NodeApiRoute #780 suites better

PR #768, opened 2017-12-27
Smart Accounts

ismagin:


build.sbt
```diff
@@ -153,12 +135,62 @@ linuxScriptReplacements += "detect-loader" ->
...
 ))
 
+commands += Command.command("packageAll") { state =>
+  "clean" :: "assembly" :: "debian:packageBin" :: state
+}
+
+lazy val lang =
+  crossProject(JSPlatform, JVMPlatform)
+    .withoutSuffixFor(JVMPlatform)
+    .crossType(CrossType.Pure)
```
vsuharnikov: [CrossType.Full](https://github.com/portable-scala/sbt-crossproject#configuration) is preferred here, because:
1. `.js` (`.jvm`, etc.) directories is hidden by OS conventions and...
2. we have a platform-specific code

So, you didn't see these files in Finder and other tools. 

Please change the type of project and move all platform-specific files.

PR #806, opened 2018-01-30
Fork parameter to doulbe approval and activation periods was implemented

alexeykiselev:
Tests were added.

### DIFF
--- a/it/src/main/resources/template.conf
+++ b/it/src/main/resources/template.conf
@@ -33,6 +33,7 @@ waves {
         pre-activated-features = {
           2 = 0
         }
+        double-features-periods-after-height = 100000000
       }
       genesis {
         average-block-delay: 10000ms
--- a/src/main/scala/com/wavesplatform/features/FeatureProvider.scala
+++ b/src/main/scala/com/wavesplatform/features/FeatureProvider.scala
@@ -1,12 +1,25 @@
 package com.wavesplatform.features
 
+import com.wavesplatform.settings.FunctionalitySettings
+
 trait FeatureProvider {
 
   protected def activationWindowSize(height: Int): Int
 
-  def approvedFeatures() : Map[Short, Int]
+  def approvedFeatures(): Map[Short, Int]
 
   def featureVotesCountWithinActivationWindow(height: Int): Map[Short, Int]
+}
+
+case class FeaturesProperties(functionalitySettings: FunctionalitySettings) {
+  def featureCheckBlocksPeriodAtHeight(height: Int): Int =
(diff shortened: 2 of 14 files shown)


it/src/main/resources/template.conf
```diff
@@ -33,6 +33,7 @@ waves {
         pre-activated-features = {
           2 = 0
         }
+        double-features-periods-after-height = 100000000
```
ismagin: We might tune that later again, I suggest this as "set-to-10k-blocks-at", easier to predict what happens

src/main/scala/com/wavesplatform/history/HistoryWriterImpl.scala
```diff
@@ -40,7 +37,13 @@ class HistoryWriterImpl private(db: DB, val synchronizationToken: ReentrantReadW
 
   private val HeightProperty = "history-height"
 
-  private lazy val preAcceptedFeatures = functionalitySettings.preActivatedFeatures.mapValues(h => h - activationWindowSize)
+  override def activationWindowSize(h: Int): Int = if (h > functionalitySettings.doubleFeaturesPeriodsAfterHeight)
```
ismagin: this is copypaste, should probably be reused

src/main/scala/com/wavesplatform/state2/BlockchainUpdaterImpl.scala
```diff
@@ -92,11 +92,23 @@ class BlockchainUpdaterImpl private(persisted: StateWriter with SnapshotStateRea
   private def featuresApprovedWithBlock(block: Block): Set[Short] = {
     val height = historyWriter.height() + 1
 
-    if (height % settings.blockchainSettings.functionalitySettings.featureCheckBlocksPeriod == 0) {
+    val featuresCheckPeriod =
```
ismagin: this is the same logic as in HistoryWriterImpl, duplicated as wel, right?

src/main/scala/com/wavesplatform/features/FeatureProvider.scala
```diff
@@ -1,14 +1,27 @@
...
+  protected def activationWindowSize(height: Int): Int
 
-  def approvedFeatures() : Map[Short, Int]
+  def approvedFeatures(): Map[Short, Int]
 
   def featureVotesCountWithinActivationWindow(height: Int): Map[Short, Int]
 }
 
+case class FeaturesProperties(functionalitySettings: FunctionalitySettings) {
+  def featureCheckBlocksPeriodAtHeight(height: Int): Int =
```
ismagin: could be methods of `FunctionalitySettings` itself

PR #810, opened 2018-02-01
NODE-520 API method to return all active leasing transaction

petermz:
JIRA: https://wavesplatform.atlassian.net/browse/NODE-520

A new API endpoint was introduced: GET /leasing/active/{address} . It returns list of all active outgoing leases from that address.

I've also fixed a bug in CompositeStateReader.activeLeases(): when `blockDiff.txsDiff.leaseState` and the inner state report different states for a lease, the former should win

### DIFF
--- a/src/main/scala/com/wavesplatform/state2/reader/CompositeStateReader.scala
+++ b/src/main/scala/com/wavesplatform/state2/reader/CompositeStateReader.scala
@@ -60,7 +60,9 @@ class CompositeStateReader private(inner: SnapshotStateReader, blockDiff: BlockD
     blockDiff.txsDiff.leaseState.getOrElse(leaseTx.id(), inner.isLeaseActive(leaseTx))
 
   override def activeLeases(): Seq[ByteStr] = {
-    blockDiff.txsDiff.leaseState.collect { case (id, isActive) if isActive => id }.toSeq ++ inner.activeLeases()
+    val leaseState = blockDiff.txsDiff.leaseState
+    leaseState.collect { case (id, isActive) if isActive => id }.toSeq ++
+      inner.activeLeases().filter(leaseState.getOrElse(_, true))
   }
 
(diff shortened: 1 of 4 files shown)


src/main/scala/com/wavesplatform/state2/reader/CompositeStateReader.scala
```diff
@@ -60,7 +60,9 @@ class CompositeStateReader private(inner: SnapshotStateReader, blockDiff: BlockD
     blockDiff.txsDiff.leaseState.getOrElse(leaseTx.id(), inner.isLeaseActive(leaseTx))
 
   override def activeLeases(): Seq[ByteStr] = {
-    blockDiff.txsDiff.leaseState.collect { case (id, isActive) if isActive => id }.toSeq ++ inner.activeLeases()
+    val leaseState = blockDiff.txsDiff.leaseState
+    leaseState.collect { case (id, isActive) if isActive => id }.toSeq ++
+      inner.activeLeases().filter(leaseState.getOrElse(_, true))
```
phearnot:
```
(inner.activeLeases() ++ blockDiff.txsDiff.leaseState).collect { case (id, true) => id }
```

src/main/scala/scorex/api/http/leasing/LeaseApiRoute.scala
```diff
@@ -58,4 +61,22 @@ case class LeaseApiRoute(settings: RestAPISettings, wallet: Wallet, utx: UtxPool
...
+  ))
+  def active: Route = (pathPrefix("active") & get) {
+    pathPrefix(Segment) { address => complete(
+      Address.fromString(address) match {
+        case Left(e) => ApiError.fromValidationError(e)
+        case Right(a) =>
+          state().activeLeases()
+            .flatMap(state().transactionInfo)
+            .flatMap(_._2)
+            .filter(_.asInstanceOf[LeaseTransaction].sender.address == address)
```
phearnot:
```
.collect {
  case Some((height, lt: LeaseTransaction, _)) if lt.sender.address == address => lt
}
```

PR #822, opened 2018-02-07
Update MatcherMassOrdersTestSuite

DrBlast:
Added check orders in all statuses


it/src/test/scala/com/wavesplatform/it/matcher/MatcherMassOrdersTestSuite.scala
```diff
@@ -28,15 +28,43 @@ class MatcherMassOrdersTestSuite extends FreeSpec  with NodesFromDocker with Mat
 
   private def bobNode = nodes(2)
 
-  private var aliceSellOrderId = ""
 
   private var aliceAsset: String = ""
```
ismagin: these

PR #833, opened 2018-02-12
NODE-537 Typechecker phase to completely traverse parsed expression

vsuharnikov:
A lot of changes :)


lang/src/main/scala/com/wavesplatform/lang/TypeChecker.scala
```diff
@@ -0,0 +1,212 @@
...
+
+    case some: Untyped.SOME =>
+      setType(ctx, EitherT.pure(some.t))
+        .map(_.asInstanceOf[Typed.BLOCK])
+        .map { t => Typed.SOME(t = t, tpe = OPTION(t.tpe)) }
+  }
+
+  def apply(c: Context, expr: Untyped.EXPR): TypeCheckResult[Typed.EXPR] = {
+    val result = setType(c, EitherT.pure(expr)).value().left.map { e => s"Typecheck failed: $e" }
+    Try(result) match {
```
ebceu4: Remove Try, duplicates Either

PR #868, opened 2018-02-22
Node 434 smart accounts

ismagin:


lang/shared/src/main/scala/com/wavesplatform/lang/Context.scala
```diff
@@ -0,0 +1,54 @@
...
+  val empty = Context(Map.empty, Map.empty, Map.empty)
+
+  case class CustomType(name: String, fields: List[(String, TYPE)])
+
+  sealed trait CustomFunction {
+    val name: String
+    val args: List[(String, TYPE)]
+    val resultType: TYPE
+    def eval(args: List[Any]) : Either[String,resultType.Underlying]
+    val types : (List[TYPE], TYPE)
```
vsuharnikov: `types` looks strange, because we have `args` and `resultType`. How about [signature](https://developer.mozilla.org/en-US/docs/Glossary/Signature/Function)? Also it is clearly separated from "crypto signature", because we are in the context of `CustomFunction`

lang/shared/src/main/scala/com/wavesplatform/lang/Evaluator.scala
```diff
@@ -1,169 +1,140 @@
...
-                      val updatedCtx = ctx.copy(defs = ctx.defs.updated(newVarName, (varBlockTpe, newVarValue)))
-                      r[blockTpe.Underlying](updatedCtx, inner)
+  type TypeResolutionError      = String
+  type ExcecutionError          = String
+  type ExecResult[T]            = Either[ExcecutionError, T]
+  type TrampolinedExecResult[T] = EitherT[Coeval, ExcecutionError, T]
+
+  private def r[T](ctx: Context, t: TrampolinedExecResult[Typed.EXPR]): TrampolinedExecResult[T] =
+    t flatMap { (typedExpr: Typed.EXPR) =>
+      (typedExpr match {
```
vsuharnikov: or just 
```scala
case Typed.BLOCK(mayBeLet, inner, blockTpe) =>
// ...
```

lang/shared/src/main/scala/com/wavesplatform/lang/Evaluator.scala
```diff
@@ -1,169 +1,140 @@
...
+          mayBeLet match {
+            case None => r[blockTpe.Underlying](ctx, EitherT.pure(inner))
+            case Some(Typed.LET(newVarName, newVarBlock)) =>
+              ctx.letDefs.get(newVarName) match {
+                case Some(_) => EitherT.leftT[Coeval, T](s"Value '$newVarName' already defined in the scope")
+                case None =>
+                  val varBlockTpe                                                  = newVarBlock.tpe
+                  val eitherTCoeval: TrampolinedExecResult[varBlockTpe.Underlying] = r[varBlockTpe.Underlying](ctx, EitherT.pure(newVarBlock))
+                  val calculatedValue: Coeval[varBlockTpe.Underlying] = eitherTCoeval.value.flatMap {
+                    case Left(err) => Coeval(???)
```
vsuharnikov: So, in a runtime we will see "Not implement error". I think, it rather better to save the error for the later usage: when the expression is evaluating, and Coeval contains an error, the whole expression should fail with this error.

Please create a task in Jira, if you agree with me :)

lang/shared/src/main/scala/com/wavesplatform/lang/ParserImpl.scala
```diff
@@ -31,6 +31,13 @@ abstract class ParserImpl { this: Base58 =>
   private def someP: P[SOME]           = P("Some" ~ "(" ~ block ~ ")").map(x => SOME(x))
   private def noneP: P[NONE.type]      = P("None").map(_ => NONE)
 
+  private def functionCallArgs : P[Seq[EXPR]] = (expr.rep(exactly=1) ~ P("," ~ expr).rep()).map {
```
vsuharnikov: 1. Will we have functions without arguments?
2. Could you try `expr.rep(min = 1, sep = ",")` instead of `(expr.rep(exactly=1) ~ P("," ~ expr).rep())`?

lang/shared/src/main/scala/com/wavesplatform/lang/TypeChecker.scala
```diff
@@ -46,42 +53,64 @@ object TypeChecker {
...
+          else {
+            import cats.instances.vector._
+            val actualArgTypes: Vector[SetTypeResult[Typed.EXPR]] = args.map(arg => setType(ctx, EitherT.pure(arg))).toVector
+            val sequencedActualArgTypes                           = actualArgTypes.sequence[SetTypeResult, Typed.EXPR].map(_.zip(argTypes))
+            sequencedActualArgTypes.subflatMap { v: Seq[(Typed.EXPR, TYPE)] =>
+              val matches = v.map {
+                case ((e, tpe)) =>
+                  findCommonType(e.tpe, tpe) match {
+                    case Some(_) => Right(e)
+                    case None    => Left(s"Types of arguments of function call '$name' don't match")
```
vsuharnikov: Could you add to the message which types have no common parent? It is a helpful information.

lang/shared/src/main/scala/com/wavesplatform/lang/ParserImpl.scala
ismagin: 1. don't know yet
2. yes, good idea. didn't know about `sep`

lang/shared/src/main/scala/com/wavesplatform/lang/TypeChecker.scala
ismagin: ok

lang/shared/src/main/scala/com/wavesplatform/lang/Context.scala
ismagin: yes that's better

lang/shared/src/main/scala/com/wavesplatform/lang/Evaluator.scala
ismagin: might be because of type compiler type resolution

ismagin: yep, types fail unless this

lang/shared/src/main/scala/com/wavesplatform/lang/Evaluator.scala
ismagin: yes, forgot about this `???`. you're right.

lang/shared/src/main/scala/com/wavesplatform/lang/ParserImpl.scala
ismagin: end up with `expr.rep(min = 0, sep = ",")`, much better

lang/jvm/src/test/scala/com/wavesplatform/lang/TypeCheckerTest.scala
```diff
@@ -165,6 +163,7 @@ class TypeCheckerTest extends PropSpec with PropertyChecks with Matchers with Sc
         case ref: Typed.REF   => Coeval(Untyped.REF(ref.key))
         case get: Typed.GET   => aux(get.opt).map(Untyped.GET)
         case some: Typed.SOME => aux(some.t).map(Untyped.SOME)
+        case f: Typed.FUNCTION_CALL => ???
```
vsuharnikov: We should not forget to fix this later

PR #880, opened 2018-03-01
Rewrite AliasTransactionSuite in sync way

monroid:
JIRA: https://wavesplatform.atlassian.net/browse/NODE-576


it/src/main/scala/com/wavesplatform/it/api/SyncHttpApi.scala
```diff
@@ -20,6 +22,13 @@ object SyncHttpApi extends Assertions{
     case _ => Assertions.fail(s"Expecting bad request")
   }
 
+  def assertBadRequestAndMessage2[R](f: => R, errorMessage: String): Assertion = Try(f) match {
```
ismagin: that doesn't look like proper function name :) please change it.
@DrBlast the existing one `assertBadRequest2` is not great either

monroid: @ismagin if function name will be assertBadRequestAndMessageS or assertBadRequestAndMessageSync, it is ok?

PR #886, opened 2018-03-02
Remove RequestErrorAssert trait

monroid:
JIRA: https://wavesplatform.atlassian.net/browse/NODE-580


it/src/main/scala/com/wavesplatform/it/RequestErrorAssert.scala
```diff
@@ -1,43 +1,7 @@
 package com.wavesplatform.it
 
-import akka.http.scaladsl.model.StatusCodes
```
ismagin: Please remove the whole file by moving stuff from `object` too

PR #891, opened 2018-03-07
Node 572 option matching generics

ismagin:


lang/jvm/src/test/scala/com/wavesplatform/lang/TypeCheckerTest.scala
```diff
@@ -147,13 +175,18 @@ class TypeCheckerTest extends PropSpec with PropertyChecks with Matchers with Sc
...
-                                       )),
-                    input) should produce(error)
+        TypeChecker(
+          TypeCheckerContext(
+            Map.empty,
+            Map.empty,
+            Map(
+              multiplierFunction.name -> multiplierFunction.signature,
+              noneFunc.name           -> noneFunc.signature,
+              genericFunc.name        -> genericFunc.signature
```
ismagin: Please make names more meaningful in tests too, like `idFunc`.
You can move test functon definitions where `multiplierFunction` is defined so you don't have to copy-paste the same function

lang/shared/src/main/scala/com/wavesplatform/lang/Terms.scala
```diff
@@ -75,4 +75,11 @@ object Terms {
         case _                          => None
       }
 
+  def inferTypeParams(actual: TYPE, expected: TYPE): List[(String, TYPE)] =
+    (actual, expected) match {
+      case (actualType, TYPEREF(name)) => List((name, actualType))
+      case (actualType, TYPEREF(name)) => List((name, actualType))
```
ismagin: duplication

lang/shared/src/main/scala/com/wavesplatform/lang/TypeChecker.scala
```diff
@@ -5,28 +5,39 @@ import cats.syntax.all._
...
     val empty = TypeCheckerContext(Map.empty, Map.empty, Map.empty)
 
     def fromContext(ctx: Context): TypeCheckerContext =
-      TypeCheckerContext(predefTypes = ctx.typeDefs,
-                         varDefs = ctx.letDefs.mapValues(_.tpe),
-                         functionDefs = ctx.functions.mapValues(_.signature))
+      TypeCheckerContext(
+        predefTypes = ctx.typeDefs,
+        varDefs = ctx.letDefs.mapValues(_.tpe),
+        functionDefs = ctx.functions.mapValues(x => FUNCTION(x.args.map(y => y._2), x.resultType))
```
ismagin: will `x.signature` fit here instead?

lang/shared/src/main/scala/com/wavesplatform/lang/TypeChecker.scala
```diff
@@ -5,28 +5,39 @@ import cats.syntax.all._
...
+      )
   }
 
   type TypeResolutionError      = String
   type TypeCheckResult[T]       = Either[TypeResolutionError, T]
   private type SetTypeResult[T] = EitherT[Coeval, String, T]
 
+  private def resolveTypes(typeWithParam: TYPE, map: Map[String, Either[String, TYPE]]): Either[String, TYPE] =
```
ismagin: are you sure `Map[String, Either[String, TYPE]])` is neccessary and `Map[String,TYPE]` won't be enough?

lang/shared/src/main/scala/com/wavesplatform/lang/TypeChecker.scala
```diff
@@ -55,24 +66,51 @@ object TypeChecker {
...
-                  matchType(tpe,e.tpe) match {
-                    case Some(_) => Right(e)
-                    case None    => Left(s"Types of arguments of function call '$name' do not match types required in signature. Expected: $tpe, Actual: ${e.tpe}")
-                  }
-              }
+              val typeParameters = v.flatMap(x => inferTypeParams(x._1.tpe, x._2))
+
+              val resolvedTypes = typeParameters
+                .groupBy { case (n, _) => n }
+                .map(
```
ismagin: it's hard to read, please consider
```
map { case(name, what_is_it_here) =>  ... }

PR #895, opened 2018-03-12
Added test with custom feeAsset

monroid:
JIRA: https://wavesplatform.atlassian.net/browse/NODE-567

### DIFF
--- a/it/src/main/scala/com/wavesplatform/it/api/AsyncHttpApi.scala
+++ b/it/src/main/scala/com/wavesplatform/it/api/AsyncHttpApi.scala
@@ -207,8 +207,8 @@ object AsyncHttpApi extends Assertions {
 
     def effectiveBalance(address: String): Future[Balance] = get(s"/addresses/effectiveBalance/$address").as[Balance]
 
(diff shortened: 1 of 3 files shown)


it/src/main/scala/com/wavesplatform/it/IssueSending.scala
```diff
@@ -0,0 +1,34 @@
...
+import scorex.crypto.encode.Base58
+import scorex.transaction.assets.{IssueTransaction}
+import scorex.utils.ScorexLogging
+
+
+object IssueSending {
+  case class Req(senderSeed: String, targetAddress: String, amount: Long, fee: Long)
+}
+
+trait IssueSending extends ScorexLogging {
```
ismagin: It doesn't look like `this: Suite with Nodes` is needed.
Second, `createSignedIssueRequest` is just a function converting 1 data type to another, trait and inheritance is definitely overengineering. Imagine what happens if every helper function would be imported via trait and inheritance.

Good practice is don't refactor function to a seperate class if it's used only once.
I suggest to place `createSignedIssueRequest` to `object CustomIssueTransactionTestSuite` until it gets reused.

PR #904, opened 2018-03-14
Added version check to MassTransfer transaction

petermz:


it/src/main/scala/com/wavesplatform/it/api/AsyncHttpApi.scala
```diff
@@ -211,7 +212,7 @@ object AsyncHttpApi extends Assertions {
       postJson("/assets/transfer", TransferRequest(assetId, feeAssetId, amount, fee, sourceAddress, None, recipient)).as[Transaction]
 
     def massTransfer(sourceAddress: String, transfers: List[Transfer], fee: Long, assetId: Option[String] = None): Future[Transaction] =
-      postJson("/assets/masstransfer", MassTransferRequest(assetId, sourceAddress, transfers, fee, None)).as[Transaction]
+      postJson("/assets/masstransfer", MassTransferRequest(Proofs.Version, assetId, sourceAddress, transfers, fee, None)).as[Transaction]
```
ismagin: I still don't get why you rely on `Proofs.Version` with `MassTransfer`, not `MassTransfer.Version` :)

PR #909, opened 2018-03-16
NODE-598, NODE-599: Func/var names intersection, Some/None as functions

ismagin:


ismagin:
Obsolete, see https://github.com/wavesplatform/Waves/pull/911

PR #919, opened 2018-03-21
NODE-627 Dynamic way of fee calculation was added to dtx tests

monroid:
JIRA: https://wavesplatform.atlassian.net/browse/NODE-627

### DIFF
--- a/it/src/test/scala/com/wavesplatform/it/sync/DataTransactionSuite.scala
+++ b/it/src/test/scala/com/wavesplatform/it/sync/DataTransactionSuite.scala
@@ -20,12 +20,11 @@ class DataTransactionSuite extends BaseTransactionSuite {
   test("sender's waves balance is decreased by fee.") {
     val (balance1, eff1) = notMiner.accountBalances(firstAddress)
     val entry = IntegerDataEntry("int", 0xcafebabe)
-    val size = entry.valueBytes.size
     val data = List(entry)
-    val transferFee = calcDataFee(size)
+    val transferFee = calcDataFee(data)
     val txId = sender.putData(firstAddress, data, transferFee).id
     nodes.waitForHeightAraiseAndTxPresent(txId)
-    
+
     notMiner.assertBalances(firstAddress, balance1 - transferFee, eff1 - transferFee)
   }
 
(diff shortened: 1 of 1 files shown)


it/src/test/scala/com/wavesplatform/it/sync/DataTransactionSuite.scala
```diff
@@ -253,10 +257,13 @@ class DataTransactionSuite extends BaseTransactionSuite {
     sender.postJson("/transactions/broadcast", withProof)
     nodes.waitForHeightAraiseAndTxPresent(id(withProof))
   }
-  
-  private def calcDataFee(dataSize: Int): Long = {
+
+  private def calcDataFee(data: List[DataEntry[_]]): Long = {
+    var dataSize = 0
```
ismagin: please don't use vars.
`val dataSize = data.map(_.toBytes.size).sum + 128`

it/src/test/scala/com/wavesplatform/it/sync/DataTransactionSuite.scala
```diff
@@ -90,46 +88,50 @@ class DataTransactionSuite extends BaseTransactionSuite {
...
     val item2 = BooleanDataEntry("bool", true)
-    val tx2 = sender.putData(secondAddress, List(item2), fee).id
+    val listItem2 = List(item2)
+    val tx2 = sender.putData(secondAddress, listItem2, calcDataFee(listItem2)).id
     nodes.waitForHeightAraiseAndTxPresent(tx2)
 
     sender.getData(secondAddress, "int") shouldBe item1
     sender.getData(secondAddress, "bool") shouldBe item2
-    sender.getData(secondAddress) shouldBe List(item2, item1)
+    sender.getData(secondAddress) shouldBe List.concat(listItem2, listItem1)
```
ismagin: not nessessary, but `listItem2 ++ listItem1` will do the same. Also mind the naming, `dataList1` and `dataList2` are more descriptive.

it/src/test/scala/com/wavesplatform/it/sync/DataTransactionSuite.scala
petermz: Or
```
    val size = data.foldLeft(128)(_ + _.toBytes.length)
    fee * (size / 1024 + 1)
```

PR #921, opened 2018-03-22
NODE-595: Oracle data from Smart Contract

ismagin:


lang/jvm/src/test/scala/com/wavesplatform/lang/IntegrationTest.scala
```diff
@@ -4,17 +4,22 @@ import org.scalatest.prop.PropertyChecks
 import org.scalatest.{Matchers, PropSpec}
 import com.wavesplatform.lang.Common._
 import com.wavesplatform.lang.ctx._
+import scala.reflect.runtime.universe.TypeTag
 
 class IntegrationTest extends PropSpec with PropertyChecks with Matchers with NoShrink {
 
-  private def eval(code: String) = {
+  private def eval[T: TypeTag](code: String) = {
```
vsuharnikov: I'd recommend to use `=:!=` from shapeless to force user specify T:
```scala
import shapeless._

private def eval[T: TypeTag](code: String)(implicit ev: T =:!= Nothing) = {
```

lang/shared/src/main/scala/com/wavesplatform/lang/Evaluator.scala
```diff
@@ -5,27 +5,29 @@ import com.wavesplatform.lang.Terms._
...
 import monix.eval.Coeval
 
+import scala.reflect.runtime.universe.TypeTag
+
 import scala.util.{Failure, Success, Try}
 
 object Evaluator {
 
-  private def r[T](ctx: Context, t: TrampolinedExecResult[Typed.EXPR]): TrampolinedExecResult[T] =
+  private def r[T: TypeTag](ctx: Context, t: TrampolinedExecResult[Typed.EXPR]): TrampolinedExecResult[T] =
```
vsuharnikov: Same here

lang/shared/src/main/scala/com/wavesplatform/lang/Evaluator.scala
```diff
@@ -110,10 +112,16 @@ object Evaluator {
...
+        val tt = implicitly[TypeTag[T]]
+        v match {
+          case x if typedExpr.tpe.typetag.tpe <:< tt.tpe => x.asInstanceOf[T]
+          case _                                         => throw new Exception(s"Bad type: expected: ${tt.tpe} actual: ${typedExpr.tpe.typetag.tpe}")
+        }
+      }
     }
 
-  def apply[A](c: Context, expr: Typed.EXPR): Either[ExecutionError, A] = {
+  def apply[A: TypeTag](c: Context, expr: Typed.EXPR): Either[ExecutionError, A] = {
```
vsuharnikov: And here

lang/shared/src/main/scala/com/wavesplatform/lang/ParserImpl.scala
```diff
@@ -25,7 +25,7 @@ abstract class ParserImpl { this: Base58 =>
   private val digit     = CharIn('0' to '9')
   private val varName   = (char.repX(min = 1, max = 1) ~~ (digit | char).repX()).!.filter(!keywords.contains(_))
 
-  private def numberP: P[CONST_LONG] = P(digit.rep(min = 1).!.map(t => CONST_LONG(t.toInt)))
+  private def numberP: P[CONST_LONG] = P(("-".rep(max=1) ~ digit.rep(min = 1)).!.map(t => CONST_LONG(t.toLong)))
```
vsuharnikov: If `+` is valid too, there should be:
```
CharIn("+", "-").rep(max=1)
```

lang/shared/src/main/scala/com/wavesplatform/lang/Terms.scala
```diff
@@ -10,15 +12,24 @@ object Terms {
...
-  case class TYPEREF(name: String)   extends TYPE            { type Underlying = Obj }
+  sealed abstract class TYPE_HELPER[T](implicit override val typetag: TypeTag[T]) extends TYPE {
+    override type Underlying = T
+  }
+  sealed trait TYPE extends TYPEPLACEHOLDER {
+    type Underlying
+    def typetag: TypeTag[Underlying]
+  }
+  case object NOTHING extends TYPE_HELPER[Nothing]
+  case object UNIT                   extends TYPE_HELPER[Unit] //TYPE            { type Underlying = Unit }
```
vsuharnikov: Please remove the comment

PR #949, opened 2018-03-29
implicit addressScheme

ismagin:
the problem:

PublicKeyAccount.toString, which is heavily used implicitly, requires network byte


ismagin:
not until matcher moved to separate entity

PR #987, opened 2018-04-06
Node 648 do not allow to create more assets than long.max

potan:


src/main/scala/com/wavesplatform/state2/diffs/AssetTransactionsDiff.scala
```diff
@@ -39,18 +39,20 @@ object AssetTransactionsDiff {
       tx: ReissueTransaction): Either[ValidationError, Diff] =
     validateAsset(tx, state, tx.assetId, issuerOnly = true).flatMap { _ =>
       val oldInfo = state.assetDescription(tx.assetId).get
-      if (oldInfo.reissuable || blockTime <= settings.allowInvalidReissueInSameBlockUntilTimestamp) {
+      if (!oldInfo.reissuable && blockTime > settings.allowInvalidReissueInSameBlockUntilTimestamp) {
```
alexeykiselev: May be better to replace with Either.cond?

PR #988, opened 2018-04-06
Extended CONST_LONG expressions check

monroid:
Also SctiptGen file formatted


lang/jvm/src/test/scala/com/wavesplatform/lang/ParserTest.scala
```diff
@@ -18,6 +18,11 @@ class ParserTest extends PropSpec with PropertyChecks with Matchers with ScriptG
     case Failure(_, _, _) => false
   }
 
+  def toString(expr: EXPR): String = expr match {
+    case CONST_LONG(x) => s"$x"
+    case _             => ???
+  }
```
vsuharnikov: This method should add 0-N spaces (\n, \t, ' ') from left and right of the stringified expression.
Also, please delete an unnecessary test:
```scala
parse("10") shouldBe CONST_LONG(10)
```

PR #993, opened 2018-04-09
NODE-642 Predef transaction height function

disfenoid:
[JIRA: https://wavesplatform.atlassian.net/browse/NODE-642](https://wavesplatform.atlassian.net/browse/NODE-642)


src/main/scala/scorex/transaction/smart/BlockchainContext.scala
```diff
@@ -28,6 +28,11 @@ class BlockchainContext(override val networkByte: Byte, tx: Coeval[Transaction],
       .map(_._2)
       .map(convert)
 
+  override def transactionHeightById(id: Array[Byte]): Option[Int] =
+    state
+      .transactionInfo(ByteStr(id))
+      .map(_._1)
```
phearnot: Since there's no need to parse transaction body, it might be a good idea to extend `LevelDBWriter` with a dedicated method to fetch only transaction height from leveldb.

PR #1001, opened 2018-04-10
NODE-557 Added Data tx to generator

petermz:


generator/src/main/scala/com.wavesplatform.generator/NarrowTransactionGenerator.scala
```diff
@@ -165,10 +166,35 @@ class NarrowTransactionGenerator(settings: Settings, val accounts: Seq[PrivateKe
...
+            val count  = r.nextInt(10)
+
+            val data = for {
+              _ <- 0 until count
+              keyLen = r.nextInt(10)
+              key    = Random.nextString(keyLen)
+              etype  = r.nextInt(Type.maxId)
+            } yield
+              etype match {
+                case t if t == Type.Integer.id => LongDataEntry(key, r.nextLong)
```
ismagin: Just curious, why
```
case Type.Integer.id  => 
```
won't work ?

petermz:
```
.../NarrowTransactionGenerator.scala:185:35: stable identifier required, but com.wavesplatform.state2.DataEntry.Type.Integer.id found.
[error]                 case Type.Integer.id           => LongDataEntry(key, r.nextLong)
```

PR #1002, opened 2018-04-10
Unit tests for simple expressions via Gen

monroid:


lang/jvm/src/test/scala/com/wavesplatform/lang/ParserTest.scala
```diff
@@ -20,31 +21,40 @@ class ParserTest extends PropSpec with PropertyChecks with Matchers with ScriptG
 
   def toString(expr: EXPR): String = expr match {
     case CONST_LONG(x) => whitespaces.sample.get + s"$x" + whitespaces.sample.get
-    case _             => ???
+    case TRUE          => whitespaces.sample.get + s"true" + whitespaces.sample.get
+    case FALSE         => whitespaces.sample.get + s"false" + whitespaces.sample.get
```
vsuharnikov: `s` in `s"…"` isn't required here. Don't forget to fix it in next PR.

PR #1008, opened 2018-04-11
Debug logging for request body

monroid:
For investigation tricky issues it would be great 
to get params of request body in debug logs.


it/src/main/scala/com/wavesplatform/it/api/AsyncHttpApi.scala
```diff
@@ -446,6 +446,7 @@ object AsyncHttpApi extends Assertions {
     def retrying(r: Request, interval: FiniteDuration = 1.second, statusCode: Int = HttpConstants.ResponseStatusCodes.OK_200): Future[Response] = {
       def executeRequest: Future[Response] = {
         n.log.trace(s"Executing request '$r'")
+        if (r.getStringData != null) n.log.debug(s"Request's body '${r.getStringData}'")
```
ismagin: Option(r.getStringData).foreach(data => n.log.debug(s"Request's body '$data'")

PR #1014, opened 2018-04-12
Node-653 Versioning mechanism in script

disfenoid:
[JIRA: https://wavesplatform.atlassian.net/browse/NODE-653](https://wavesplatform.atlassian.net/browse/NODE-653)


ismagin:
TypeChecker is the compiler to ExprV1, maybe it should also reside at `/v1/` dir?

disfenoid:
Now, compiler will use version 1 if version not specified, i also added a simple test for the compiler, a more complex variant requires a script (source code) generator.

PR #1017, opened 2018-04-13
NODE-602 Require min fee proportional to computational cost

vsuharnikov:
TODO
* [x] Additional base fee for transactions going from a scripted account;
* [x] Unit tests for right fee calculation;
* [x] Stack-safe implementation for ScriptComplexityCalculator;


src/main/scala/scorex/transaction/Transaction.scala
```diff
@@ -31,6 +34,12 @@ object Transaction {
         Portfolio(balance = 0, lease = LeaseBalance.empty, assets = Map(asset -> fee))
       case (None, fee) => Portfolio(balance = fee, lease = LeaseBalance.empty, assets = Map.empty)
     }
+
+    def processingScript(state: SnapshotStateReader): Option[Script] = tx match {
```
ismagin: we can remove this func. used only for script cost, `state.accountScript(tx.sender)` is enough!

vsuharnikov:
@ismagin Done!

PR #1019, opened 2018-04-16
NODE-434 Smart Contracts - Added Gen for BLOCK structure

monroid:


lang/jvm/src/test/scala/com/wavesplatform/lang/ParserTest.scala
```diff
@@ -4,27 +4,30 @@ import com.wavesplatform.lang.Common._
...
-class ParserTest extends PropSpec with PropertyChecks with Matchers with ScriptGen with NoShrink {
+class ParserTest extends PropSpec with PropertyChecks with Matchers with ScriptGenParser with NoShrink {
 
   def parse(x: String): EXPR = Parser(x).get.value
   def isParsed(x: String): Boolean = Parser(x) match {
     case Success(_, _)    => true
     case Failure(_, _, _) => false
   }
 
+  def withWhitespaces(expr: String): String = whitespaces.sample.get + expr + whitespaces.sample.get
```
ismagin: No need to use `sample`. The good way is 
```
for {
 ws1 <- whitespaces
 ws2 <- whitespaces
} yield ws1 + expr + ws2
```
that will change signature of `toString` to return `Gen[String]`

ismagin: After that, `genElementCheck` will look a little different, you might want to drop `toString` completely

lang/shared/src/main/scala/com/wavesplatform/lang/v1/testing/ScriptGen.scala
```diff
@@ -73,14 +73,28 @@ trait ScriptGen {
...
     } yield BLOCK(Some(let), body)
 
   private val spaceChars: Seq[Char] = Vector('\u0020', '\u0009', '\u000D', '\u000A')
 
   def whitespaceChar: Gen[Char] = Gen.oneOf(spaceChars)
   val whitespaces: Gen[String]  = Gen.listOf(whitespaceChar).map(_.mkString)
 }
+
+trait ScriptGenParser extends ScriptGen {
+  override def BOOLgen(gas: Int): Gen[EXPR] = {
```
ismagin: This is not needed. `ScriptGen` is not a stateful a stateful object, so no inheritance-oop is neccessary. 
I suggest the following: Ensure that you really need 2 implementations of `IntGen` if so(and that's ok!), just parameterise `BLOCKgen` so that it'd accept `INTGen` and `BOOLgen`.

monroid: yep, we need two implementations of INTgen and BOOLgen because of difference with context usage for parser and typechecker. For one we don't need references, but for another it is necessary to have valid reference to param.

lang/jvm/src/test/scala/com/wavesplatform/lang/ParserTest.scala
monroid: hm, parse function works with Strings, so from my point of view we should make transformation to String at the begin, than use Gen[String] through many functions and do final transformation only at the end

PR #1039, opened 2018-04-23
Node 129 Fair proof of stake

alexeykiselev:


src/main/scala/com/wavesplatform/consensus/PoSCalculator.scala
```diff
@@ -0,0 +1,98 @@
...
+    if (baseTarget < MinBaseTarget) MinBaseTarget else if (baseTarget > maxBaseTarget) maxBaseTarget else baseTarget
+  }
+
+}
+
+class PoSSelector(val blockchain: Blockchain) extends PoSCalculator {
+  override def hit(generatorSignature: Array[Byte]): BigInt =
+    if (fair(blockchain.height)) FairPoSCalculator.hit(generatorSignature) else NxtPoSCalculator.hit(generatorSignature)
+
+  override def time(hit: BigInt, bt: Long, balance: Long): Long =
```
ismagin: is there a way to make through function composition rather then inheritance? everything is pure here, no need to use overriding. `Strategy` pattern wll do that too :)

PR #1040, opened 2018-04-23
Node 687 sponsored transactions

potan:


src/main/scala/com/wavesplatform/state/Diff.scala
```diff
@@ -71,14 +76,46 @@ object AccountDataInfo {
...
 case class Diff(transactions: Map[ByteStr, (Int, Transaction, Set[Address])],
                 portfolios: Map[Address, Portfolio],
                 issuedAssets: Map[AssetId, AssetInfo],
                 aliases: Map[Alias, Address],
                 orderFills: Map[ByteStr, VolumeAndFee],
                 leaseState: Map[ByteStr, Boolean],
                 scripts: Map[Address, Option[Script]],
-                accountData: Map[Address, AccountDataInfo]) {
+                accountData: Map[Address, AccountDataInfo],
+                sponsorship: Map[AssetId, Sponsorship]) {
```
ismagin: We already have `issuedAssets: Map[AssetId, AssetInfo]` why can't `Sponsorship be part of `AssetInfo` ?

src/main/scala/com/wavesplatform/database/Caches.scala
```diff
@@ -4,12 +4,25 @@ import java.util
 
 import cats.syntax.monoid._
 import com.google.common.cache.{CacheBuilder, CacheLoader, LoadingCache}
-import com.wavesplatform.state.{AccountDataInfo, AssetDescription, AssetInfo, Blockchain, ByteStr, Diff, LeaseBalance, Portfolio, VolumeAndFee}
+import com.wavesplatform.state.{
```
ismagin: just wonder why not `com.wavesplatform.state._`?

src/main/scala/com/wavesplatform/state/Diff.scala
```diff
@@ -71,14 +76,46 @@ object AccountDataInfo {
...
+  def sponsoredFeesSwitchHeight(blockchain: Blockchain, fs: FunctionalitySettings): Int =
+    blockchain
+      .featureActivationHeight(BlockchainFeatures.FeeSponsorship.id)
+      .map(h => h + fs.blocksForFeatureActivation(h))
+      .getOrElse(Int.MaxValue)
+
+  def toWaves(assetFee: Long, sponsorship: Long): Long = {
+    val waves = (BigDecimal(assetFee) * BigDecimal(Sponsorship.FeeUnit)) / BigDecimal(sponsorship)
+    if (waves > Long.MaxValue) {
+      throw new java.lang.ArithmeticException("Overflow")
```
ismagin: why throw, not `[Either..]` ? is is caught later?

PR #1048, opened 2018-04-25
NODE-604 Limit size/complexity of script

vsuharnikov:
TODO:
* [x] Extra fee is fixed
* [x] Extra fee is encoded in consensus (after sponsored transactions), no settings required
* [x] Simplify complexity;
* [x] Limit a complexity
* [x] Limit size of the script
* [x] Sponsorship should work with smart accounts
* [x] Fee for smart accounts in CommonValidation
* [x] Test how works Sponsorship works with smart accounts
* [x] Benchmark for ScriptEstimator

### DIFF
--- a/src/main/scala/scorex/transaction/smart/script/ScriptReader.scala
+++ b/src/main/scala/scorex/transaction/smart/script/ScriptReader.scala
@@ -30,7 +30,7 @@ object ScriptReader {
             .flatMap { _ =>
               Serde.codec.decode(scodec.bits.BitVector(scriptBytes)) match {
                 case Failure(e)    => Left(e.toString())
-                case Successful(x) => ScriptV1(x.value, checkBytes = false)
+                case Successful(x) => ScriptV1(x.value, checkSize = false)
               }
             }
             .left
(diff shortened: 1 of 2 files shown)


src/test/scala/scorex/transaction/smart/script/ScriptV1Test.scala
```diff
@@ -21,4 +26,55 @@ class ScriptV1Test extends PropSpec with PropertyChecks with Matchers with Typed
...
+        FUNCTION_CALL(
+          function = FunctionHeader(name = "sigVerify", List(FHT.BYTEVECTOR, FHT.BYTEVECTOR, FHT.BYTEVECTOR)),
+          args = List(byteVector, byteVector, byteVector),
+          BOOLEAN
+        )
+      }
+      .reduceLeft[EXPR](IF(_, _, FALSE, BOOLEAN))
+
+    val r = ScriptV1(expr)
+    r shouldBe 'left
```
ismagin: it's better to assert error message via `should produce ...`

PR #1054, opened 2018-04-27
s/minAssetFee/minSponsoredAssetFee/g

potan:
Renaming


monroid:
commit name contains "/", it isn't good idea

PR #1075, opened 2018-05-07
Node 717 for mass transfer transaction user cannot get predefined data structures for recipient

potan:


ismagin:
I strongly suggest e try to do this without introducing new built-in type but rather go with unions and migrate option there too.

PR #1080, opened 2018-05-08
DataTransaction v2

disfenoid:


ismagin:
Need to discuss this idea first, might have better solution

PR #1085, opened 2018-05-10
Evaluator Optimizations

disfenoid:
StateT and Logging removed


lang/shared/src/main/scala/com/wavesplatform/lang/v1/evaluator/EvalM.scala
```diff
@@ -0,0 +1,64 @@
...
+  }
+}
+
+object EvalM {
+
+  implicit val monadInstance: Monad[EvalM] = new StackSafeMonad[EvalM] {
+    override def pure[A](x: A): EvalM[A] =
+      EvalM(Kleisli.pure[Coeval, CoevalRef[EvaluationContext], Either[ExecutionError, A]](x.asRight[ExecutionError]))
+
+    override def flatMap[A, B](fa: EvalM[A])(f: A => EvalM[B]): EvalM[B] = {
```
ismagin: is it stack-safe?

disfenoid: It's pass stack-safety check in tests (successful evaluation on long expression) but i'm not 100% sure , i will try to implement tailRecM later.

PR #1090, opened 2018-05-14
NODE-741 Added LevelDB snapshot store

phearnot:


src/main/scala/com/wavesplatform/matcher/MatcherSnapshotStore.scala
```diff
@@ -0,0 +1,204 @@
...
+    val rw = new RW(writableDB)
+    try f(rw)
+    finally rw.close()
+  }
+
+  private var pendingActions = Map.empty[String, Future[_]]
+
+  private def nextAction[A](persistenceId: String, f: => A): Future[A] = {
+    val newAction = pendingActions
+      .getOrElse(persistenceId, Future(())(streamDispatcher))
```
vsuharnikov: Small tip: `Future.successful` is better in such situations, because it doesn't run any computation. Also it doesn't require a scheduler

PR #1092, opened 2018-05-14
NODE-742 Parsing of invalid expressions

vsuharnikov:
- [x] Parsing of invalid expressions (in general);
- [x] Correct parsing of string containing unicode and special symbols;
- [x] Invalid strings handling: invalid/incomplete unicode and special symbols;
- [x] Invalid Base58 handling;
- [x] Lightweight definition of expressions with PART;
- [x] CompilerV1 - better PART handling;
- [x] Additional tests for CompilerV1;
- [x] JsAPI.parse/compile works with sequences;
- [x] Rebase after parser changes;
- [x] maybeAccessP;
- [x] Parsing of invalid pattern matching.

### DIFF
--- a/lang/shared/src/main/scala/com/wavesplatform/lang/v1/parser/Expressions.scala
+++ b/lang/shared/src/main/scala/com/wavesplatform/lang/v1/parser/Expressions.scala
@@ -4,20 +4,70 @@ import scodec.bits.ByteVector
 
 object Expressions {
 
-  case class LET(name: String, value: EXPR, types: Seq[String] = Seq.empty)
+  sealed trait PART[+T]
+  object PART {
+    case class VALID[T](v: T)                             extends PART[T]
+    case class INVALID(consumed: String, message: String) extends PART[Nothing]
+  }
+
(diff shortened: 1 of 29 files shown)


lang/shared/src/main/scala/com/wavesplatform/lang/v1/parser/Expressions.scala
```diff
@@ -4,20 +4,70 @@ import scodec.bits.ByteVector
...
+
   sealed trait EXPR
-  case class CONST_LONG(value: Long)                            extends EXPR
-  case class GETTER(ref: EXPR, field: String)                   extends EXPR
-  case class CONST_BYTEVECTOR(value: ByteVector)                extends EXPR
-  case class CONST_STRING(value: String)                        extends EXPR
+  case class CONST_LONG(value: Long) extends EXPR
+
+  case class GETTER(ref: EXPR, field: PART[String]) extends EXPR
+  object GETTER {
```
vsuharnikov: Companion objects will be removed in next PR for some reasons

PR #1101, opened 2018-05-16
Node 129 fair proof of stake

disfenoid:
[JIRA: https://wavesplatform.atlassian.net/browse/NODE-129](https://wavesplatform.atlassian.net/browse/NODE-129)


src/main/scala/com/wavesplatform/consensus/PoSCalculator.scala
```diff
@@ -0,0 +1,140 @@
...
+
+trait PoSCalculator {
+  protected val HitSize: Int              = 8
+  protected val MeanCalculationDepth: Int = 3
+
+  // Min BaseTarget value is 9 because only in this case it is possible to get to next integer value (10)
+  // then increasing base target by 11% and casting it to Long afterward (see lines 55 and 59)
+  protected val MinBaseTarget: Long = 9
+
+  protected val MinBlockDelaySeconds = 53
```
ismagin: move some of these to PosCalculator where they are actually used

it/src/main/scala/com/wavesplatform/it/BaseTargetChecker.scala
```diff
@@ -3,43 +3,42 @@ package com.wavesplatform.it
...
       case cfg if cfg.as[Boolean]("waves.miner.enable") =>
-        val publicKey = PublicKeyAccount(cfg.as[ByteStr]("public-key").arr)
-        val address   = publicKey.toAddress
-        PoSCalc.nextBlockGenerationTime(1, bu, fs, genesisBlock, publicKey) match {
-          case Right((_, ts)) => f"$address: ${(ts - startTs) * 1e-3}%10.3f s"
-          case _              => s"$address: n/a"
-        }
+        val account   = PublicKeyAccount(cfg.as[ByteStr]("public-key").arr)
+        val address   = account.toAddress
+        val balance   = bu.balance(address, None)
```
vsuharnikov: should it be an effective balance? or it doesn't make sense here?

src/main/scala/com/wavesplatform/consensus/PoSCalculator.scala
```diff
@@ -0,0 +1,140 @@
...
+  def validBlockDelay(genSig: Array[Byte], publicKey: Array[Byte], baseTarget: Long, balance: Long): Long = {
+    calculateDelay(hit(generatorSignature(genSig, publicKey)), baseTarget, balance)
+  }
+
+  protected def normalize(value: Long, targetBlockDelaySeconds: Long): Double =
+    value * targetBlockDelaySeconds / (60: Double)
+
+  protected def normalizeBaseTarget(baseTarget: Long, targetBlockDelaySeconds: Long): Long = {
+    val maxBaseTarget = Long.MaxValue / targetBlockDelaySeconds
+    if (baseTarget < MinBaseTarget) MinBaseTarget else if (baseTarget > maxBaseTarget) maxBaseTarget else baseTarget
```
vsuharnikov: Small tip: it could be simpler
```scala
else math.min(maxBaseTarget, baseTarget)
```
or even simpler, but please doesn't change, it's ok

src/test/resources/genesis.it.conf
```diff
@@ -1,25 +1,21 @@
...
-    "node07": 40000000000000
-    "node08": 50000000000000
-    "node09": 60000000000000
-    "node10": 830500000000000
+    "create genesis wallet devnet-0": 200000000000000
+    "create genesis wallet devnet-0-d": 500000000000000
+    "create genesis wallet devnet-1": 1000000000000000
+    "create genesis wallet devnet-1-d": 1500000000000000
+    "create genesis wallet devnet-2": 2000000000000000
+    "create genesis wallet devnet-2-d": 2500000000000000
```
vsuharnikov: Hm… please rollback these changes, it is for integration tests. But you could create `genesis.dev.conf`

src/test/scala/com/wavesplatform/consensus/FairPoSCalculatorTest.scala
```diff
@@ -0,0 +1,72 @@
...
+
+  case class Block(height: Int, baseTarget: Long, timestamp: Long, delay: Long)
+
+  def generationSignature: Array[Byte] = {
+    val arr = new Array[Byte](32)
+    Random.nextBytes(arr)
+    arr
+  }
+
+  property("") {
```
vsuharnikov: What's the property? :)

src/main/scala/com/wavesplatform/consensus/PoSCalculator.scala
phearnot: `baseTarget.max(MinBaseTarget).min(maxBaseTarget)`

PR #1102, opened 2018-05-16
NODE-758: string entry for data transaction

ismagin:


it/src/test/scala/com/wavesplatform/it/sync/DataTransactionSuite.scala
```diff
@@ -106,6 +106,12 @@ class DataTransactionSuite extends BaseTransactionSuite {
     val tx2       = sender.putData(secondAddress, boolList, calcDataFee(boolList)).id
     nodes.waitForHeightAriseAndTxPresent(tx2)
 
+    // define boolean entry
```
petermz: this comment lies

src/main/scala/com/wavesplatform/state/DataEntry.scala
```diff
@@ -71,7 +76,12 @@ object DataEntry {
                 case JsDefined(JsString(base58)) =>
                   val t = if (base58.isEmpty) Success(Array.emptyByteArray) else Base58.decode(base58) /// Base58 bug
                   t.fold(ex => JsError(ex.getMessage), arr => JsSuccess(BinaryDataEntry(key, ByteStr(arr))))
-                case _ => JsError("value is missing or not a string")
+                case _ => JsError("value is missing or not a base58")
```
petermz: not-base58 case is handled in line 77, this line handles case when value is not a string, e.g. 8 or true

src/main/scala/com/wavesplatform/state/DataEntry.scala
```diff
@@ -103,3 +113,11 @@ case class BinaryDataEntry(override val key: String, override val value: ByteStr
 
   override def valid: Boolean = super.valid && value.arr.length <= MaxValueSize
 }
+
+case class StringDataEntry(override val key: String, override val value: String) extends DataEntry[String](key, value) {
+  override def valueBytes: Array[Byte] = Type.String.id.toByte +: Deser.serializeArray(value.getBytes("UTF-8"))
```
petermz: We have a constant named UTF8, better to use it for performance

PR #1104, opened 2018-05-17
Support new patterns

potan:
case _ : XXX | YYY => ... // Don't create new variable.
 case _ => ...  // All other type.

### DIFF
--- a/lang/jvm/src/test/scala/com/wavesplatform/lang/ParserTest.scala
+++ b/lang/jvm/src/test/scala/com/wavesplatform/lang/ParserTest.scala
@@ -18,6 +18,7 @@ class ParserTest extends PropSpec with PropertyChecks with Matchers with ScriptG
   def parse(x: String): EXPR = Parser(x) match {
     case Success(r, _) => r
     case e @ Failure(_, i, _) =>
+      println(x)
       println(
         s"Can't parse (len=${x.length}): <START>\n$x\n<END>\nError: $e\nPosition ($i): '${x.slice(i, i + 1)}'\nTraced:\n${e.extra.traced.fullStack
           .mkString("\n")}")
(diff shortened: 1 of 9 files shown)


lang/jvm/src/test/scala/com/wavesplatform/lang/ParserTest.scala
```diff
@@ -278,9 +279,9 @@ class ParserTest extends PropSpec with PropertyChecks with Matchers with ScriptG
     isParsed("match tx { case => } ") shouldBe false
     isParsed("match tx { case => 1} ") shouldBe false
     isParsed("match tx { case TypeA => } ") shouldBe false
-    isParsed("match tx { case TypeA => 1 } ") shouldBe false
     isParsed("match tx { case  :TypeA => 1 } ") shouldBe false
-    isParsed("match tx { case  x => 1 } ") shouldBe false
+    isParsed("match tx { case  x => 1 } ") shouldBe true
```
ismagin: please check what's really parsed here

lang/shared/src/main/scala/com/wavesplatform/lang/v1/compiler/CompilerV1.scala
```diff
@@ -208,22 +208,28 @@ object CompilerV1 {
         case _        => EitherT.fromEither[Coeval](Left("Only union type can be matched"))
       }
       matchingTypes = cases.flatMap(_.types)
-      _ <- EitherT.cond[Coeval](UNION.eq(u, UNION(matchingTypes.toList.map(CASETYPEREF))),
+      val lastEmpty = cases.last.types.isEmpty
+      _ <- EitherT.cond[Coeval](lastEmpty || UNION.eq(u, UNION(matchingTypes.toList.map(CASETYPEREF))),
```
ismagin: not strict enough. if last is empty, matchching types must be subset of possible types, if not empty, it should be exactly the same set.

PR #1118, opened 2018-05-22
NODE-575 Fee from one object

monroid:


it/src/test/scala/com/wavesplatform/it/sync/package.scala
```diff
@@ -0,0 +1,25 @@
...
+import com.wavesplatform.it.util._
+
+package object sync {
+  val fee                        = 0.001.waves
+  val smartFee                   = 0.004.waves
+  val issueFee                   = 1.waves
+  val transferAmount             = 10.waves
+  val leasingAmount              = transferAmount
+  val issueAmount                = transferAmount
+  val massTransferFeePerTransfer = 0.0005.waves
```
vsuharnikov: It's hard to find a such variable during the code reading (snakeCase is for local variables). It's better to use ConstStyleNaming for such variables.

vsuharnikov: Please rename it and merge the PR ;)

PR #1136, opened 2018-05-28
Implements multiplication and division in smartcontracts language.

potan:


ismagin:
I suggest we just reimplement that from scratch

PR #1144, opened 2018-05-30
NODE-779: Return EvaluationContext in case of unsuccessful evaluation…

disfenoid:
[JIRA: https://wavesplatform.atlassian.net/browse/NODE-779](https://wavesplatform.atlassian.net/browse/NODE-779)


ismagin:
Please write a test, showing that given a multisig script:
```
let a = sigerify ...
let b = sigverify ...
if a then b else false
```
and wrong A signature, user can observe that "a = false,  b = not calculated".
integration test seems to be good fit, maybe even editing existing one.

PR #1145, opened 2018-05-30
WIP: DON'T MERGE : sponsorship-rollback

monroid:


monroid:
After rollback of all nodes for (n <- nodes) n.rollback(height, false) asset configuration returns to default.
Sponsorship rollback works fine.

PR #1150, opened 2018-06-01
NODE-747 Support Sponsorship rollback in tests

monroid:


it/src/test/scala/com/wavesplatform/it/sync/RollbackSuite.scala
```diff
@@ -11,7 +11,14 @@ import scala.concurrent.duration._
 import scala.concurrent.Await
 
 class RollbackSuite extends FunSuite with CancelAfterFailure with TransferSending with NodesFromDocker with Matchers {
-  override protected val nodeConfigs: Seq[Config] = Seq(NodeConfigs.randomMiner, NodeConfigs.NotMiner)
+  override def nodeConfigs: Seq[Config] =
```
ismagin: can you just preactivate the feature for all tests so no need to override config here?

monroid: Since sponsorship feature doesn't use for transfers, I removed the config lines. For other tests we decided do not use sponsorship by default.

PR #1163, opened 2018-06-05
NODE-789-remove-types-from-function-header

potan:
Make runtime functions names unique.


lang/shared/src/main/scala/com/wavesplatform/lang/v1/evaluator/ctx/PredefFunction.scala
```diff
@@ -19,32 +19,33 @@ trait PredefFunction {
...
 
-  abstract class PredefFunctionImplT(val name: String,
-                                 val cost: Long,
-                                 val resultType: TYPEPLACEHOLDER,
-                                 val args: List[(String, TYPEPLACEHOLDER)])
-      extends PredefFunction {
-    override val signature              = FunctionTypeSignature(args.map(_._2), resultType)
-    override val header: FunctionHeader = FunctionHeader(name, args.map(_._2).map(FunctionHeader.FunctionHeaderType.fromTypePlaceholder))
-  }
+//  abstract class PredefFunctionImplT(val name: String,
```
ismagin: clean-up

lang/shared/src/main/scala/com/wavesplatform/lang/v1/compiler/CompilerV1.scala
```diff
@@ -192,7 +191,7 @@ object CompilerV1 {
       for {
         resolvedTypeParams <- TypeInferrer(typePairs).leftMap(Generic(start, end, _))
         resolvedResultType <- TypeInferrer.inferResultType(f.result, resolvedTypeParams).leftMap(Generic(start, end, _))
-        header = FunctionHeader(funcName, f.args.map(FunctionHeaderType.fromTypePlaceholder))
+        header = FunctionHeader(f.name)
```
ismagin: are you sure? why not header = f.header `or `header = FunctionHeader(f.innerName)`?
if that's an error, please write a test

lang/shared/src/main/scala/com/wavesplatform/lang/v1/evaluator/ctx/EvaluationContext.scala
```diff
@@ -22,8 +22,13 @@ object EvaluationContext {
       EvaluationContext(letDefs = x.letDefs ++ y.letDefs, functions = x.functions ++ y.functions)
   }
 
-  def build(letDefs: Map[String, LazyVal], functions: Seq[PredefFunction]): EvaluationContext =
+  def build(letDefs: Map[String, LazyVal], functions: Seq[PredefFunction]): EvaluationContext = {
+    if(functions.map(_.header.name).distinct.size != functions.size) {
```
ismagin: maybe check other things like `let` intersections, let-function intersection too?

lang/shared/src/main/scala/com/wavesplatform/lang/v1/evaluator/ctx/impl/PureContext.scala
```diff
@@ -19,70 +19,70 @@ object PureContext {
...
 
-  val isDefined: PredefFunction = PredefFunction("isDefined", 1, BOOLEAN, List(("opt", optionT))) {
+  val isDefined: PredefFunction = PredefFunction("isDefined", 1, BOOLEAN, List(("opt", optionT)), "isDefined") {
     case Some(_) :: Nil => Right(true)
     case None :: Nil    => Right(false)
     case _              => ???
   }
 
-  val size: PredefFunction = PredefFunction("size", 1, LONG, List(("byteVector", BYTEVECTOR))) {
+  val size: PredefFunction = PredefFunction("size", 1, LONG, List(("byteVector", BYTEVECTOR)), "_vs") {
```
ismagin: why "size" and "_vs" ? 
why not call them the same? the idea of "_" is that call to functions which start with underscore can't be parsed in parser

PR #1168, opened 2018-06-05
NODE-795 Additional operators

disfenoid:
[JIRA: https://wavesplatform.atlassian.net/browse/NODE-795](https://wavesplatform.atlassian.net/browse/NODE-795)

### DIFF
--- a/lang/shared/src/main/scala/com/wavesplatform/lang/v1/evaluator/ctx/impl/PureContext.scala
+++ b/lang/shared/src/main/scala/com/wavesplatform/lang/v1/evaluator/ctx/impl/PureContext.scala
@@ -19,6 +19,17 @@ object PureContext {
   val err           = LazyVal(EitherT(nothingCoeval))
   val errRef        = "throw"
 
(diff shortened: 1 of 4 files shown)


lang/shared/src/main/scala/com/wavesplatform/lang/v1/evaluator/ctx/impl/PureContext.scala
```diff
@@ -19,6 +19,11 @@ object PureContext {
   val err           = LazyVal(EitherT(nothingCoeval))
   val errRef        = "throw"
 
+  val fraction: PredefFunction = PredefFunction("fraction", 1, LONG, List(("value", LONG), ("numerator", LONG), ("denominator", LONG))) {
+    case (v: Long) :: (n: Long) :: (d: Long) :: Nil => Right((BigInt(v) * n / d).toLong)
```
ismagin: can toLong overflow? please add tests that 

fraction (Long.MaxValue, 2,  3) doesn't overflow,
fraction (Long.MaxVaue,3,2) overflows (doesn't return negative)

PR #1190, opened 2018-06-07
NODE-818 User functions

vsuharnikov:


lang/shared/src/main/scala/com/wavesplatform/lang/v1/evaluator/ctx/impl/waves/WavesContext.scala
```diff
@@ -18,36 +19,88 @@ object WavesContext {
...
+          FUNCTION_CALL(
+            FunctionHeader.Native(ADDRESSFROMBYTES),
+            List(
+              BLOCK(
+                LET(
+                  "@afpk_withoutChecksum",
+                  FUNCTION_CALL(
+                    FunctionHeader.Native(SUM_BYTES),
+                    List(
+                      CONST_BYTEVECTOR(ByteVector(EnvironmentFunctions.AddressVersion, env.networkByte)),
```
vsuharnikov: I'll reference `EnvironmentFunctions.AddressVersion` and other constants in RIDE's variables in a next PR.

lang/shared/src/main/scala/com/wavesplatform/lang/v1/evaluator/FunctionIds.scala
```diff
@@ -3,60 +3,60 @@ package com.wavesplatform.lang.v1.evaluator
...
-  val GETTRANSACTIONBYID: Short = 135
-  val ACCOUNTBALANCE: Short = 136
-  val ACCOUNTASSETBALANCE: Short = 137
-  val TRANSACTIONHEIGHTBYID: Short = 138
+  val DATA_LONG: Short    = 650
+  val DATA_BOOLEAN: Short = 651
+  val DATA_BYTES: Short   = 652
+  val DATA_STRING: Short  = 653
+
+  val ADDRESSFROMPUBKEY: Short    = 700
```
vsuharnikov: Forgot about it :( Will remove in next PR

PR #1197, opened 2018-06-08
NODE-809 Part 2: issue with parsing of incomplete block in a pattern matching 

vsuharnikov:


lang/shared/src/main/scala/com/wavesplatform/lang/v1/parser/Parser.scala
```diff
@@ -101,36 +101,54 @@ object Parser {
...
+      P(Index ~~ "then" ~/ Index ~ baseExpr.? ~ elsePart ~~ Index).map {
+        case (start, ifTrueExprPos, ifTrueRaw, ifFalse, end) =>
+          val ifTrue = ifTrueRaw.getOrElse(INVALID(ifTrueExprPos, ifTrueExprPos, "expected a true branch's expression"))
+          IF(start, end, INVALID(start, start, "expected a condition"), ifTrue, ifFalse)
+      } |
+      P(Index ~~ "else" ~/ Index ~ baseExpr.? ~~ Index).map {
+        case (start, ifFalseExprPos, ifFalseRaw, end) =>
+          val ifFalse = ifFalseRaw.getOrElse(INVALID(ifFalseExprPos, ifFalseExprPos, "expected a false branch's expression"))
+          IF(start, end, INVALID(start, start, "expected a condition"), INVALID(start, start, "expected a true branch"), ifFalse)
+      }
```
vsuharnikov: Sad, but we can't simplify this expression, because we need a greedy parsing for incomplete IF expressions

PR #1198, opened 2018-06-08
require delimiter after keyword

potan:


lang/shared/src/main/scala/com/wavesplatform/lang/v1/parser/Parser.scala
```diff
@@ -131,17 +131,17 @@ object Parser {
     val thenPart = optionalPart("then", "true")
     val elsePart = optionalPart("else", "false")
 
-    P(Index ~~ "if" ~/ Index ~ baseExpr.? ~ thenPart ~ elsePart ~~ Index).map {
+    P(Index ~~ "if" ~~ &(CharIn(" \t\n\r({")) ~/ Index ~ baseExpr.? ~ thenPart ~ elsePart ~~ Index).map {
```
vsuharnikov: Looks like i should be a separate variable

PR #1199, opened 2018-06-09
NODE-785 Added Sender's address to proven part

monroid:


lang/shared/src/main/scala/com/wavesplatform/lang/v1/evaluator/ctx/impl/waves/Types.scala
```diff
@@ -24,6 +24,7 @@ object Types {
     "version"   -> LONG,
   )
   val proven = List(
+    "sender"    -> STRING,
```
ismagin: sender should be `Address`, not string

PR #1201, opened 2018-06-09
NODE-799 Added names and description comparison

monroid:


monroid:
Comparison doesn't work in such annotation:
let name = t.name == base58'${ByteStr(t.name).base58}'
let description = t.description == base58'${ByteStr(t.description).base58}'

what I did incorrect?

PR #1210, opened 2018-06-13
NODE-749 Shadowing verification with transaction gens

monroid:


src/test/scala/com/wavesplatform/state/diffs/smart/predef/CommonFunctionsTest.scala
```diff
@@ -105,4 +106,45 @@ class CommonFunctionsTest extends PropSpec with PropertyChecks with Matchers wit
...
+    }
+  }
+
+  property("negative shadowing verification") {
+    forAll(Gen.oneOf(transferV2Gen, issueGen, massTransferGen(10))) {
+      case (transfer) =>
+        try {
+          runScript[Boolean](
+            s"""
+               |let tx = 100
```
vsuharnikov: Should be:
```
let t = 100
match tx {
  case  t: TransferTransaction =>
  // ...
```
becase an error is about this line.

monroid:
@ismagin for now we agreed put shadowing case in CommonFunctionsTest.scala

PR #1213, opened 2018-06-14
NODE-785 Compatibility with testnet addresses

monroid:
- renamed senderPk to senderPublicKey (need careful review)

### DIFF
--- a/lang/shared/src/main/scala/com/wavesplatform/lang/v1/evaluator/ctx/impl/waves/Bindings.scala
+++ b/lang/shared/src/main/scala/com/wavesplatform/lang/v1/evaluator/ctx/impl/waves/Bindings.scala
@@ -18,9 +18,9 @@ object Bindings {
 
   private def provenTxPart(tx: Proven): Map[String, Any] =
     Map(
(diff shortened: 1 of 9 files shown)


lang/shared/src/main/scala/com/wavesplatform/lang/v1/evaluator/ctx/impl/waves/Bindings.scala
```diff
@@ -18,9 +18,9 @@ object Bindings {
 
   private def provenTxPart(tx: Proven): Map[String, Any] =
     Map(
-      "sender"    -> tx.sender,
-      "senderPk"  -> tx.senderPk,
-      "bodyBytes" -> tx.bodyBytes,
+      "sender"          -> CaseObj(addressType.typeRef, Map("bytes" -> tx.sender.bytes)),
```
ismagin: This is a duplicate mapping as of MapRecipient, it's best to avoid that.
Another case where this mapping is needed is `order.sender/senderPk`, so let's refactor that to separate function

PR #1215, opened 2018-06-14
NODE-575 Rewrite Async tests in Sync way

DrBlast:
payment, reissue,
moved smartcontracts to another package


it/src/main/scala/com/wavesplatform/it/api/AsyncHttpApi.scala
```diff
@@ -93,6 +93,9 @@ object AsyncHttpApi extends Assertions {
     def getOrderStatus(asset: String, orderId: String): Future[MatcherStatusResponse] =
       matcherGet(s"/matcher/orderbook/$asset/WAVES/$orderId", waitForStatus = true).as[MatcherStatusResponse]
 
+    def getOrderStatus1(asset: String, orderId: String): Future[MatcherStatusResponse] =
```
phearnot: Instead of declaring a new method, please just add a default argument to the original one.

PR #1230, opened 2018-06-18
Merged NODE-830 and NODE-842 from version-0.13.x

petermz:


petermz:
Testnet blockchain import fails after block 331300 with `GenericError(Duplicate keys found)`

PR #1243, opened 2018-06-21
NODE-834 Node connected to itself

petermz:
JIRA: https://wavesplatform.atlassian.net/browse/NODE-834

I'm filtering outgoing connections using a list of IPs known to be local. In case that doesn't help (due to some kind of IP masquerading) I'm additionally filtering incoming connections by comparing nonces on both ends.


petermz:
I thought the real fix would be not to connect to wildcard addresses at all

PR #1277, opened 2018-07-02
NODE-895 User functions cost to be calculated not hardcoded

petermz:
JIRA: https://wavesplatform.atlassian.net/browse/NODE-895

Costs are calculated for all user functions at once in
com/wavesplatform/utils/package.scala

### DIFF
--- a/src/main/scala/com/wavesplatform/utils/package.scala
+++ b/src/main/scala/com/wavesplatform/utils/package.scala
@@ -4,9 +4,11 @@ import cats.kernel.Monoid
 import com.google.common.base.Throwables
 import com.wavesplatform.db.{Storage, VersionedStorage}
 import com.wavesplatform.lang.Global
+import com.wavesplatform.lang.v1.{FunctionHeader, ScriptEstimator}
 import com.wavesplatform.lang.v1.compiler.CompilerContext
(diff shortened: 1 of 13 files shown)


src/main/scala/com/wavesplatform/utils/package.scala
```diff
@@ -80,8 +83,28 @@ package object utils extends ScorexLogging {
...
   }
 
-  lazy val dummyNetworkByte: Byte                    = AddressScheme.current.chainId
-  lazy val dummyEvaluationContext: EvaluationContext = BlockchainContext.build(dummyNetworkByte, Coeval(???), Coeval(???), null)
+  lazy val dummyNetworkByte: Byte                           = AddressScheme.current.chainId
+  lazy val dummyEvaluationContext: EvaluationContext        = BlockchainContext.build(dummyNetworkByte, Coeval(???), Coeval(???), null)
+  lazy val functionCosts: Map[FunctionHeader, Coeval[Long]] = estimate(dummyEvaluationContext)
+
+  def estimate(ctx: EvaluationContext): Map[FunctionHeader, Coeval[Long]] = {
+    val costs: mutable.Map[FunctionHeader, Coeval[Long]] = ctx.typeDefs.collect {
```
ismagin: what's the reason for mutability here?

PR #1282, opened 2018-07-03
NODE-803 All functions from context validation

monroid:


src/test/scala/com/wavesplatform/state/diffs/smart/predef/ContextFunctionsTest.scala
```diff
@@ -0,0 +1,166 @@
...
+       | let sig = sigVerify(base58'333', base58'123', base58'567') != true
+       | let str58 = fromBase58String(toBase58String(tx.id)) == tx.id
+       | let str64 = fromBase64String(toBase64String(tx.id)) == tx.id
+       |
+       | let crypto = bks && sig && str58 && str64
+       |
+       | pure && waves && crypto
+       |
+    """.stripMargin
+
```
petermz: Please use || somewhere :)

vsuharnikov:
Also we need to check the backward compatibility for functions. From task:
> tests for serialization backward compatibility

src/test/scala/com/wavesplatform/state/diffs/smart/predef/ContextFunctionsTest.scala
monroid: added to line 95

PR #1300, opened 2018-07-09
NODE-931-price-assets-api

potan:
Add API to get configuration parameter price-assets.
Remove assets-pairs support.

### DIFF
--- a/src/main/scala/com/wavesplatform/matcher/api/MatcherApiRoute.scala
+++ b/src/main/scala/com/wavesplatform/matcher/api/MatcherApiRoute.scala
@@ -63,7 +63,7 @@ case class MatcherApiRoute(wallet: Wallet,
   @ApiOperation(value = "Matcher Settings", notes = "Get matcher settings", httpMethod = "GET")
   def getSettings: Route = (path("settings") & get) {
     val priceAssets = matcherSettings.priceAssets
-    complete(StatusCodes.OK -> Json.obj("settings" -> Json.obj("price-assets" -> priceAssets)))
+    complete(StatusCodes.OK -> Json.obj("priceAssets" -> priceAssets))
   }
 
   @Path("/orderbook/{amountAsset}/{priceAsset}")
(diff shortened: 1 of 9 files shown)


src/main/scala/com/wavesplatform/matcher/api/MatcherApiRoute.scala
```diff
@@ -58,6 +59,14 @@ case class MatcherApiRoute(wallet: Wallet,
         .getOrElse[JsValue](JsString("")))
   }
 
+  @Path("/settings")
+  @ApiOperation(value = "Matcher Settings", notes = "Get matcher settings", httpMethod = "GET")
+  def getSettings: Route = (path("settings") & get) {
+    val priceAssets = matcherSettings.priceAssets
+    println(priceAssets)
+    complete(StatusCodes.OK -> Json.obj("settings" -> Json.obj("price-assets" -> priceAssets)))
```
phearnot: Please remove `println`, return an array of strings instead of an object and change path to `/priceAssets`

waves-testnet.conf
```diff
@@ -36,7 +36,7 @@ waves {
     enable = yes
 
     # Matcher's account address
-    # account = ""
+    account = "3N2KUTKtMgJdKXGeE9rtFM2o1GLmGNNqSdn"
```
phearnot: This should not be here

waves-testnet.conf
```diff
@@ -76,7 +65,7 @@ waves {
     port = 6869
 
     # Hash of API key string
-    api-key-hash = "H6nsiifwYKYEx6YzYD7woP1XCn72RVvx6tC1zjjLXqsu"
+    # api-key-hash = "H6nsiifwYKYEx6YzYD7woP1XCn72RVvx6tC1zjjLXqsu"
```
phearnot: Uncomment this, please

BlockchainJames:
Why did we remove predefined-pairs functionality? Is there an equivalent way of obtaining this functionality now? for example, if I wish to reverse a pair order from that that is auto-chosen

PR #1309, opened 2018-07-13
NODE-900 Height of transaction waiting decreased

monroid:


it/src/test/scala/com/wavesplatform/it/sync/smartcontract/AtomicSwapSmartContractSuite.scala
```diff
@@ -138,8 +136,8 @@ class AtomicSwapSmartContractSuite extends BaseTransactionSuite with CancelAfter
         .create(
           version = 2,
           assetId = None,
-          sender = PrivateKeyAccount.fromSeed(sender.seed(swapBC1)).explicitGet(),
-          recipient = PrivateKeyAccount.fromSeed(sender.seed(BobBC1)).explicitGet(),
+          sender = pkFromAddress(swapBC1),
+          recipient = pkFromAddress(BobBC1),
```
ismagin: recipient is just `Bob1` I guess, no need to get PK.
I'd also change function name to `pkOfAddress` , since there's no way to get Pk FROM address.

PR #1311, opened 2018-07-15
Node 944 disable smart asset

ismagin:


src/main/scala/scorex/api/http/assets/AssetsApiRoute.scala
```diff
@@ -283,10 +283,6 @@ case class AssetsApiRoute(settings: RestAPISettings, wallet: Wallet, utx: UtxPoo
           "decimals"       -> JsNumber(tx.decimals.toInt),
           "reissuable"     -> JsBoolean(description.reissuable),
           "quantity"       -> JsNumber(BigDecimal(description.totalVolume)),
-          "script"         -> Json.toJson(description.script.map(_.bytes().base58)),
```
monroid: You have to fit:
case class AssetInfo(assetId: String,
                     issueHeight: Int,
                     issueTimestamp: Long,
                     issuer: String,
                     name: String,
                     description: String,
                     decimals: Int,
                     reissuable: Boolean,
                     quantity: Long,
                     script: Option[String],
                     scriptText: Option[String],
                     complexity: Int,
                     extraFee: Long,
                     minSponsoredAssetFee: Option[Long])
to the new structure.

PR #1342, opened 2018-07-25
Node 962 Sript execution for exchange transaction

disfenoid:
[JIRA: https://wavesplatform.atlassian.net/browse/NODE-962](https://wavesplatform.atlassian.net/browse/NODE-962)


src/main/scala/com/wavesplatform/transaction/smart/Verifier.scala
```diff
@@ -31,22 +36,75 @@ object Verifier {
...
+    }
+  }
+
+  def verifyExchange(et: ExchangeTransaction,
+                     blockchain: Blockchain,
+                     matcherScriptOpt: Option[Script],
+                     height: Int): Either[ValidationError, Transaction] = {
+    val AssetPair(amountAsset, priceAsset) = et.sellOrder.assetPair
+
+    val amountAssetVerification =
```
ismagin: can we make this calculations lazy?

src/main/scala/com/wavesplatform/transaction/assets/exchange/Order.scala
```diff
@@ -185,19 +182,63 @@ case class Order(@ApiModelProperty(dataType = "java.lang.String") senderPublicKe
...
 
-  @ApiModelProperty(hidden = true)
-  def getSignedDescendants: Coeval[Seq[Signed]] = signedDescendants
-  @ApiModelProperty(hidden = true)
-  def getSignaturesValidMemoized: Task[Either[InvalidSignature, Order.this.type]] = signaturesValidMemoized
-  @ApiModelProperty(hidden = true)
-  def getSignaturesValid: Coeval[Either[InvalidSignature, Order.this.type]] = signaturesValid
+  def updateProofs(proofs: Proofs): Order
+
+  // For tests
```
phearnot: I don't really like the idea of leaking test-specific methods into model interfaces, but the only improvement I could think of was an extension class like this one:
```scala
  def copyHelper(o: Order, o1: OrderV1 => Order, o2: OrderV2 => Order) = o match {
    case ov1: OrderV1 => o1(ov1)
    case ov2: OrderV2 => o2(ov2)
  }

  implicit class OrderExt(order: Order) extends AnyVal {
    def updateExpiration(expiration: Long) = copyHelper(_.copy(expiration = expiration), _.copy(expiration = expiration))
    ...
  }

PR #1345, opened 2018-07-26
NODE-961-process-proofs-from-order

potan:
Add field.

### DIFF
--- a/src/main/scala/com/wavesplatform/transaction/smart/RealTransactionWrapper.scala
+++ b/src/main/scala/com/wavesplatform/transaction/smart/RealTransactionWrapper.scala
@@ -48,7 +48,7 @@ object RealTransactionWrapper {
       timestamp = o.timestamp,
       expiration = o.expiration,
       matcherFee = o.matcherFee,
-      signature = ByteVector(o.signature)
+      proofs = o.proofs.proofs.map(a => ByteVector(a.arr)).toIndexedSeq
     )
 
   implicit def aoaToRecipient(aoa: AddressOrAlias): Recipient = aoa match {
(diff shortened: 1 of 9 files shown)


lang/jvm/src/test/scala/com/wavesplatform/lang/Common.scala
```diff
@@ -67,10 +69,10 @@ object Common {
   def sampleUnionContext(instance: CaseObj) =
     EvaluationContext.build(Map.empty, Map("p" -> LazyVal(EitherT.pure(instance))), Seq.empty)
 
-  def emptyBlockchainEnvironment(h: Int = 1, tx: Coeval[Tx] = Coeval(???), nByte: Byte = 'T'): Environment = new Environment {
+  def emptyBlockchainEnvironment(h: Int = 1, in: Coeval[Tx :+: Ord :+: CNil] = Coeval(???), nByte: Byte = 'T'): Environment = new Environment {
```
petermz: `Tx :+: Ord :+: CNil` should probably be aliased

lang/shared/src/main/scala/com/wavesplatform/lang/v1/traits/domain/APair.scala
```diff
@@ -0,0 +1,5 @@
+package com.wavesplatform.lang.v1.traits.domain
+
+import scodec.bits.ByteVector
+
+case class APair(amountAsset: Option[ByteVector], priceAsset: Option[ByteVector])
```
petermz: Can this be `AssetPair` ?

src/main/scala/com/wavesplatform/transaction/smart/RealTransactionWrapper.scala
```diff
@@ -48,7 +49,8 @@ object RealTransactionWrapper {
       timestamp = o.timestamp,
       expiration = o.expiration,
       matcherFee = o.matcherFee,
-      signature = ByteVector(o.signature)
+      bodyBytes = ByteVector(o.bytes()), //TODO: Change to real bodyBytes
```
petermz: Really, why not `bodyBytes`? It's called `toSign` now, but should probably be renamed

lang/shared/src/main/scala/com/wavesplatform/lang/v1/traits/domain/APair.scala
potan: APair and AssetPair both is used in src/main/scala/com/wavesplatform/transaction/smart/RealTransactionWrapper.scala

PR #1409, opened 2018-08-15
Node 992 add active param

alexeykiselev:


it/src/main/scala/com/wavesplatform/it/api/AsyncMatcherHttpApi.scala
```diff
@@ -59,7 +80,7 @@ object AsyncMatcherHttpApi extends Assertions {
       matcherGet(s"/matcher/orderbook/$amountAsset/$priceAsset").as[OrderBookResponse]
     }
 
-    def parseAssetPair(assetPair: AssetPair) = {
+    def parseAssetPair(assetPair: AssetPair): (io.Serializable, io.Serializable) = {
```
vsuharnikov: io.Serializable? So weird :D

PR #1434, opened 2018-08-28
NODE-1014 Add order's run script

monroid:


src/test/scala/com/wavesplatform/state/diffs/smart/predef/package.scala
```diff
@@ -26,6 +27,17 @@ package object predef {
     } yield r
   }
 
+  def runScript2[T](script: String, ord: Transaction :+: Order :+: CNil, networkByte: Byte = networkByte): Either[String, T] = {
```
ismagin: this can be refactored better, no need to copy-paste

monroid: This is a new method for script running, in case of refactoring of the first one, I should change all invokes of runScript to construction -> runScript(s, Coproduct, 'T') or smth. I see it as overcomplicated refactoring.

PR #1439, opened 2018-08-29
NODE-1103 many expressions fix

disfenoid:
[JIRA: https://wavesplatform.atlassian.net/browse/NODE-1103](https://wavesplatform.atlassian.net/browse/NODE-1103)


ismagin:
will be merged through .14

PR #1445, opened 2018-08-31
NODE-956: Reliable order cancellation

phearnot:


src/main/scala/com/wavesplatform/matcher/api/MatcherApiRoute.scala
```diff
@@ -124,12 +127,40 @@ case class MatcherApiRoute(wallet: Wallet,
...
+      case status if status.isFinal => StatusCodes.BadRequest -> Json.obj("message" -> s"Order is already ${status.name}")
+      case _ =>
+        DBUtils.order(db, orderId) match {
+          case None =>
+            log.warn(s"Order $orderId was not found in history")
+            StatusCodes.NotFound
+          case Some(order) if senderPublicKey.exists(_ != order.senderPublicKey) =>
+            StatusCodes.BadRequest -> Json.obj("message" -> "Public Key mismatch")
+          case Some(order) =>
+            orderBook(order.assetPair) match {
```
vsuharnikov: What if the matcher is starting (not yet started), and we receive the cancel request?

PR #1452, opened 2018-09-03
NODE-1049-fix-message

potan:
Place actual height to error message about duplicate transactions.

### DIFF
--- a/src/main/scala/com/wavesplatform/state/diffs/CommonValidation.scala
+++ b/src/main/scala/com/wavesplatform/state/diffs/CommonValidation.scala
@@ -81,10 +81,8 @@ object CommonValidation {
                                              tx: T): Either[ValidationError, T] = tx match {
     case _: PaymentTransaction => Right(tx)
     case _ =>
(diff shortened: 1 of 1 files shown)


src/main/scala/com/wavesplatform/state/diffs/CommonValidation.scala
```diff
@@ -80,7 +80,9 @@ object CommonValidation {
                                              height: Int,
                                              tx: T): Either[ValidationError, T] = tx match {
     case _: PaymentTransaction => Right(tx)
-    case _                     => if (blockchain.containsTransaction(tx.id())) Left(AlreadyInTheState(tx.id(), 0)) else Right(tx)
+    case _ =>
+      val id = tx.id()
+      Either.cond(!blockchain.containsTransaction(id), tx, AlreadyInTheState(id, blockchain.transactionInfo(id).get._1))
```
ismagin: Why not `transactionHeight(id: ByteStr): Option[Int]`? with this, no transaction will ever be loaded

PR #1453, opened 2018-09-03
NODE-1109 Order history optimization

vsuharnikov:
TODO:
* [x] address+pair index for orders;
* [x] Track last active order index in common list of orders and limit steps by it;
* Update address+pair index:
  * [x] on place;
* Update last active order index in common list:
  * [x] on place;
  * [x] on fill;
  * [x] on cancel;
  * [x] on delete?
* [x] Unit tests.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -92,6 +92,7 @@ inTask(assembly)(
 
 inConfig(Compile)(
   Seq(
+    // mainClass := Some("com.wavesplatform.matcher.MigrationTool"),
     mainClass := Some("com.wavesplatform.Application"),
     publishArtifact in packageDoc := false,
     publishArtifact in packageSrc := false,
--- a/build.sbt
+++ b/build.sbt
@@ -92,7 +92,6 @@ inTask(assembly)(
 
 inConfig(Compile)(
   Seq(
-    // mainClass := Some("com.wavesplatform.matcher.MigrationTool"),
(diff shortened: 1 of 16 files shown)


src/main/scala/com/wavesplatform/matcher/model/OrderHistory.scala
```diff
@@ -49,22 +50,29 @@ class OrderHistory(db: DB, settings: MatcherSettings) {
...
+        val changedKeys2 = if (origInfo.isEmpty) {
+          saveOrder(rw, order)
+          addOrderIndexes(rw, change, changedKeys1)
+        } else changedKeys1
+
+        (updateOldestActiveNr(rw, change, changedKeys2), origChanges.updated(orderId, change))
+    }
+
+    val opDiff = diff(event, changes)
+    saveOpenVolume(rw, opDiff)
```
vsuharnikov: Just realised, I can make saveOpenVolume work with (Address, OpenPortfolio), not the Map, and then able to simplify this method. But this changes will be later in a bug-fix task :)

build.sbt
```diff
@@ -92,6 +92,7 @@ inTask(assembly)(
 
 inConfig(Compile)(
   Seq(
+    // mainClass := Some("com.wavesplatform.matcher.MigrationTool"),
```
vsuharnikov: Will remove

PR #1455, opened 2018-09-03
Node 1104 sigvalid 2

ismagin:


ismagin:
Won't merge, decided top decline TXV1 from scripted accounts

PR #1459, opened 2018-09-05
NODE-1123 Added test with migration tool execution

monroid:


it/src/main/scala/com/wavesplatform/it/Docker.scala
```diff
@@ -527,6 +527,68 @@ class Docker(suiteConfig: Config = empty, tag: String = "", enableProfiling: Boo
...
+    log.debug("Change script for migration tool launch")
+    val scriptCmd: Array[String] =
+      Array(
+        "sh",
+        "-c",
+        s"""rm /opt/waves/start-waves.sh && echo '#!/bin/bash' >> /opt/waves/start-waves.sh &&
+             |echo 'java ${renderProperties(asProperties(genesisOverride))} -cp /opt/waves/waves.jar com.wavesplatform.matcher.MigrationTool /opt/waves/template.conf cb > /opt/waves/migration-tool.log' >> /opt/waves/start-waves.sh &&
+             |echo 'while true; do echo 1; sleep 10; done' >> /opt/waves/start-waves.sh &&
+             |chmod +x /opt/waves/start-waves.sh
+           """.stripMargin
```
vsuharnikov: We could replace the script at the end. With this change:
1. We doesn't need to wait 60 seconds above
2. We should check the status of the container in a loop until it closed

PR #1466, opened 2018-09-06
NODE-979: Ask for password if not specified

disfenoid:
[JIRA: https://wavesplatform.atlassian.net/browse/NODE-979](https://wavesplatform.atlassian.net/browse/NODE-979)

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -118,7 +118,7 @@ waves {
     file = ${waves.directory}"/wallet/wallet.dat"
 
     # Password to protect wallet file
-    password = "some string as password"
+//    password = "some string as password"
 
     # By default, the node will attempt to generate a new seed. To use a specific seed, uncomment the following line and
     # specify your base58-encoded seed.
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -118,7 +118,7 @@ waves {
     file = ${waves.directory}"/wallet/wallet.dat"
 
     # Password to protect wallet file
-//    password = "some string as password"
+    # password = "some string as password"
 
     # By default, the node will attempt to generate a new seed. To use a specific seed, uncomment the following line and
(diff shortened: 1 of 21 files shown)


src/main/scala/com/wavesplatform/settings/WalletSettings.scala
```diff
@@ -2,6 +2,59 @@ package com.wavesplatform.settings
...
+    val walletFile =
+      config
+        .as[Option[File]](s"$configPath.file")
+    val seed =
+      config
+        .as[Option[ByteStr]](s"$configPath.seed")
+
+    WalletSettings(walletFile, password, seed)
+  }
+  def readPasswordFromCommandLine(): String = {
```
potan: Function readPasswordFromCommandLine read a password from console but not from command line. Should we rename it?

src/main/resources/application.conf
```diff
@@ -118,7 +118,7 @@ waves {
     file = ${waves.directory}"/wallet/wallet.dat"
 
     # Password to protect wallet file
-    password = "some string as password"
+//    password = "some string as password"
```
ismagin: does this comment work? shouldn't it be # ?

src/main/scala/com/wavesplatform/settings/WalletSettings.scala
```diff
@@ -2,6 +2,59 @@ package com.wavesplatform.settings
...
+        |**********************************
+      """.stripMargin
+
+    try {
+      message.lines
+        .foreach(println)
+
+      System
+        .console()
+        .readPassword("Password > ")
```
ismagin: is there a way to hide password chars during input?

src/test/resources/password.conf
```diff
@@ -0,0 +1 @@
+some string as password
```
ismagin: where is this used?

PR #1471, opened 2018-09-10
NODE-1127: Deny rolling back more than 1990 blocks

disfenoid:
[JIRA: https://wavesplatform.atlassian.net/browse/NODE-1127](https://wavesplatform.atlassian.net/browse/NODE-1127)


src/main/scala/com/wavesplatform/http/DebugApiRoute.scala
```diff
@@ -192,11 +194,15 @@ case class DebugApiRoute(ws: WavesSettings,
     ))
   def rollback: Route = (path("rollback") & post & withAuth) {
     json[RollbackParams] { params =>
-      ng.blockAt(params.rollbackTo) match {
-        case Some(block) =>
-          rollbackToBlock(block.uniqueId, params.returnTransactionsToUtx)
-        case None =>
-          (StatusCodes.BadRequest, "Block at height not found")
+      if (ng.height - params.rollbackTo > 1990)
```
ismagin: can we use the `2000` number from state here? 
in case that will be changed later

src/main/scala/com/wavesplatform/http/DebugApiRoute.scala
```diff
@@ -277,12 +283,27 @@ case class DebugApiRoute(ws: WavesSettings,
...
+        signature <- ByteStr
+          .decodeBase58(signature)
+          .toEither
+          .leftMap(_ => InvalidRequestSignature)
+        height <- blockchain
+          .heightOf(signature)
+          .toRight(GenericError("Block with such signature not found"))
+        _ <- Either
+          .cond(
+            ng.height - height <= 1990,
```
ismagin: the same as previous

src/main/scala/com/wavesplatform/http/DebugApiRoute.scala
```diff
@@ -192,11 +195,15 @@ case class DebugApiRoute(ws: WavesSettings,
     ))
   def rollback: Route = (path("rollback") & post & withAuth) {
     json[RollbackParams] { params =>
-      ng.blockAt(params.rollbackTo) match {
-        case Some(block) =>
-          rollbackToBlock(block.uniqueId, params.returnTransactionsToUtx)
-        case None =>
-          (StatusCodes.BadRequest, "Block at height not found")
+      if (ng.height - params.rollbackTo > LevelDBWriter.MAX_DEPTH - 10)
+        (StatusCodes.BadRequest, "Rollback of more than 1990 blocks is forbidden")
```
petermz: SHould be `${MAX_DEPTH - 10}` really

src/main/scala/com/wavesplatform/http/DebugApiRoute.scala
```diff
@@ -277,12 +284,27 @@ case class DebugApiRoute(ws: WavesSettings,
...
+          .toEither
+          .leftMap(_ => InvalidRequestSignature)
+        height <- blockchain
+          .heightOf(signature)
+          .toRight(GenericError("Block with such signature not found"))
+        _ <- Either
+          .cond(
+            ng.height - height < (LevelDBWriter.MAX_DEPTH - 10),
+            (),
+            GenericError("Rollback of more than 1990 blocks is forbidden")
```
petermz: Same here

PR #1475, opened 2018-09-11
NODE-1144: PaymentApiRoute, WavesApiRoute removed

disfenoid:
[JIRA: https://wavesplatform.atlassian.net/browse/NODE-1144](https://wavesplatform.atlassian.net/browse/NODE-1144)


ismagin:
exchanges might use it, let's analyze the impact of this change first

ismagin:
decided to hide instead of deletion

Issue #1472, opened 2018-09-10
java8-runtime-headless as dependency instead of openjdk-8-jre-headless in .deb package


wallisonalves:
Why the .deb package asks for a java8-runtime-headless dependency instead of openjdk-8-jre-headless where you do not need to use ppa to use?
Avoid using ppa packages please change this
use openjdk-8-jre-headless as dependency instead of java8-runtime-headless that only exists in a ppa

wallisonalves:
Packages providing java8-runtime-headless

default-jre-headless
    Standard Java or Java compatible Runtime (headless)
openjdk-8-jre-headless
    OpenJDK Java runtime, using Hotspot JIT (headless) 
https://packages.debian.org/stretch/java8-runtime-headless

wallisonalves:
why not use the package that is in the official debian repository and its derivatives? choosing a ppa is a serious mistake that should never happen!

phearnot:
Although the safety of PPAs is debatable, DEB package we build does not explicitly depend on non-official packages. [OpenJDK 8](https://packages.debian.org/stretch/openjdk-8-jre-headless), which is a part of an official Debian distribution, provides the virtual `java8-runtime-headless` package.
Also unlike `java8-runtime-headless`, `default-jre-headless` does not specify an exact Java version, so one may end up with an unexpected runtime installed (e.g. on Ubuntu 18.04 LTS, `default-jre-headless` depends on OpenJDK 11)

wallisonalves:
So just switch to: openjdk-8-jre-headless

PR #1485, opened 2018-09-16
Use the official debian repository

wallisonalves:
Use the official debian repository because PPA is not safe to use in critical software

Packages providing java8-runtime-headless

default-jre-headless
Standard Java or Java compatible Runtime (headless)
openjdk-8-jre-headless
OpenJDK Java runtime, using Hotspot JIT (headless)
https://packages.debian.org/stretch/java8-runtime-headless

https://github.com/wavesplatform/Waves/issues/1472


phearnot:
See https://github.com/wavesplatform/Waves/issues/1472#issuecomment-423078238.

PR #1515, opened 2018-09-27
[NODE-1186] add scripts in IssueTransactionV2 JSON and SignedIssueV2Request

d-kiselev:


src/main/scala/com/wavesplatform/transaction/assets/IssueTransactionV2.scala
```diff
@@ -36,6 +39,18 @@ case class IssueTransactionV2 private (version: Byte,
     ))
   override val bytes: Coeval[Array[Byte]] = Coeval.evalOnce(Bytes.concat(Array(0: Byte), bodyBytes(), proofs.bytes()))
 
+  override val json = Coeval.evalOnce(
```
ismagin: need to reuse this

PR #1519, opened 2018-09-28
NODE-1159: Add OrderType binding and test

disfenoid:
[JIRA: https://wavesplatform.atlassian.net/browse/NODE-1105](https://wavesplatform.atlassian.net/browse/NODE-1105)


src/main/scala/com/wavesplatform/transaction/smart/BlockchainContext.scala
```diff
@@ -16,6 +17,18 @@ object BlockchainContext {
   private val baseContext = Monoid.combine(PureContext.ctx, CryptoContext.build(Global)).evaluationContext
 
   type In = Transaction :+: Order :+: CNil
-  def build(nByte: Byte, in: Coeval[In], h: Coeval[Int], blockchain: Blockchain): EvaluationContext =
-    Monoid.combine(baseContext, WavesContext.build(new WavesEnvironment(nByte, in, h, blockchain)).evaluationContext)
+  def build(nByte: Byte, in: Coeval[In], h: Coeval[Int], blockchain: Blockchain): EvaluationContext = {
+    val typeVarsEnabled = Option(blockchain)
```
ismagin: this `Option(.)` is for tests' `null` only. Let's fix tests and pass empty blockchain there.
when we'll test scripts for new lib version we'll have to do it anyway

PR #1520, opened 2018-10-01
NODE-1178 Cleanup for inactive orders

vsuharnikov:


vsuharnikov:
An integration test was broken before in Friday

PR #1541, opened 2018-10-04
NODE-1193 New indexes for order history

vsuharnikov:


src/main/scala/com/wavesplatform/matcher/api/DBUtils.scala
```diff
@@ -1,52 +1,96 @@
...
+      maxOrders,
+      activeIndex = indexes.active.iterator(ro, address).map { case (_, id) => id },
+      nonActiveIndex = if (activeOnly) Iterator.empty else indexes.finalized.common.iterator(ro, address)
+    )
+  }
+
+  private def mergeOrders(ro: ReadOnlyDB,
+                          maxOrders: Int,
+                          activeIndex: Iterator[Order.Id],
+                          nonActiveIndex: Iterator[Order.Id]): IndexedSeq[(Order, OrderInfo)] = {
```
phearnot: inactive

src/main/scala/com/wavesplatform/matcher/api/MatcherApiRoute.scala
```diff
@@ -234,14 +235,8 @@ case class MatcherApiRoute(wallet: Wallet,
...
-    withAssetPair(p) { pair =>
-      json[CancelOrderRequest] { req =>
-        if (req.isSignatureValid()) {
-          log.trace(s"Deleting ${req.orderId.fold(s"all orders for ${req.sender}")(_.base58)}")
-          (orderHistory ? DeleteOrderFromHistory(pair, req.sender, req.orderId, NTP.correctedTime()))
-            .mapTo[MatcherResponse]
-        } else InvalidSignature
-      }
+    json[CancelOrderRequest] { req =>
+      req.orderId.fold[MatcherResponse](NotImplemented("Batch order deletion is not supported yet"))(OrderDeleted)
```
phearnot: just remove batch deletion support

vsuharnikov: I can't, CancelOrderRequest is used in other methods :) A current implementation is ok, I think...

PR #1543, opened 2018-10-04
NODE-1173 Create matcher base class with predefined addresses

monroid:


it/src/main/scala/com/wavesplatform/it/MatcherNode.scala
```diff
@@ -0,0 +1,60 @@
...
+  protected lazy val bobAddress: String     = bobNode.createAddress()
+
+  protected lazy val matcherAcc = PrivateKeyAccount.fromSeed(matcherNode.seed(matcherAddress)).right.get
+  protected lazy val aliceAcc   = PrivateKeyAccount.fromSeed(aliceNode.seed(aliceAddress)).right.get
+  protected lazy val bobAcc     = PrivateKeyAccount.fromSeed(bobNode.seed(bobAddress)).right.get
+
+  private val addresses = Seq(matcherAddress, aliceAddress, bobAddress)
+
+  //really before all tests, because FreeSpec issue with "-" and "in"
+  initialBalances()
```
vsuharnikov: Actually it's a bad idea, because we should think about traits linearisation in tests. E.g.:
```scala
trait Foo { println("foo") }
trait Bar { println("bar") }
trait Baz {}

class X extends Baz with Foo with Bar {}
class Y extends Baz with Bar with Foo {}

new X
// foo
// bar

new Y
// bar
// foo
```

So, if someone write a similar class and mix it with `MatcherNode`, it will face bizarre problems.

it/src/main/scala/com/wavesplatform/it/api/AsyncMatcherHttpApi.scala
```diff
@@ -114,10 +111,9 @@ object AsyncMatcherHttpApi extends Assertions {
       matcherPost(s"/matcher/orderbook/$amountAsset/$priceAsset/cancel", signedRequest).as[MatcherStatusResponse]
     }
 
-    def cancelAllOrders(sender: Node, timestamp: Option[Long]): Future[MatcherStatusResponse] = {
+    def cancelAllOrders(sender: PrivateKeyAccount, timestamp: Option[Long]): Future[MatcherStatusResponse] = {
```
vsuharnikov: I suggest to use `Address` if it possible, because `PrivateKeyAccount` is a subclass of `PublicKeyAccount`. Also we have an implicit conversion from `PublicKeyAccount` to `Address` (see `PublicKeyAccount.toAddress`). So it's a more general.

it/src/main/scala/com/wavesplatform/it/matcher/MatcherSuiteBase.scala
```diff
@@ -0,0 +1,38 @@
...
+  protected def nodeConfigs: Seq[Config] =
+    NodeConfigs.newBuilder
+      .withDefault(4)
+      .buildNonConflicting()
+
+  override protected def nodes: Seq[Node] = dockerNodes()
+
+  protected override def beforeAll(): Unit = {
+    super.beforeAll()
+  }
```
vsuharnikov: Pointless :)

PR #1555, opened 2018-10-08
NODE-1147-online-serialization

potan:


src/main/scala/com/wavesplatform/api/http/UtilsApiRoute.scala
```diff
@@ -211,6 +212,26 @@ case class UtilsApiRoute(timeService: Time, settings: RestAPISettings) extends A
...
+        required = true,
+        paramType = "body",
+        dataType = "string",
+        value = "Transaction data including <a href='transaction-types.html'>type</a> and signature/proofs"
+      )
+    ))
+  def transactionSerialize: Route = (pathPrefix("transactionSerialize") & post) {
+    handleExceptions(jsonExceptionHandler) {
+      json[JsObject] { jsv =>
+        TransactionFactory.fromSignedRequest(jsv).fold(ApiError.fromValidationError, tx => Json.obj("bytes" -> tx.bytes()))
```
petermz: I think this should be unsigned request? The point of this method is to return the array to be signed

PR #1567, opened 2018-10-11
NODE-1223 Generating Balance has negative value

petermz:
JIRA: https://wavesplatform.atlassian.net/browse/NODE-1223
* Fixed generating balance calculation in LevelDBWriter
* Added an "import up to this height" parameter to Importer


petermz:
Do not merge. This fix can fork

PR #1568, opened 2018-10-11
NODE-1220-SetAssetScriptTransaction

potan:
Serialization
Tests refactoring


src/main/scala/com/wavesplatform/state/diffs/CommonValidation.scala
```diff
@@ -128,6 +129,12 @@ object CommonValidation {
       case _: SetScriptTransaction     => activationBarrier(BlockchainFeatures.SmartAccounts)
       case _: TransferTransactionV2    => activationBarrier(BlockchainFeatures.SmartAccounts)
       case it: IssueTransactionV2      => activationBarrier(if (it.script.isEmpty) BlockchainFeatures.SmartAccounts else BlockchainFeatures.SmartAssets)
+      case it: SetAssetScriptTransaction =>
+        if (it.script.isEmpty) {
+          Left(GenericError("Empty sctipt is disabled."))
```
petermz: scRipt

PR #1579, opened 2018-10-16
NODE-1242-fix-transactionHeight

potan:
Fix transactionHeight for transactions not placed in liquid block.
Replace Int to Long type in RIDE functions interfases.


ismagin:
decided to remove the function from hotfix release instead.

PR #1610, opened 2018-10-26
Node 1261 true false terms remove

dvshur:


lang/jvm/src/test/scala/com/wavesplatform/lang/SerdeTest.scala
```diff
@@ -20,33 +20,33 @@ class SerdeTest extends FreeSpec with PropertyChecks with Matchers with ScriptGe
...
-          args = List(TRUE)
+          args = List(CONST_BOOLEAN(true))
         )
       )
 
       "user" in roundTripTest(
         FUNCTION_CALL(
           function = FunctionHeader.User("foo"),
-          args = List(TRUE)
+          args = List(CONST_BOOLEAN(true))
```
ismagin: This repeats one too many times.
Maybe 
```
object CONST_BOOLEAN {
   val TRUE = ...
   val FALSE = ...
}
```

PR #1612, opened 2018-10-29
DEX-10 Batch cancel

phearnot:


src/main/scala/com/wavesplatform/matcher/model/OrderHistory.scala
```diff
@@ -25,13 +25,32 @@ class OrderHistory(db: DB, settings: MatcherSettings) extends ScorexLogging {
...
+        rw.put(k, Some(newTimestamp))
+        Right(newTimestamp)
+      case Some(prevTimestamp) =>
+        val earliestTimestamp = prevTimestamp - settings.orderTimestampDrift
+        if (newTimestamp < earliestTimestamp) {
+          Left(s"Timestamp must be >= $earliestTimestamp")
+        } else if (newTimestamp <= prevTimestamp) {
+          Right(prevTimestamp)
+        } else {
+          db.readWrite(_.put(k, Some(newTimestamp)))
```
vsuharnikov: `rw.put(k, Some(newTimestamp))` ?

PR #1620, opened 2018-10-30
NODE-1221-disable-sponsorship-on-smartasset

potan:


src/test/scala/com/wavesplatform/state/diffs/CommonValidationTest.scala
```diff
@@ -188,25 +188,25 @@ class CommonValidationTest extends PropSpec with PropertyChecks with Matchers wi
...
-        withStateAndHistory(settings) { blockchain =>
-          val (preconditionDiff, preconditionFees, _) =
-            BlockDiffer.fromBlock(settings, blockchain, None, genesisBlock, MiningConstraint.Unlimited).explicitGet()
-          blockchain.append(preconditionDiff, preconditionFees, genesisBlock)
-
-          CommonValidation.checkFee(blockchain, settings, 1, transferTx) shouldBe 'right
-        }
-    }
-  }
+//  property("checkFee sponsored + smart tokens + smart accounts sunny") {
```
ismagin: maybe remove

PR #1644, opened 2018-11-06
NODE-1276: treat null in json as empty proof

disfenoid:
[JIRA: https://wavesplatform.atlassian.net/browse/NODE-1276](https://wavesplatform.atlassian.net/browse/NODE-1276)


it/src/test/scala/com/wavesplatform/it/sync/transactions/SignAndBroadcastApiSuite.scala
```diff
@@ -106,21 +106,29 @@ class SignAndBroadcastApiSuite extends BaseTransactionSuite {
         version = v
       )
 
-      signBroadcastAndCalcFee(
+      println(s"Issue: $issueId")
```
ismagin: please cleanup printlns

PR #1647, opened 2018-11-06
DEX-155 Smart assets trading

vsuharnikov:
* OrderValidator checks an order against the fake one

### DIFF
--- a/it/src/test/scala/com/wavesplatform/it/sync/matcher/smartcontracts/OrdersFromScriptedAssetTestSuite.scala
+++ b/it/src/test/scala/com/wavesplatform/it/sync/matcher/smartcontracts/OrdersFromScriptedAssetTestSuite.scala
@@ -1,0 +1,222 @@
+package com.wavesplatform.it.sync.matcher.smartcontracts
+
+import com.typesafe.config.{Config, ConfigFactory}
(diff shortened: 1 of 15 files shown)


it/src/test/scala/com/wavesplatform/it/sync/matcher/smartcontracts/OrdersFromScriptedAssetTestSuite.scala
```diff
@@ -0,0 +1,234 @@
...
+      timestamp = System.currentTimeMillis(),
+      expiration = System.currentTimeMillis() + 60 * 60 * 30,
+      matcherFee = smartTradeFee,
+      version = 2
+    )
+    matcherNode.expectIncorrectOrderPlacement(order, 400, "OrderRejected")
+  }
+
+  "can execute against unscripted, if the script returns TRUE" in {
+    markup("place counter")
```
msmolyakov: Better to use `info()` or `Given()/When()/Then()/And()` (extend the class with `GivenWhenThen`).
The `markup()` is more intended for HTML reports.

vsuharnikov: Copied this from your RestOrderLimitTestSuite :D

PR #1649, opened 2018-11-06
NODE-1268 Check full asset info

monroid:


src/main/scala/com/wavesplatform/api/http/assets/AssetsApiRoute.scala
```diff
@@ -361,8 +361,8 @@ case class AssetsApiRoute(settings: RestAPISettings, wallet: Wallet, utx: UtxPoo
             case 0           => JsNull
             case sponsorship => JsNumber(sponsorship)
           })
-        ) ++ (script.toSeq.flatMap { script =>
-          Seq(
+        ) ++ (script.toSeq.map { script =>
+          "scriptDetails" -> Json.obj(
```
ismagin: these are breaking changes, sure we can introduce them?

PR #1664, opened 2018-11-12
NODE-1254: Unit removed from TYPE ADT

disfenoid:
[JIRA: https://wavesplatform.atlassian.net/browse/NODE-1254](https://wavesplatform.atlassian.net/browse/NODE-1254)

### DIFF
--- a/lang/shared/src/main/scala/com/wavesplatform/lang/v1/evaluator/ctx/impl/PureContext.scala
+++ b/lang/shared/src/main/scala/com/wavesplatform/lang/v1/evaluator/ctx/impl/PureContext.scala
@@ -328,8 +328,6 @@ object PureContext {
     uNot
   )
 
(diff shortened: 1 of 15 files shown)


lang/shared/src/main/scala/com/wavesplatform/lang/v1/evaluator/ctx/impl/PureContext.scala
```diff
@@ -328,8 +328,6 @@ object PureContext {
     uNot
   )
 
-  lazy val unit: EVALUATED = CaseObj(CASETYPEREF("Unit", List.empty), Map.empty)
-
   private lazy val vars: Map[String, ((FINAL, String), LazyVal)] = Map(("unit", ((UNIT, "Single instance value"), LazyVal(EitherT.pure(unit)))))
```
ismagin: reference impl.UNIT maybe ?

PR #1685, opened 2018-11-16
NODE-1271: Increase interval between mining attempts if quorum not available

disfenoid:
[JIRA: https://wavesplatform.atlassian.net/browse/NODE-1271](https://wavesplatform.atlassian.net/browse/NODE-1271)


src/main/scala/com/wavesplatform/mining/Miner.scala
```diff
@@ -320,6 +323,12 @@ class MinerImpl(allChannels: ChannelGroup,
                   if (ngEnabled && !totalConstraint.isEmpty) startMicroBlockMining(account, block, estimators, totalConstraint)
                 case Right(None) => log.warn("Newly created block has already been appended, should not happen")
               }
+
+          case Left(err) if err contains "Quorum not available" =>
+            log.debug(s"$err. Next attempt in 5 seconds")
+            generateBlockTask(account)
+              .delayExecution(5.seconds)
```
vsuharnikov: The problem here is that `generateBlockTask` takes `height` and `lastBlock` and only then it creates a task. So, we can receive a block in this 5 seconds, but forge an own, based on previous

PR #1696, opened 2018-11-23
Node 1251 Smart assets fee

monroid:
- checking of fee for smart assets
- deleted duplication of exchange transaction usage
- small code improvements for easy reading


vsuharnikov:
We're replacing:
```
issue transaction
wait it on all nodes
```
by:
```
issue transaction
wait it on sender's node
```

Sometimes this can lead to situations:
1. We send a transfer transaction from `Account A` to `Account B` by `Node K`. Actually, it can be an issue/lease/... transaction.
2. We wait until it is mined on `Node K`, but not on other nodes
3. We send a transfer transaction from `Account B` to `Account C` by `Node M`
4. `Node M` says, that the transaction is an invalid

So, this two ways aren't the same. From my opinion, i'd better to do:
```
// Send to one of nodes and wait on all 
nodes.transfer(secondAddress, firstAddress, 1000, minFee + smartFee, Some(assetB), waitForTx = true)
```

What do you think?

monroid:
> We're replacing:
> 
> ```
> issue transaction
> wait it on all nodes
> ```
> by:
> 
> ```
> issue transaction
> wait it on sender's node
> ```
> Sometimes this can lead to situations:
> 
> 1. We send a transfer transaction from `Account A` to `Account B` by `Node K`. Actually, it can be an issue/lease/... transaction.
> 2. We wait until it is mined on `Node K`, but not on other nodes
> 3. We send a transfer transaction from `Account B` to `Account C` by `Node M`
> 4. `Node M` says, that the transaction is an invalid
> 
> So, this two ways aren't the same. From my opinion, i'd better to do:
> 
> ```
> // Send to one of nodes and wait on all 
> nodes.transfer(secondAddress, firstAddress, 1000, minFee + smartFee, Some(assetB), waitForTx = true)
> ```
> What do you think?

yep, I agree that sometimes we have to wait a tx on set of nodes, but those tests use only one node, so I can a little bit cheat here and wait for tx only there. For you case I will do nodes.waitForTx(...) instead of node.tx(...,waitForTx = true)

PR #1726, opened 2018-12-06
NODE-1319-stop-checking-txs-order

potan:
No check order of transaction while block is validated.


ismagin:
I guess you can also drop `requireSortedTransactionsAfter` and everything related?

potan:
> I guess you can also drop `requireSortedTransactionsAfter` and everything related?

Done

PR #1732, opened 2018-12-10
NODE-1324 test for unsupported transactions in smart asset script

msmolyakov:


it/src/test/scala/com/wavesplatform/it/sync/smartcontract/smartasset/NoOrderProofsSuite.scala
```diff
@@ -27,18 +29,17 @@ class NoOrderProofsSuite extends BaseTransactionSuite {
...
+              |  case _ => false
+              |}""".stripMargin,
+            isAssetScript = true
           ).explicitGet()._1.bytes.value.base64)
       )
-    catch {
-      case ex: java.lang.Exception => assert(ex.getMessage.contains("Compilation failed: Matching not exhaustive"))
-      case _: Throwable            => throw new Exception("ScriptCompiler works incorrect for orders with smart assets")
+    } match {
+      case Success(_) => fail("ScriptCompiler didn't throw expected error")
```
monroid: I don't think that we should here use "some sort of sugar" for "better reading". My point was to catch exactly java.lang.Exception exception, if we will customise our exception this test catches that. But in your interpretation we won't see changes.

it/src/test/scala/com/wavesplatform/it/sync/smartcontract/smartasset/AssetUnsupportedTransactionsSuite.scala
```diff
@@ -0,0 +1,66 @@
...
+      "LeaseTransaction",
+      "LeaseCancelTransaction",
+      "CreateAliasTransaction",
+      "SetScriptTransaction",
+      "DataTransaction",
+      "IssueTransaction"
+    )) { tx =>
+    test(s"Smart Asset script should not support $tx") {
+      Try {
+        sender.setAssetScript(
```
monroid: For such checks you can use issue tx, instead of setassetscript. It drops before section and one unnecessary tx :)

it/src/test/scala/com/wavesplatform/it/sync/smartcontract/smartasset/NoOrderProofsSuite.scala
msmolyakov: Previous check was not exact catching too. It catched any exception inherited of java.lang.Exception.

it/src/test/scala/com/wavesplatform/it/sync/smartcontract/smartasset/AssetUnsupportedTransactionsSuite.scala
msmolyakov: I agree. Fixed it, thx

PR #1743, opened 2018-12-13
DEX-30 Redundancy

vsuharnikov:
High-level:
* New architecture of requests processing:
  * Mutable requests are validated and sent as events to the queue;
  * DEX processes events from the queue only;
  * Each instance of DEX processes the event and generates ExchangeTransaction (Event is translated to ExchangeTransactions in idenpotent manner);
  * DEX watches procession results and sends reponses to clients as soon as results are available;
* DEX can work with local events queue or distributed (Kafka);
* ExchangeTransactions are still validated in local UTX Pool before sending;

Matcher:
* All matcher responses now in MatcherResponse.scala;
* Matcher commands order books to create snapshots from time to time;

LocalMatcherQueue:
* As alternative for Kafka if it isn't needed;
* Limit max elements per poll;
* Cleaning before consume;

MatcherActor:
* Child actors (OrderBook) will be stopped if an error happens inside of them. They are restarted before;

OrderBookActor:
* Doesn't use putIfNew;
* Has no Events;
* Snapshot contains the latest command number;
* Does snapshots on SaveSnapshot command;

OrderHistory:
* Events to OrderHistory are applied only once;

ExchangeTransaction:
* Timestamp is taken from Kafka's submitted order (Kafka queue);
* Timestamp is taken from local corrected time (Local queue);
* Thread pool for exchange transaction validation and sending;

ExchangeTransactionCreator:
* Doesn't depend on Time. A timestamp of ExchangeTransaction is the newest order timestamp;

MatcherSettings:
* New section "waves.matcher.events-queue". Contains settings for kafka, local queue and selected type;

API:
* API for DEX queue diagnostics;

API internals:
* ApiRoute._json alternative to json, when a better composition is needed;
* MatcherApiRoute - shutdownBarrier, unavailableOrderBookBarrier;
* Fixed an akka-http warning about Content-Type;

Integration tests:
* Ability to run with Kafka;
* OrderInfo, ExchangeTransaction - added version and proofs;
* InMemory snapshots and journal stores for unit tests. They are cleaned before each test;
* Integration test for DEX recovery;
* Multiple DEX instances test;
* expectSignedBroadcastRejected;
* OrdersFromScriptedAssetTestSuite, OrderTypeTestSuite, ProofAndAssetPairTestSuite - an exchange transaction isn't verified during matching, so we check it later;
* Logging a request's method name in integration tests;
* Annotate requests with id for better logs;

Logging:
* Reduced amount of Kafka client logs;

Benchmark fixes:
* Can run on Java8;
* Fixed smart contract code in StateSyntheticBenchmark;
* Disabled logging;

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -65,6 +65,8 @@ object Dependencies {
   lazy val matcher = Seq(
     akkaModule("persistence"),
     akkaModule("persistence-tck") % "test",
+    "com.github.dnvriend"         %% "akka-persistence-inmemory" % "2.5.15.1" % "test",
+    "com.typesafe.akka"           %% "akka-stream-kafka" % "1.0-M1",
     "org.ethereum"                % "leveldbjni-all" % "1.18.3"
   )
 
--- a/benchmark/src/main/resources/logback.xml
+++ b/benchmark/src/main/resources/logback.xml
@@ -15,7 +15,7 @@
 
     <logger name="org.aspectj" level="INFO"/>
 
-    <root level="TRACE">
+    <root level="OFF">
         <appender-ref ref="STDOUT"/>
     </root>
 </configuration>
--- a/benchmark/src/test/scala/com/wavesplatform/state/StateSyntheticBenchmark.scala
+++ b/benchmark/src/test/scala/com/wavesplatform/state/StateSyntheticBenchmark.scala
@@ -73,7 +73,7 @@ object StateSyntheticBenchmark {
     override def init(): Unit = {
       super.init()
 
-      val textScript    = "sigVerify(tx.bodyBytes,tx.proofs[0],tx.senderPk)"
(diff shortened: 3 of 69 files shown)


project/Dependencies.scala
```diff
@@ -65,7 +65,10 @@ object Dependencies {
   lazy val matcher = Seq(
     akkaModule("persistence"),
     akkaModule("persistence-tck") % "test",
-    "org.ethereum"                % "leveldbjni-all" % "1.18.3"
+    "com.github.dnvriend"         %% "akka-persistence-inmemory" % "2.5.15.1" % "test",
+    "com.typesafe.akka"           %% "akka-stream-kafka" % "1.0-M1",
+    // "org.apache.kafka"            % "kafka-clients" % "2.1.0",
```
vsuharnikov: Will remove later :)

PR #1757, opened 2018-12-17
NODE-1369: Asset distribution fix and integration test

disfenoid:


src/main/scala/com/wavesplatform/database/LevelDBWriter.scala
```diff
@@ -787,7 +785,7 @@ class LevelDBWriter(writableDB: DB, fs: FunctionalitySettings, val maxCacheSize:
         (for {
           addressId <- takeAfter(addressIds, maybeAddressId).take(count)
           history = db.get(Keys.assetBalanceHistory(addressId, assetId))
-          actualHeight <- history.partition(_ > height)._2.headOption
+          actualHeight <- history.filterNot(_ > height).headOption
```
petermz: Why not just `find` ?

PR #1792, opened 2019-01-11
NODE-1363-avoid-read-portfolio

potan:


src/main/scala/com/wavesplatform/state/diffs/CommonValidation.scala
```diff
@@ -67,23 +67,31 @@ object CommonValidation {
...
-          val oldAssetBalances = blockchain.portfolio(sender).assets
-          val balanceError = spendings.assets.collectFirst {
-            case (aid, delta) if oldAssetBalances.getOrElse(aid, 0L) + delta < 0 =>
-              val availableBalance = oldAssetBalances.getOrElse(aid, 0L)
-              GenericError(
-                "Attempt to transfer unavailable funds: Transaction application leads to negative asset " +
-                  s"'$aid' balance to (at least) temporary negative state, current balance is $availableBalance, " +
-                  s"spends equals $delta, result is ${availableBalance + delta}")
+        } else {
+          val balanceError = spendings.assets.iterator.flatMap {
```
phearnot: Instead of `flatMap(...).toSeq.headOption` one could just use `find`

PR #1803, opened 2019-01-15
NODE-1269 Moved transaction back/forward time offsets into config params

petermz:
JIRA: https://wavesplatform.atlassian.net/browse/NODE-1269
Moved two constants from `CommonValidation`:
```
  val MaxTimeTransactionOverBlockDiff: FiniteDuration     = 90.minutes
  val MaxTimePrevBlockOverTransactionDiff: FiniteDuration = 2.hours
```
into config parameters:
```
        max-transaction-time-back-offset = 120m
        max-transaction-time-forward-offset = 90m
```

### DIFF
--- a/src/main/scala/com/wavesplatform/state/diffs/CommonValidation.scala
+++ b/src/main/scala/com/wavesplatform/state/diffs/CommonValidation.scala
@@ -14,15 +14,12 @@ import com.wavesplatform.transaction.lease._
 import com.wavesplatform.transaction.smart.SetScriptTransaction
 import com.wavesplatform.transaction.transfer._
 
-import scala.concurrent.duration._
 import scala.util.{Left, Right}
 
 object CommonValidation {
 
-  val MaxTimeTransactionOverBlockDiff: FiniteDuration     = 90.minutes
-  val MaxTimePrevBlockOverTransactionDiff: FiniteDuration = 2.hours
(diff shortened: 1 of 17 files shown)


src/main/scala/com/wavesplatform/state/diffs/CommonValidation.scala
```diff
@@ -153,14 +150,14 @@ object CommonValidation {
...
       Left(Mistiming(s"Transaction ts ${tx.timestamp} is from far future. BlockTime: $time"))
     else Right(tx)
   }
 
-  def disallowTxFromPast[T <: Transaction](prevBlockTime: Option[Long], tx: T): Either[ValidationError, T] =
+  def disallowTxFromPast[T <: Transaction](settings: FunctionalitySettings, prevBlockTime: Option[Long], tx: T): Either[ValidationError, T] =
     prevBlockTime match {
-      case Some(t) if (t - tx.timestamp) > MaxTimePrevBlockOverTransactionDiff.toMillis =>
+      case Some(t) if (t - tx.timestamp) > settings.maxTransactionTimeBackOffset.toMillis =>
         Left(Mistiming(s"Transaction ts ${tx.timestamp} is too old. Previous block time: $prevBlockTime"))
```
msmolyakov: The errors should be more informative. How older? And how newer?

PR #1815, opened 2019-01-18
Node 915 scodec removing

DenisSerebryansky:


lang/shared/src/main/scala/com/wavesplatform/lang/v1/compiler/Types.scala
```diff
@@ -19,7 +19,7 @@ object Types {
   case class PARAMETERIZEDUNION(l: List[SINGLE]) extends PARAMETERIZED
   case object NOTHING                            extends FINAL { override val name = "Nothing"; override val l = List() }
   case object LONG                               extends REAL { override val name = "Int"; override val l = List(this) }
-  case object BYTEVECTOR                         extends REAL { override val name = "ByteVector"; override val l = List(this) }
+  case object BYTESTR                            extends REAL { override val name = "ByteStr"; override val l = List(this) }
```
petermz: Please keep `ByteVector` here

PR #1822, opened 2019-01-22
NODE-1281-cleaning

potan:
LevelDB merge
Duplicates names.


src/test/scala/com/wavesplatform/state/diffs/smart/predef/ScriptVersionsTest.scala
```diff
@@ -45,39 +45,6 @@ class ScriptVersionsTest extends FreeSpec with PropertyChecks with Matchers with
...
-            blockDiffEi should produce("duplicate variable names")
-          }
-
-          assertDiffEi(Seq(block(Seq(genesis)), block(Seq())), block(Seq(tx)), settings) { blockDiffEi =>
-            blockDiffEi shouldBe 'right
-          }
-      }
-    }
-
-    "does not have bindings defined in V2" in {
```
petermz: I think this test case should be preserved. It checks that things defined in one context do not leak into another.

PR #1831, opened 2019-01-24
SC-18 not accept expression with blockV2 before activation

mrkraft:


src/main/scala/com/wavesplatform/transaction/smart/script/v1/ScriptV1.scala
```diff
@@ -38,6 +38,26 @@ object ScriptV1 {
         val s = Array(version.toByte) ++ Serde.serialize(expr)
         ByteStr(s ++ crypto.secureHash(s).take(checksumLength))
       }
+    override val maxBlockVersion: Coeval[Int] = Coeval.evalOnce(calcMaxBlockVersion(expr))
+  }
+
+  def calcMaxBlockVersion(e: EXPR): Int = {
+    def horTraversal(queue: Iterable[EXPR]): Int = {
```
ismagin: We need to consider very deep expressions too.
Please finc `stack overflow check` tests and implement something similar

src/main/scala/com/wavesplatform/transaction/smart/script/v1/ScriptV1.scala
```diff
@@ -50,4 +70,5 @@ case class ScriptV2(version: Version, expr: Contract) extends Script {
       val s = Array(version.toByte) ++ ContractSerDe.serialize(expr)
       ByteStr(s ++ crypto.secureHash(s).take(checksumLength))
     }
+  override val maxBlockVersion: Coeval[Int] = Coeval.evalOnce(2)
```
ismagin: still, 1 and 2 are magic numbers, maybe we don't need the actual version but rather
```
override val containsBlockV2: Coeval[Boolean]
```
?

src/test/scala/com/wavesplatform/state/diffs/SetScriptTransactionDiffTest.scala
```diff
@@ -70,4 +71,36 @@ class SetScriptTransactionDiffTest extends PropSpec with PropertyChecks with Mat
...
+      genesis = GenesisTransaction.create(master, ENOUGH_AMT, ts).explicitGet()
+      expr    = BLOCKV2(LET("x", CONST_LONG(3)), CONST_BOOLEAN(true))
+      script  = ScriptV1(V1, expr, checkSize = false).explicitGet()
+      tx      = SetScriptTransaction.selfSigned(1, master, Some(script), 100000, ts + 1).explicitGet()
+    } yield (genesis, tx)
+
+    forAll(setup) {
+      case (genesis, tx) =>
+        assertDiffEi(Seq(block(Seq(genesis))), block(Seq(tx)), settingsUnactivated) { blockDiffEi =>
+          blockDiffEi should produce("SetScriptTransaction has not been activated yet")
```
ismagin: The error should be different. It's not SetScriptTransaction that has not been activated(it has), but rather RIDE4DApps feature that enables BlockV2 has not been activated yet. Mind the signature of  `activationBarrier`, it allows for that.

PR #1843, opened 2019-01-29
DEX-171 Expired order cleanup

phearnot:


src/test/scala/com/wavesplatform/matcher/AddressActorSpec.scala
```diff
@@ -0,0 +1,10 @@
+package com.wavesplatform.matcher
```
vsuharnikov: There is AddressActorSpecification

PR #1851, opened 2019-01-29
Added SignatureError to ApiError (compat)

kantefier:


src/main/scala/com/wavesplatform/api/http/ApiError.scala
```diff
@@ -268,6 +268,12 @@ case class TransactionNotAllowedByScript(tx: Transaction, log: Log, scriptSrc: S
   override lazy val json: JsObject = ScriptErrorJson(id, tx, message, scriptSrc, log)
 }
 
+case class SignatureError(error: String) extends ApiError {
+  override val id: Int          = 308
+  override val code: StatusCode = StatusCodes.InternalServerError
+  override val message: String  = s"Signature error: $message"
```
disfenoid: Maybe  "Signature error: $error"?

kantefier: oh, God, you're right of course :O

kantefier:
incremented SignatureError's id

PR #1853, opened 2019-01-30
NODE-1399: KeyBlock validation and effective balance calculation rework

Karasiq:


src/main/scala/com/wavesplatform/consensus/GeneratingBalanceProvider.scala
```diff
@@ -23,8 +24,10 @@ object GeneratingBalanceProvider {
         .get(BlockchainFeatures.SmallerMinimalGeneratingBalance.id)
         .exists(height >= _) && effectiveBalance >= MinimalEffectiveBalanceForGenerator2
 
-  def balance(blockchain: Blockchain, fs: FunctionalitySettings, height: Int, account: Address): Long = {
+  def balance(blockchain: Blockchain, fs: FunctionalitySettings, account: Address, block: ByteStr = ByteStr.empty): Long = {
```
petermz: Can you rename `block` to `blockId` here? `block` suggests something of type `Block`

src/test/scala/com/wavesplatform/history/BlockchainUpdaterKeyAndMicroBlockConflictTest.scala
```diff
@@ -0,0 +1,215 @@
...
+          totalRefTo = genesisBlock.signerData.signature,
+          base = Seq(transfer1),
+          micros = Seq(Seq(transfer2)),
+          signer = richAccount,
+          version = 3,
+          timestamp = System.currentTimeMillis()
+        )
+
+        val (keyBlock1, _) = unsafeChainBaseAndMicro(
+          totalRefTo = keyBlock.signerData.signature,
```
petermz: Could you add a scenario where `micros == Seq(Seq(), Seq(transfer))`, and `keyBlock1` refers to `micros(0)` rather than `keyBlock`?

src/test/scala/com/wavesplatform/history/BlockchainUpdaterKeyAndMicroBlockConflictTest.scala
```diff
@@ -0,0 +1,215 @@
...
+          secondAccount,
+          3,
+          System.currentTimeMillis()
+        )
+
+        (genesisBlock, leaseBlock, keyBlock, microBlocks, transferBlock, secondAccount)
+      }
+    }
+
+    private[this] def unsafeChainBaseAndMicro(totalRefTo: ByteStr,
```
petermz: These are copy-pasted, right? Could you move them somewhere, TestHelpers maybe?

src/main/scala/com/wavesplatform/consensus/GeneratingBalanceProvider.scala
Karasiq: Ok

src/test/scala/com/wavesplatform/history/BlockchainUpdaterKeyAndMicroBlockConflictTest.scala
Karasiq: Ok

src/test/scala/com/wavesplatform/history/BlockchainUpdaterKeyAndMicroBlockConflictTest.scala
Karasiq: Ok

Karasiq:
Fixed

PR #1869, opened 2019-02-05
SC-58 Define already defined var

mrkraft:


ismagin:
Hm, this is a legit expression:
```
let a = 8
func x(a:String) = {
	throw()
}
x(a)
```
therefore

`let a = 8

@Callable(i)
func x(a:String) = {
 thow()
}

should be legit too.


So we sohuld either deny annotation-bindings overwriting something or just be ok with that if it compiles and runs the same way.

mrkraft:
let and callable have the same name = compilation should fail
let and arg have the same name = compilation should succeed
callable and arg have the same name = compilation should fail

PR #1884, opened 2019-02-08
Sc 4 check smartassets for fee

potan:


src/main/scala/com/wavesplatform/state/diffs/ContractInvocationTransactionDiff.scala
```diff
@@ -80,6 +102,16 @@ object ContractInvocationTransactionDiff {
                     )
                     _ <- Either.cond(true, (), ValidationError.NegativeAmount(-42, "")) //  - sum doesn't overflow
                     _ <- Either.cond(true, (), ValidationError.NegativeAmount(-42, "")) //  - whatever else tranfser/massTransfer ensures
+                    _ <- {
+                      val minWaves = (tx.checkedAssets().count(blockchain.hasAssetScript) +
```
ismagin: Hm, but if contract sends 1 smart token, we should require fees for 2 invokations, not just 1, right>?

src/main/scala/com/wavesplatform/state/diffs/ContractInvocationTransactionDiff.scala
```diff
@@ -80,6 +102,16 @@ object ContractInvocationTransactionDiff {
...
+                        ps.count(_._3.fold(false)(blockchain.hasAssetScript))) * ScriptExtraFee + FeeConstants(ContractInvocationTransaction.typeId) * FeeUnit
+                      Either.cond(
+                        minWaves <= wavesFee,
+                        (),
+                        GenericError(s"Fee in ${tx.assetFee._1
+                          .fold("WAVES")(_.toString)} for ${tx.builder.classTag} does not exceed minimal value of $minWaves WAVES: ${tx.assetFee._2}")
```
ismagin: We can specifty error better here, explaining why exactly fees are not enough and how many do we expect

src/main/scala/com/wavesplatform/state/diffs/ContractInvocationTransactionDiff.scala
ismagin: this better be tested

PR #1894, opened 2019-02-12
sc-13: trampolinezation

rigidus:


potan:
> Why not use Coeval for trampoline?
> Please consult the team

There isn't tail recursion.

PR #1901, opened 2019-02-14
Sc 11 cons

potan:


lang/shared/src/main/scala/com/wavesplatform/lang/v1/parser/BinaryOperation.scala
```diff
@@ -13,12 +13,13 @@ sealed abstract class BinaryOperation {
 
 object BinaryOperation {
 
-  val opsByPriority: List[List[BinaryOperation]] = List(
-    List(OR_OP, AND_OP),
-    List(EQ_OP, NE_OP),
-    List(GT_OP, GE_OP, LT_OP, LE_OP),
-    List(SUM_OP, SUB_OP),
-    List(MUL_OP, DIV_OP, MOD_OP)
+  val opsByPriority: List[Either[List[BinaryOperation], List[BinaryOperation]]] = List(
```
ismagin: maybe a small comment here? like
```
No monadic notion here, Left and Right mean `left-assosiative and `right-assosiative`
```

lang/shared/src/main/scala/com/wavesplatform/lang/v1/evaluator/ctx/impl/PureContext.scala
```diff
@@ -357,7 +340,10 @@ object PureContext {
     uNot
   )
 
-  private lazy val vars: Map[String, ((FINAL, String), LazyVal)] = Map(("unit", ((UNIT, "Single instance value"), LazyVal(EitherT.pure(unit)))))
+  private lazy val vars: Map[String, ((FINAL, String), LazyVal)] = Map(
+    ("unit", ((UNIT, "Single instance value"), LazyVal(EitherT.pure(unit)))),
+    ("nil", ((LIST(NOTHING), "empty list of any type"), LazyVal(EitherT.pure(ARR(IndexedSeq.empty[EVALUATED])))))
```
ismagin: nil should only appear in `V3` !

PR #1905, opened 2019-02-15
Sc 80 handler utils scipt decompile

rigidus:


src/test/scala/com/wavesplatform/http/AddressRouteSpec.scala
```diff
@@ -182,10 +183,12 @@ class AddressRouteSpec
     Get(routePath(s"/scriptInfo/${allAddresses(3)}")) ~> route ~> check {
       val response = responseAs[JsObject]
       (response \ "address").as[String] shouldBe allAddresses(3)
-      (response \ "script").as[String] shouldBe "base64:AAIDAAAAAAAAAAAAAAAAAAAAAQAAAAF0AQAAAAZ2ZXJpZnkAAAAABiDCPeI="
-      (response \ "scriptText").as[String] shouldBe Decompiler(
-        testContract,
-        Monoid.combineAll(Seq(PureContext.build(com.wavesplatform.lang.StdLibVersion.V3), CryptoContext.build(Global))).decompilerContext)
+      // [WAIT] (response \ "script").as[String] shouldBe "base64:AAIDAAAAAAAAAAAAAAAAAAAAAQAAAAF0AAAABnZlcmlmeQAAAAAAAAABBt/lCgQ="
```
ismagin: what is this

src/main/scala/com/wavesplatform/api/http/UtilsApiRoute.scala
```diff
@@ -26,7 +26,38 @@ case class UtilsApiRoute(timeService: Time, settings: RestAPISettings) extends A
...
+      )
+    ))
+  @ApiResponses(
+    Array(
+      new ApiResponse(code = 200, message = "string or error")
+    ))
+  def decompile: Route = path("script" / "decompile") {
+    (post & entity(as[String])) { code =>
+        val ret = Script.fromBase64String(code) match {
+          case Left(err)     => "Wrong input"
```
ismagin: Hm, we shouldn't return 200 here, more like a Bad request, similar to what we do when user provides bad address or any general malformed bytes

PR #1908, opened 2019-02-15
SC-27-dataentry-to-evaluated

potan:
s/DataEntry/EVALUATED


src/main/scala/com/wavesplatform/api/http/ContractInvocationRequest.scala
```diff
@@ -1,35 +1,56 @@
...
 
 import scala.annotation.meta.field
 
 object ContractInvocationRequest {
 
-  case class FunctionCallPart(function: String, args: List[DataEntry[_]])
-  implicit val functionCallReads = Json.reads[FunctionCallPart]
+  case class FunctionCallPart(function: String, args: List[EVALUATED])
 
+  implicit val EVALUATED_Reads = new Reads[EVALUATED] {
```
ismagin: better EvaluatedReads since we don't use underscore in names

PR #1923, opened 2019-02-19
NODE-1480 blacklistsuite to sync package

finnoio:


it/src/test/scala/com/wavesplatform/it/sync/BlacklistTestSuite.scala
```diff
@@ -0,0 +1,45 @@
...
+  private def primaryNode = dockerNodes().last
+
+  private def otherNodes = dockerNodes().init
+
+  "primary node should blacklist other nodes" in {
+    otherNodes.foreach(n => primaryNode.blacklist(n.containerNetworkAddress))
+
+    val expectedBlacklistedPeers = nodes.size - 1
+
+    primaryNode.waitFor[Seq[BlacklistedPeer]](s"blacklistedPeers.size == $expectedBlacklistedPeers")(
```
disfenoid: It would be great to put it in separate method (smth like node.getBlacklistedPeers) in sync api.

PR #1924, opened 2019-02-19
sc-40: test fo contract invocation

rigidus:


it/src/test/scala/com/wavesplatform/it/sync/utils/TransactionSerializeSuite.scala
```diff
@@ -349,6 +350,19 @@ class TransactionSerializeSuite extends BaseTransactionSuite with TableDrivenPro
     .right
     .get
 
+  private val contractInvocation = ContractInvocationTransaction
+    .create(
+      None,
+      publicKey,
+      where_i_can_function_for_this_call?,
```
rigidus: @ismagin : need help

PR #1948, opened 2019-02-25
SC-84-postfix-call

potan:


lang/jvm/src/test/scala/com/wavesplatform/lang/IntegrationTest.scala
```diff
@@ -428,5 +428,13 @@ class IntegrationTest extends PropSpec with PropertyChecks with ScriptGen with M
       """.stripMargin
     eval[EVALUATED](src) shouldBe Left("test fail")
   }
+  property("postfix syntax") {
+    val src =
+      """
+        |let x = true
+        |x.ensure("test fail")
```
ismagin: please add a few more lines:
0-args, may-args, not just 1-arg example

lang/jvm/src/test/scala/com/wavesplatform/lang/IntegrationTest.scala
```diff
@@ -437,4 +438,25 @@ class IntegrationTest extends PropSpec with PropertyChecks with ScriptGen with M
...
+    val src =
+      """ let x = 5 ; x.fraction(7,2) """
+    eval[EVALUATED](src) shouldBe Right(CONST_LONG(17L))
+  }
+
+  property("postfix syntax (users defined function)") {
+    val src =
+      """
+        |func dub(s:String) = { s+s }
+        |("qwe").dub()
```
ismagin: "qwe".dub() should work

PR #1950, opened 2019-02-25
UTX pool error logging

Karasiq:


src/main/scala/com/wavesplatform/network/UtxPoolSynchronizer.scala
```diff
@@ -50,8 +50,14 @@ object UtxPoolSynchronizer extends ScorexLogging {
                 xs.foreach {
                   case (_, tx) =>
                     utx.putIfNew(tx) match {
-                      case Right((true, _)) => allChannels.write(RawBytes.from(tx), channelMatcher)
-                      case _                =>
+                      case Right((true, _)) =>
+                        allChannels.write(RawBytes.from(tx), channelMatcher)
+
+                      case Left(error)  =>
+                        log.error(s"Error adding transaction to UTX pool: $error")
```
phearnot:
```suggestion
                        log.debug(s"Error adding transaction to UTX pool: $error")
```
This isn't actually an error.

PR #1974, opened 2019-02-28
NODE-1491: UTX cleanup optimization

Karasiq:


src/main/scala/com/wavesplatform/utx/UtxPoolImpl.scala
```diff
@@ -125,12 +135,28 @@ class UtxPoolImpl(time: Time,
...
-    removeAll(transactionsToRemove)
+  private[this] def removeInvalid(list: Iterable[Transaction] = this.transactions.values.asScala, checkFuture: Int = smartCleanupBlocksAhead): Unit = {
+    val (transactionsToRemove, transactionsToCheckInNextBlock, transactionsPreChecked) = list.map { t =>
+      def isPreChecked(heightOffset: Int = 0) = this.transactionsChecked.contains((t.id(), blockchain.height + heightOffset))
+      val currentlyInvalid = !isPreChecked() && TransactionDiffer(fs, blockchain.lastBlockTimestamp, time.correctedTime(), blockchain.height)(blockchain, t).isLeft
+
+      if (currentlyInvalid) {
+        (Some(t), None, None)
+      } else {
+        val futureInvalid = (1 to checkFuture).exists { heightOffset =>
```
phearnot: There's no need to eagerly check each transaction for validity. Instead, cleanup should be initiated whenever blockchain height changes (and another cleanup task is not running!)

```scala
val height: Observable[Int] = ...
val h: Observable[Int] = ???
    h.observeOn(utxScheduler).whileBusyDropEvents.foreach { _ =>
      transactions.entrySet().removeIf { e =>
        TransactionDiffer(..)(blockchain, e.getValue) match {
          case Right(_) => false
          case Left(e) =>
            log.debug(s"Removing transaction: $e")
            true
        }
      }
    }
```

Karasiq: Instant check will incur constant CPU load on a new block arrival, and if a full check takes more time, than a block time is, it will skip removal of an invalid transactions.

PR #1981, opened 2019-03-01
NODE-1453 Added metric to network broadcast

petermz:
Added metric `network-broadcast-time` with an additional tag, `object`, that contains type of object being sent, such as `BlockForged`, `LocalScoreChanged`, `RawBytes` etc

### DIFF
--- a/src/main/scala/com/wavesplatform/network/package.scala
+++ b/src/main/scala/com/wavesplatform/network/package.scala
@@ -17,10 +17,14 @@ import monix.reactive.Observable
 import com.wavesplatform.block.Block
 import com.wavesplatform.common.state.ByteStr
 import com.wavesplatform.transaction.Transaction
+import kamon.Kamon
+import kamon.metric.MeasurementUnit.time.milliseconds
 
 import scala.concurrent.duration._
 
 package object network extends ScorexLogging {
+  private val broadcastTimeStats = Kamon.histogram("network-broadcast-time", milliseconds)
+
   def inetSocketAddress(addr: String, defaultPort: Int): InetSocketAddress = {
     val uri = new URI(s"node://$addr")
     if (uri.getPort < 0) new InetSocketAddress(addr, defaultPort)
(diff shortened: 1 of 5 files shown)


src/main/scala/com/wavesplatform/network/package.scala
```diff
@@ -17,10 +17,14 @@ import monix.reactive.Observable
 import com.wavesplatform.block.Block
 import com.wavesplatform.common.state.ByteStr
 import com.wavesplatform.transaction.Transaction
+import kamon.Kamon
+import kamon.metric.MeasurementUnit.time.milliseconds
 
 import scala.concurrent.duration._
 
 package object network extends ScorexLogging {
+  private val broadcastTimeStats = Kamon.histogram("network-broadcast-time", milliseconds)
```
phearnot:
```suggestion
  private val broadcastTimeStats = Kamon.timer("network-broadcast-time")
```

src/main/scala/com/wavesplatform/network/package.scala
```diff
@@ -69,9 +73,16 @@ package object network extends ScorexLogging {
 
     def broadcast(message: AnyRef, except: Set[Channel]): ChannelGroupFuture = {
       logBroadcast(message, except)
-      allChannels.writeAndFlush(message, { (channel: Channel) =>
-        !except.contains(channel)
-      })
+      val t0 = System.currentTimeMillis()
```
phearnot:
```suggestion
      val st = broadcastTimeStats.refine("object", message.getClass.getSimpleName).start()
```

src/main/scala/com/wavesplatform/network/package.scala
```diff
@@ -69,9 +73,16 @@ package object network extends ScorexLogging {
...
+      allChannels
+        .writeAndFlush(message, { (channel: Channel) =>
+          !except.contains(channel)
+        })
+        .addListener { _: ChannelGroupFuture =>
+          val t1  = System.currentTimeMillis()
+          val tpe = message.getClass.getSimpleName
+          broadcastTimeStats.refine("object", tpe).record(t1 - t0)
```
phearnot:
```suggestion
          st.stop()
```

PR #1998, opened 2019-03-07
Dex 207 order fee settings in rest api

DenisSerebryansky:


src/main/scala/com/wavesplatform/matcher/model/OrderValidator.scala
```diff
@@ -133,17 +144,68 @@ object OrderValidator {
...
+    lazy val spentAmount   = order.getSpendAmount(order.amount, order.price).explicitGet()
+
+    val amountFactor = percentSettings.assetType match {
+      case AssetType.AMOUNT    => order.amount
+      case AssetType.PRICE     => if (order.orderType == OrderType.BUY) spentAmount else receiveAmount
+      case AssetType.RECEIVING => receiveAmount
+      case AssetType.SPENDING  => spentAmount
+    }
+
+    multiplyLongByDouble(amountFactor, percentSettings.minFee / 100)
```
vsuharnikov: Could you inline the implementation of `multiplyLongByDouble` here? `multiplyLongByDouble` rounds the result using half up. This conflicts with the usual way we multiple longs with doubles: `((x: Long) * (y: Double)).toLong`. `multiplyLongByDouble` has the right behaviour only in `getMinValidFee`.

src/main/scala/com/wavesplatform/matcher/model/OrderValidator.scala
```diff
@@ -133,17 +144,68 @@ object OrderValidator {
...
+  }
+
+  def validateOrderFee(order: Order, orderFeeSettings: OrderFeeSettings): ValidationResult = {
+    for {
+      _ <- (Right(order): ValidationResult)
+        .ensure(s"Matcher's fee asset in order (${order.matcherFeeAssetId}) does not meet matcher's settings requirements") { order =>
+          val isMatcherFeeAssetValid = orderFeeSettings match {
```
vsuharnikov: `lazy val`

src/main/scala/com/wavesplatform/matcher/model/OrderValidator.scala
```diff
@@ -133,17 +144,68 @@ object OrderValidator {
...
+            case _: FixedWavesSettings            => order.matcherFeeAssetId.isEmpty
+            case FixedSettings(defaultAssetId, _) => order.matcherFeeAssetId == defaultAssetId
+            case PercentSettings(assetType, _)    => order.matcherFeeAssetId == getValidFeeAsset(order, assetType)
+          }
+          order.version != 3 || isMatcherFeeAssetValid
+        }
+      _ <- (Right(order): ValidationResult)
+        .ensure(s"Matcher's fee (${order.matcherFee}) is less than minimally admissible one") { order =>
+          val isMatcherFeeValid = orderFeeSettings match {
```
vsuharnikov: `lazy val`

src/main/scala/com/wavesplatform/matcher/model/OrderValidator.scala
```diff
@@ -133,17 +144,68 @@ object OrderValidator {
...
+            case FixedWavesSettings(wavesMinFee)  => order.matcherFee >= wavesMinFee
```
vsuharnikov: Actually, we should add to the `wavesMinFee` additional `CommonValidation.ScriptExtraFee` for each paragraph:
1. Scripted matcher's account
2. Scripted amount asset
3. Scripted price asset

vsuharnikov: Also it looks like this validation should be in `blockchainAware` by this reason :(

src/main/scala/com/wavesplatform/settings/fee/OrderFeeSettings.scala
```diff
@@ -0,0 +1,129 @@
...
+import scala.util.Try
+
+object OrderFeeSettings {
+
+  type ErrorsListOr[A] = Validated[List[String], A]
+
+  sealed trait OrderFeeSettings {
+
+    /** Returns json for order fee settings taking into account fee that should be paid for matcher's account script invocation */
+    def getOrderFeeJson(minMarcherFee: Long): Coeval[JsObject] = Coeval.evalOnce {
```
vsuharnikov: Just `json` as other entities has?

DenisSerebryansky: Which one?
Orders and transactions have `Coeval[JsObject]` type of `json` field

src/main/scala/com/wavesplatform/matcher/model/OrderValidator.scala
DenisSerebryansky: Done

src/main/scala/com/wavesplatform/matcher/model/OrderValidator.scala
DenisSerebryansky: Done

src/main/scala/com/wavesplatform/matcher/model/OrderValidator.scala
DenisSerebryansky: Done

src/main/scala/com/wavesplatform/matcher/model/OrderValidator.scala
DenisSerebryansky: We already have such validation in `blockchainAware`

src/main/scala/com/wavesplatform/settings/fee/OrderFeeSettings.scala
vsuharnikov: I meant, may be we use the same name :) We are in OrderFeeSettings, `orderFee` here is unnecessary

DenisSerebryansky: Done

src/main/scala/com/wavesplatform/matcher/model/OrderValidator.scala
```diff
@@ -106,7 +109,13 @@ object OrderValidator {
       exchangeTx.flatMap(verifySmartToken(blockchain, assetId, _))
     }
 
-    lazy val mof = ExchangeTransactionCreator.minFee(blockchain, matcherAddress, order.assetPair)
+    val baseFee = orderFeeSettings match {
+      case FixedWavesSettings(fee) => fee
+      case _                       => exchangeTransactionCreationFee
+    }
```
vsuharnikov: Actually, this isn't the true for `case _`. I agree, this case isn't real, but IMO it'd better to do something like:
```scala
def mof(baseFee: Long) = ExchangeTransactionCreator.minFee(blockchain, matcherAddress, order.assetPair, baseFee)
// ...
        .ensure(MatcherError.FeeNotEnough(mof, order.matcherFee, None)) { o =>
          orderFeeSettings match {
            case FixedWavesSettings(baseFee) => o.matcherFee >= mof(baseFee)
            case _                           => true
          }
        }
```

src/main/scala/com/wavesplatform/matcher/api/MatcherApiRoute.scala
```diff
@@ -132,7 +133,13 @@ case class MatcherApiRoute(assetPairBuilder: AssetPairBuilder,
   @Path("/settings")
   @ApiOperation(value = "Matcher Settings", notes = "Get matcher settings", httpMethod = "GET")
   def getSettings: Route = (path("settings") & get) {
-    complete(StatusCodes.OK -> Json.obj("priceAssets" -> matcherSettings.priceAssets))
+
+    complete(
+      StatusCodes.OK -> Json.obj(
+        "priceAssets" -> matcherSettings.priceAssets,
+        "order-fee"   -> matcherSettings.orderFee.getJson(minMatcherFee).value
```
vsuharnikov: Looks like we have a different naming style. It'd better to avoid this in API. Sorry, but we need to use camelCase here (`orderFee`), because people already use `priceAssets`. And in `getJson` too.

src/main/scala/com/wavesplatform/settings/fee/OrderFeeSettings.scala
```diff
@@ -6,19 +6,44 @@ import com.wavesplatform.settings.fee.AssetType.AssetType
...
+            )
+          case FixedSettings(defaultAssetId, minFee) =>
+            "fixed" -> Json.obj(
+              "asset-id" -> defaultAssetId.map(_.base58),
+              "min-fee"  -> minFee
+            )
+          case PercentSettings(assetType, minFee) =>
+            "percent" -> Json.obj(
+              "type"    -> assetType,
+              "min-fee" -> minFee
```
vsuharnikov: caseCase 🙏

src/main/scala/com/wavesplatform/matcher/model/OrderValidator.scala
vsuharnikov: Slightly better way:
```scala
for {
// …
_ <- orderFeeSettings match {
  case FixedWavesSettings(baseFee) =>
    val mof = ExchangeTransactionCreator.minFee(blockchain, matcherAddress, order.assetPair, baseFee)
    Either.cond(o.matcherFee >= mof, success, MatcherError.FeeNotEnough(mof, order.matcherFee, None))
  case _                           => success
}
// …
]
```

vsuharnikov: Slightly better way:
```scala
for {
// …
_ <- orderFeeSettings match {
  case FixedWavesSettings(baseFee) =>
    val mof = ExchangeTransactionCreator.minFee(blockchain, matcherAddress, order.assetPair, baseFee)
    Either.cond(o.matcherFee >= mof, success, MatcherError.FeeNotEnough(mof, order.matcherFee, None))
  case _                           => success
}
// …
]
```

src/main/scala/com/wavesplatform/matcher/api/MatcherApiRoute.scala
DenisSerebryansky: Done

src/main/scala/com/wavesplatform/matcher/model/OrderValidator.scala
DenisSerebryansky: Done

src/main/scala/com/wavesplatform/settings/fee/OrderFeeSettings.scala
DenisSerebryansky: Done

PR #2015, opened 2019-03-13
NODE-1456 Added two new modes to Explorer

petermz:
- `AP address`: dump an account portfolio
- `APS`: dump all WAVES and assets balances (takes about an hour)


src/main/scala/com/wavesplatform/Explorer.scala
```diff
@@ -260,6 +262,51 @@ object Explorer extends ScorexLogging {
...
+          val wbals = reader.wavesDistribution(blockchainHeight).explicitGet().toSeq
+          println(s"\nWAVES balances (${wbals.size} addresses)")
+          val addrs = mutable.Set[Address]()
+          wbals.sortBy(_._1.toString).foreach {
+            case (addr, bal) =>
+              println(s"$addr : $bal")
+              addrs += addr
+          }
+
+          val assets = (for {
```
phearnot: Please use `db.iterateOver(Keys.assetInfoHistory(..)`. Also there's an ordering for `ByteStr`, which sorts byte strings lexicographically.

PR #2017, opened 2019-03-13
NODE-1519: REST API /utils/ limited execution context

Karasiq:


phearnot:
We'll just wait for decompiler to be fixed.

PR #2022, opened 2019-03-13
NODE-1495: gRPC API

Karasiq:


src/main/scala/com/wavesplatform/api/common/CommonBlocksApi.scala
```diff
@@ -0,0 +1,82 @@
...
+import com.wavesplatform.block.Block.BlockId
+import com.wavesplatform.block.BlockHeader
+import com.wavesplatform.common.state.ByteStr
+import com.wavesplatform.protobuf.block.VanillaBlock
+import com.wavesplatform.state.Blockchain
+import monix.reactive.Observable
+
+private[api] class CommonBlocksApi(blockchain: Blockchain) {
+  def blocksRange(fromHeight: Int, toHeight: Int): Observable[(VanillaBlock, Int)] = {
+    def fixHeight(h: Int) = if (h <= 0) blockchain.height + h else h
```
dvshur: Мне кажется при h < 0 нужно вернуть ошибку валидации параметров (аналог 400 Bad Request в http)

Karasiq: Это для ренжей с конца как в питоне

node/src/main/resources/logback.xml
```diff
@@ -37,6 +37,7 @@
 
     <logger name="com.wavesplatform.network.PeerSynchronizer" level="DEBUG"/>
     <logger name="com.wavesplatform.transaction.smart" level="INFO"/>
+    <logger name="com.wavesplatform.network.RxScoreObserver$" level="INFO"/>
```
phearnot: Is this really needed?

grpc/src/main/protobuf/blocks_api.proto
```diff
@@ -0,0 +1,43 @@
...
+package com.wavesplatform.api.grpc;
+
+import "block.proto";
+import "google/protobuf/empty.proto";
+import "google/protobuf/wrappers.proto";
+
+service BlocksApi {
+    rpc GetBlock (BlockRequest) returns (BlockAndHeight);
+    rpc GetBlocksRange (BlocksRangeRequest) returns (stream BlockAndHeight);
+    rpc CalcBlocksDelay (BlocksDelayRequest) returns (google.protobuf.UInt64Value);
```
phearnot: How about we just drop this method? Block delays can easily be calculated from block headers.

grpc/src/main/protobuf/transactions_api.proto
```diff
@@ -0,0 +1,34 @@
...
+message TransactionByIdRequest {
+    bytes transaction_id = 1;
+}
+
+message CalculateFeeResponse {
+    bytes asset_id = 1;
+    uint64 amount = 2;
+}
+
+message TransactionSignRequest {
```
phearnot:
```suggestion
message SignTransactionRequest {
```

grpc/src/main/protobuf/blocks_api.proto
```diff
@@ -0,0 +1,43 @@
...
+    oneof request {
+        bytes block_id = 1;
+        int32 height = 2;
+        bytes parent_id = 3;
+    }
+
+    bool include_transactions = 100;
+}
+
+message BlocksRangeRequest {
```
phearnot:
```suggestion
message BlockRangeRequest {
```

grpc/src/main/protobuf/transactions_api.proto
```diff
@@ -0,0 +1,34 @@
+syntax = "proto3";
+package com.wavesplatform.api.grpc;
+import "recipient.proto";
+import "transaction.proto";
+import "google/protobuf/empty.proto";
+
+service TransactionsApi {
+    rpc GetTransactionsByAddress (TransactionsByAddressRequest) returns (stream SignedTransaction);
+    rpc GetTransactionById (TransactionByIdRequest) returns (SignedTransaction);
```
phearnot: Both `GetTransaction*` methods should return `SignedTransactionAndHeight`. Also how about
```suggestion
    rpc GetTransactions (TransactionsRequest) returns (SignedTransactionWithHeight);
    rpc GetTransaction (TransactionRequest) returns (SignedTransactionWithHeight);
```
`GetTransactions` also returns transactions for an alias, not only for an address.

grpc/src/main/protobuf/blocks_api.proto
```diff
@@ -0,0 +1,43 @@
...
+    rpc GetBlocksRange (BlocksRangeRequest) returns (stream BlockAndHeight);
+    rpc CalcBlocksDelay (BlocksDelayRequest) returns (google.protobuf.UInt64Value);
+    rpc GetCurrentHeight (google.protobuf.Empty) returns (google.protobuf.UInt32Value);
+}
+
+message BlockRequest {
+    oneof request {
+        bytes block_id = 1;
+        int32 height = 2;
+        bytes parent_id = 3;
```
phearnot: for consistency:
```suggestion
        bytes reference = 3;
```

grpc/src/main/protobuf/blocks_api.proto
```diff
@@ -0,0 +1,43 @@
...
+
+    bool include_transactions = 100;
+}
+
+message BlocksDelayRequest {
+    bytes block_id = 1;
+    uint32 block_num = 2;
+}
+
+message BlockAndHeight {
```
phearnot: Height is a property of the block, not an independent entity, so maybe
```suggestion
message BlockWithHeight {
```

grpc/src/main/protobuf/transactions_api.proto
```diff
@@ -0,0 +1,34 @@
...
+    rpc GetUnconfirmedTransactions (google.protobuf.Empty) returns (stream SignedTransaction);
+    rpc GetUnconfirmedTransactionById (TransactionByIdRequest) returns (SignedTransaction);
+    rpc CalculateFee (Transaction) returns (CalculateFeeResponse);
+    rpc SignTransaction (TransactionSignRequest) returns (SignedTransaction);
+    rpc BroadcastTransaction (SignedTransaction) returns (SignedTransaction);
+}
+
+message TransactionsByAddressRequest {
+    Recipient address = 1;
```
phearnot:
```suggestion
    Recipient recipient = 1;
```

node/src/main/resources/application.conf
```diff
@@ -346,9 +346,7 @@ waves {
     supported = []
   }
 
-  extensions = [
-    # com.wavesplatform.matcher.Matcher
-  ]
```
phearnot: Listing all our plugins here seemed like a good idea. Is there a reason to change this?

PR #2034, opened 2019-03-14
SC-112-limit-loops-in-deserializer

potan:


lang/shared/src/main/scala/com/wavesplatform/lang/v1/Serde.scala
```diff
@@ -95,8 +102,12 @@ object Serde {
           .now((bb.getFunctionHeader, bb.getInt))
           .flatMap {
             case (header, argc) =>
-              val args: List[Coeval[EXPR]] = (1 to argc).map(_ => desAux(bb))(collection.breakOut)
-              args.sequence[Coeval, EXPR].map(FUNCTION_CALL(header, _))
+              if (true || argc <= (bb.limit() - bb.position())/2) {
```
ismagin: `if(true || ...)` ?

lang/shared/src/main/scala/com/wavesplatform/lang/v1/Serde.scala
```diff
@@ -58,7 +58,14 @@ object Serde {
       case DEC_FUNC =>
         for {
           name <- Coeval.now(bb.getString)
-          args <- Coeval.now(for (_ <- 1 to bb.getInt) yield bb.getString)
+          args <- ({
+            val argsCnt = bb.getInt
+            if (true || argsCnt <= (bb.limit() - bb.position())/2) {
```
ismagin: if(true || ...) ?

PR #2056, opened 2019-03-18
SC-121-no-check-complexity

potan:


src/main/scala/com/wavesplatform/api/http/UtilsApiRoute.scala
```diff
@@ -48,7 +48,7 @@ case class UtilsApiRoute(timeService: Time, settings: RestAPISettings) extends A
     ))
   def decompile: Route = path("script" / "decompile") {
     (post & entity(as[String])) { code =>
-      Script.fromBase64String(code) match {
+      Script.fromBase64String(code, checks = false) match {
```
ismagin: unsound parameter name, please change to `checkSize` and remove default value -- especially false

PR #2061, opened 2019-03-19
NODE-1529: Fix REST API execution contexts

Karasiq:


src/main/scala/com/wavesplatform/http/DebugApiRoute.scala
```diff
@@ -63,6 +64,8 @@ case class DebugApiRoute(ws: WavesSettings,
 
   import DebugApiRoute._
 
+  private[this] val rollbackScheduler = Scheduler.singleThread("debug-rollback")
```
phearnot: There's no need for a dedicated scheduler. `rollbackScheduler` is only used to perform some housekeeping, which can easily be done on one of API threads.

src/main/scala/com/wavesplatform/http/BroadcastRoute.scala
```diff
@@ -5,26 +5,39 @@ import com.wavesplatform.network._
 import com.wavesplatform.transaction.{Transaction, ValidationError}
 import com.wavesplatform.utx.UtxPool
 import io.netty.channel.group.ChannelGroup
+import monix.execution.Scheduler
 
-import scala.concurrent.Future
+import scala.concurrent.{ExecutionContextExecutor, Future}
+
+private[wavesplatform] object BroadcastRoute {
+  lazy val executionContext = new ExecutionContextExecutor {
```
phearnot: Why isn't the scheduler just instantiated here the way it is done in `ApiMarshallers`?

Karasiq: To be compatible with akka-http directive

src/main/scala/com/wavesplatform/settings/RestAPISettings.scala
```diff
@@ -3,28 +3,30 @@ package com.wavesplatform.settings
...
 
-case class RestAPISettings(enable: Boolean,
-                           bindAddress: String,
-                           port: Int,
-                           apiKeyHash: String,
-                           cors: Boolean,
-                           apiKeyDifferentHost: Boolean,
-                           transactionByAddressLimit: Int,
-                           distributionAddressLimit: Int)
+final case class RestAPISettings(enable: Boolean,
```
phearnot: Is there really a reason to make this final?

src/main/scala/com/wavesplatform/settings/RestAPISettings.scala
```diff
@@ -3,28 +3,30 @@ package com.wavesplatform.settings
...
-      transactionByAddressLimit = config.as[Int](s"$configPath.transactions-by-address-limit"),
-      distributionAddressLimit = config.as[Int](s"$configPath.distribution-address-limit")
+      enable = config.as[Boolean]("enable"),
+      bindAddress = config.as[String]("bind-address"),
+      port = config.as[Int]("port"),
+      apiKeyHash = config.as[String]("api-key-hash"),
+      cors = config.as[Boolean]("cors"),
+      apiKeyDifferentHost = config.as[Boolean]("api-key-different-host"),
+      transactionByAddressLimit = config.as[Int]("transactions-by-address-limit"),
+      distributionAddressLimit = config.as[Int]("distribution-address-limit")
```
phearnot: Maybe just remove `fromConfig`, and use `config.as[RestAPISettings]`, just like with `WalletSettings` and `UtxSettings`?

src/main/scala/com/wavesplatform/http/ApiMarshallers.scala
```diff
@@ -1,58 +1,66 @@
...
 
-  private lazy val jsonStringMarshaller =
+  private[this] lazy val jsonStringMarshaller =
     Marshaller.stringMarshaller(`application/json`)
 
-  implicit def playJsonUnmarshaller[A](implicit reads: Reads[A]): FromEntityUnmarshaller[A] =
-    jsonStringUnmarshaller map { data =>
-      val json = nonFatalCatch.withApply(t => throw PlayJsonException(cause = Some(t)))(Json.parse(data))
+  implicit def playJsonUnmarshaller[A](implicit reads: Reads[A]): FromEntityUnmarshaller[A] = {
+    jsonStringUnmarshaller.andThen(Unmarshaller(implicit ec =>
```
phearnot: This looks just like map, which transforms unmarshaller's output on the provided execution context

PR #2064, opened 2019-03-20
Sc 133 add invoke script tx fields

mrkraft:


ismagin:
we'll merge from branch 0.17

PR #2068, opened 2019-03-21
Data 264 update notifications

dvshur:
Sending blockchain updates (including balances and lease updates) from Waves Node to a consumer by TCP.


dvshur:
Moving to DATA-291

PR #2080, opened 2019-03-22
SC-137 NegativeAmount error when payment amount is 0

xrtm000:


it/src/test/scala/com/wavesplatform/it/sync/transactions/IssueTransactionSuite.scala
```diff
@@ -78,9 +78,9 @@ class IssueTransactionSuite extends BaseTransactionSuite with TableDrivenPropert
   val invalidAssetValue =
     Table(
       ("assetVal", "decimals", "message"),
-      (0l, 2, "negative amount"),
+      (0l, 2, "non-positive amount"),
```
ismagin: this breaks backward compatibility.
not only we change message, but also error code, I don't think UI team willbe ok with this.

PR #2082, opened 2019-03-25
NODE-1542: Invalidate UTX synchronizer transactions cache on new block arrival

Karasiq:


src/main/scala/com/wavesplatform/Application.scala
```diff
@@ -179,7 +179,7 @@ class Application(val actorSystem: ActorSystem, val settings: WavesSettings, con
 
     rxExtensionLoaderShutdown = Some(sh)
 
-    UtxPoolSynchronizer.start(utxStorage, settings.synchronizationSettings.utxSynchronizerSettings, allChannels, transactions)
+    UtxPoolSynchronizer.start(utxStorage, settings.synchronizationSettings.utxSynchronizerSettings, allChannels, transactions, blocks)
```
phearnot: `blocks` are all blocks which come from peers, even (possibly) those which are invalid and hence will never be appended to the blockchain.

Karasiq: Replaced with newBlocks

PR #2083, opened 2019-03-25
UTX pool caching and address fixes

Karasiq:


src/main/scala/com/wavesplatform/account/Address.scala
```diff
@@ -17,44 +18,59 @@ sealed trait Address extends AddressOrAlias {
...
 
-  def fromPublicKey(publicKey: Array[Byte], chainId: Byte = scheme.chainId): Address = {
-    val withoutChecksum = ByteBuffer
-      .allocate(1 + 1 + HashLength)
-      .put(AddressVersion)
-      .put(chainId)
-      .put(crypto.secureHash(publicKey), 0, HashLength)
-      .array()
+  private[this] val publicKeyBytesCache: Cache[ByteStr, Address] = CacheBuilder.newBuilder()
+    .weakKeys()
```
phearnot: > <p><b>Warning:</b> when this method is used, the resulting cache will use identity (<code> == </code>) comparison to determine equality of keys.

This should probably be `Cache[PublicKey, Address]`: weak keys will work fine for interned instances.

src/main/scala/com/wavesplatform/account/KeyPair.scala
```diff
@@ -0,0 +1,36 @@
...
+
+final case class KeyPair(seed: ByteStr) {
+  private[this] lazy val pair = crypto.createKeyPair(seed)
+
+  lazy val privateKey: PrivateKey = PrivateKey(pair._1)
+  lazy val publicKey: PublicKey   = PublicKey(pair._2)
+}
+
+object KeyPair {
+  private[this] val interner = Interners.newWeakInterner[KeyPair]()
```
phearnot: Unlike public keys, key pairs are not instantiated so often, so we probably don't need an interner here.

src/main/scala/com/wavesplatform/account/KeyPair.scala
```diff
@@ -0,0 +1,36 @@
...
+import com.google.common.collect.Interners
+import com.wavesplatform.common.state.ByteStr
+import com.wavesplatform.common.utils.Base58
+import com.wavesplatform.crypto
+import com.wavesplatform.transaction.ValidationError.GenericError
+
+import scala.util.{Failure, Success}
+
+final case class KeyPair(seed: ByteStr) {
+  private[this] lazy val pair = crypto.createKeyPair(seed)
```
phearnot:
```suggestion
lazy val (PrivateKey(privateKey), PublicKey(publicKey)) = crypto.createKeyPair(seed)
```
`unapply` needs to be tweaked tho

PR #2088, opened 2019-03-26
SC-150 Script compilation result - 500 (internal server error)

mrkraft:


lang/shared/src/main/scala/com/wavesplatform/lang/utils/package.scala
```diff
@@ -35,11 +35,11 @@ package object utils {
       .map(d =>
         Try(d.value) match {
           case Success(v) =>
-            val cType = ContentType.parseString(v)
+            val cType = Try(ContentType.parseString(v))
```
ismagin: can't we change ContentType.parseString itself to return either and flatmap everythin else?

PR #2092, opened 2019-03-27
NODE-1500: UTX pool packUnconfirmed time metric

Karasiq:


src/main/scala/com/wavesplatform/utx/UtxPoolImpl.scala
```diff
@@ -59,8 +59,9 @@ class UtxPoolImpl(time: Time,
   private[this] object PoolMetrics {
     private[this] val sizeStats  = Kamon.rangeSampler("utx-pool-size", MeasurementUnit.none, Duration.of(500, ChronoUnit.MILLIS))
     private[this] val bytesStats = Kamon.rangeSampler("utx-pool-bytes", MeasurementUnit.information.bytes, Duration.of(500, ChronoUnit.MILLIS))
-    val processingTimeStats      = Kamon.histogram("utx-transaction-processing-time", MeasurementUnit.time.milliseconds)
-    val putRequestStats          = Kamon.counter("utx-pool-put-if-new")
+    val putTimeStats                    = Kamon.histogram("utx-transaction-processing-time", MeasurementUnit.time.milliseconds)
```
phearnot: These should be timers, which is a wrapper around a histogram. It's also a fortunate opportunity to rename all metrics the way we did in matcher (e.g. `node.utx.put-if-new` and `node.utx.pack-unconfirmed`)

src/main/scala/com/wavesplatform/utx/UtxPoolImpl.scala
```diff
@@ -152,31 +153,34 @@ class UtxPoolImpl(time: Time,
...
-          if (updatedRest.isOverfilled) {
-            (invalid, valid, diff, currRest, isEmpty)
-          } else {
-            differ(updatedBlockchain, tx) match {
-              case Right(newDiff) =>
-                (invalid, tx +: valid, Monoid.combine(diff, newDiff), updatedRest, currRest.isEmpty)
-              case Left(_) =>
-                (tx.id() +: invalid, valid, diff, currRest, isEmpty)
+
+    Instrumented.measure(PoolMetrics.packTimeStats) {
```
phearnot: Once `packTimeStats` is a timer, you can use `TimerExt.measure`

src/main/scala/com/wavesplatform/metrics/Instrumented.scala
```diff
@@ -1,78 +1,94 @@
 package com.wavesplatform.metrics
 
 import com.wavesplatform.utils.ScorexLogging
-import kamon.metric.{Counter, Histogram}
-
-trait Instrumented {
-  self: ScorexLogging =>
+import kamon.metric.{Counter, Histogram, MeasurementUnit}
 
+trait Instrumented { self: ScorexLogging =>
```
phearnot: I suggest removing all methods duplicating `Timer` functionality and deprecating all `measure*Log` methods.

Karasiq: Ok

PR #2095, opened 2019-03-27
NODE-1545: UTX packUnconfirmed quick stop on constraints overfill

Karasiq:


src/main/scala/com/wavesplatform/utx/UtxPoolImpl.scala
```diff
@@ -217,11 +219,21 @@ class UtxPoolImpl(time: Time,
...
+        diff <- TransactionDiffer(fs, lastBlockTimestamp, currentTime, height)(blockchain, transaction)
+      } yield diff
+    }
+
     def isValid(transaction: Transaction,
                 lastBlockTimestamp: Option[Long] = blockchain.lastBlockTimestamp,
                 currentTime: Long = time.correctedTime(),
                 height: Int = blockchain.height): Boolean = {
-      !isExpired(transaction) && TransactionDiffer(fs, lastBlockTimestamp, currentTime, height)(blockchain, transaction).isRight
+      validate(transaction, lastBlockTimestamp, currentTime, height).isRight
```
phearnot: The whole method doesn't seem to be used anywhere

PR #2099, opened 2019-03-27
Data 291 update notifications kafka

dvshur:


blockchain-updates/build.sbt
```diff
@@ -0,0 +1,9 @@
+name := "blockchain-updates"
+
+libraryDependencies += "org.apache.kafka" %% "kafka" % "2.1.0"
```
phearnot: Kafka version should probably be fixed in `Dependencies.scala` (see `dependencyOverrides`)

blockchain-updates/build.sbt
```diff
@@ -0,0 +1,9 @@
...
+
+inConfig(Compile)(Seq(
+  PB.targets += scalapb.gen(flatPackage = true) -> sourceManaged.value
+))
+
+enablePlugins(RunApplicationSettings, ExtensionPackaging)
```
phearnot: Please make sure all files end with a new line (IDEA can keep track of that)

blockchain-updates/src/main/protobuf/events.proto
```diff
@@ -0,0 +1,56 @@
...
+        SignedMicroBlock micro_block = 5;
+    }
+
+    enum UpdateType {
+        BLOCK = 0;
+        ROLLBACK = 1;
+        MICROBLOCK = 2;
+        MICROBLOCK_ROLLBACK = 3;
+    }
+    UpdateType type = 6;
```
phearnot: Is there a better way to keep `UpdateType` and `reason` in sync?

blockchain-updates/src/main/scala/com/wavesplatform/events/BlockchainUpdates.scala
```diff
@@ -0,0 +1,157 @@
...
+      case Some(kafkaHeight) =>
+        if (kafkaHeight <= (blockchainHeight + 1)) {
+          /*
+          Always rollback node on startup at least 1 block, since node loses ngState
+          and it's hard to tell whether the block in Kafka is full or not.
+
+          Rollback node to (kafkaHeight - 1) to be sure, since previous block is guaranteed to be solid.
+           */
+
+          val heightToRollbackTo = Math.max(kafkaHeight - 1, 1)
```
phearnot: You can rely on last block ID here, not height

blockchain-updates/src/main/scala/com/wavesplatform/events/BlockchainUpdates.scala
```diff
@@ -0,0 +1,157 @@
...
+
+          context.blockchain
+            .blockHeaderAndSize(heightToRollbackTo)
+            .toRight(new IllegalStateException(s"No block at height $heightToRollbackTo"))
```
phearnot: Use `Try[A]` instead of `Either[Throwable, A]`

node/src/main/scala/com/wavesplatform/Application.scala
```diff
@@ -143,6 +153,32 @@ class Application(val actorSystem: ActorSystem, val settings: WavesSettings, con
...
+    val extensionContext = new Context {
+      override def settings: WavesSettings                                  = app.settings
+      override def blockchain: Blockchain with BlockchainUpdater            = app.blockchainUpdater
+      override def time: Time                                               = app.time
+      override def wallet: Wallet                                           = app.wallet
+      override def utx: UtxPool                                             = utxStorage
+      override def broadcastTx(tx: Transaction): Unit                       = allChannels.broadcastTx(tx, None)
+      override def spendableBalanceChanged: Observable[(Address, Asset)]    = app.spendableBalanceChanged
+      override def actorSystem: ActorSystem                                 = app.actorSystem
+      override def blockchainUpdated: Option[Observable[BlockchainUpdated]] = maybeBlockchainUpdated
```
phearnot:
```suggestion
      override def blockchainUpdated: Observable[BlockchainUpdated] = ???
```
When update notifications are disabled, this could just be a dummy observable

node/src/main/scala/com/wavesplatform/ImporterSending.scala
```diff
@@ -0,0 +1,82 @@
...
+import com.wavesplatform.wallet.Wallet
+import monix.execution.Scheduler
+import monix.reactive.Observable
+import monix.reactive.subjects.ConcurrentSubject
+
+import scala.concurrent.duration._
+import scala.concurrent.Await
+import scala.util.{Failure, Success, Try}
+
+object ImporterSending extends ScorexLogging {
```
phearnot: The fact that there are two importers, one which publishes updates and another which isn't, hints at some design issues. Importer should not care about node configuration, it just feeds blocks to the node, no matter which plugins are enabled.

node/src/main/scala/com/wavesplatform/state/BlockchainUpdated.scala
```diff
@@ -0,0 +1,89 @@
...
+                                 height: Int,
+                                 microBlockStateUpdate: StateUpdated,
+                                 transactionsStateUpdates: Seq[StateUpdated])
+    extends BlockchainUpdated
+final case class RollbackCompleted(to: ByteStr, height: Int)           extends BlockchainUpdated
+final case class MicroBlockRollbackCompleted(to: ByteStr, height: Int) extends BlockchainUpdated
+
+object BlockchainUpdateNotifier {
+
+  private def stateUpdateFromDiff(blockchain: Blockchain, diff: Diff): StateUpdated = {
```
phearnot: There's already a method which does precisely that: see `Caches.append`. To avoid duplicating the logic, consider extracting common pieces into shared methods.

blockchain-updates/src/main/scala/com/wavesplatform/events/BlockchainUpdates.scala
dvshur: `blockchain.removeAfter(sigToRollback)` method returns Either. I used Either here to make it flatMap-compatible.

PR #2102, opened 2019-03-27
DEX-270 Exchange transaction base fee setting

DenisSerebryansky:


src/main/scala/com/wavesplatform/matcher/MatcherSettings.scala
```diff
@@ -64,14 +67,19 @@ object MatcherSettings {
...
-    val account                      = config.as[String]("account")
-    val bindAddress                  = config.as[String]("bind-address")
-    val port                         = config.as[Int]("port")
+    val enabled     = config.as[Boolean]("enable")
+    val account     = config.as[String]("account")
+    val bindAddress = config.as[String]("bind-address")
+    val port        = config.as[Int]("port")
+
+    val exchangeTxBaseFee = config.getValidatedByPredicate[Long]("exchange-tx-base-fee")(
+      predicate = _ >= OrderValidator.exchangeTransactionCreationFee,
```
vsuharnikov: @koloale Not sure, we need such validation, what do you think?

PR #2105, opened 2019-03-28
NODE-1297: Option to turn off storage of transactions by address

Karasiq:


src/main/scala/com/wavesplatform/database/Keys.scala
```diff
@@ -154,4 +156,6 @@ object Keys {
       writeTransactionHN
     )
 
+  val DisableTxsByAddress =
```
phearnot: This option should be configurable from `waves.conf`. There's no need to store this value in LevelDB.

Karasiq: It should be persisted to force user to rebuild the state, if he re-enables transactions by address.

src/main/scala/com/wavesplatform/metrics/Metrics.scala
```diff
@@ -25,6 +28,11 @@ object Metrics extends ScorexLogging {
 
   case class Settings(enable: Boolean, nodeId: Int, influxDb: InfluxDbSettings)
 
+  object Settings {
+    import com.wavesplatform.settings._ // implicit hypherCase
+    implicit val metricsSettingsValueReader: ValueReader[Settings] = arbitraryTypeValueReader
```
phearnot: Please remove all explicit usages of `arbitraryTypeValueReader`, just one import in `WavesSettings.scala` seems to do the trick.

Karasiq: But then a macro reader takes priority and ignores all custom writers from companion objects.

PR #2115, opened 2019-03-29
Merge version-0.16.x into master

phearnot:


src/main/scala/com/wavesplatform/protobuf/transaction/PBRecipients.scala
```diff
@@ -16,7 +16,7 @@ object PBRecipients {
   def toAddress(r: Recipient): Either[ValidationError, Address] = r.recipient match {
     case Recipient.Recipient.Address(bytes) =>
       val withHeader = Bytes.concat(Array(Address.AddressVersion, AddressScheme.current.chainId), bytes.toByteArray)
-      Right(Address.createUnsafe(Bytes.concat(withHeader, Address.calcCheckSum(withHeader))))
+      Address.fromBytes(Bytes.concat(withHeader, Address.calcCheckSum(withHeader)))
```
Karasiq: fromBytes incurs unnecessary checksum calculation, and not needed since its calculated here internally

PR #2128, opened 2019-04-02
SC-128-limit-base58-use

potan:


common/shared/src/main/scala/com/wavesplatform/common/state/ByteStr.scala
```diff
@@ -7,14 +7,25 @@ import scala.util.Try
 case class ByteStr(arr: Array[Byte]) {
   lazy val base58: String = Base58.encode(arr)
 
-  lazy val base64: String = "base64:" + Base64.encode(arr)
-
-  lazy val trim: String = base58.toString.take(7) + "..."
-
-  override lazy val toString: String = base58
+  lazy val base64_row: String = Base64.encode(arr)
```
ismagin: please use `camelCaseNaming`, also you probably intended to say `raw`, not `row`

common/shared/src/main/scala/com/wavesplatform/common/state/ByteStr.scala
```diff
@@ -7,14 +7,25 @@ import scala.util.Try
...
+  lazy val base64: String = "base64:" + base64_row
+
+  lazy val trim: String = (if (arr.length < 1024) {
+    base58.toString.take(7)
```
ismagin: this trim is unsafe because it generates full base58 first?

PR #2133, opened 2019-04-03
SC-11 additional tests

GIvanovWaves:
Tests with returned lists from user functions


lang/jvm/src/test/scala/com/wavesplatform/lang/compiler/ContractCompilerTest.scala
```diff
@@ -550,4 +550,38 @@ class ContractCompilerTest extends PropSpec with PropertyChecks with Matchers wi
     }
     compiler.ContractCompiler(ctx, expr) shouldBe 'right
   }
+
+  property("list for sets from user functions") {
```
ismagin: tthis test covers the same code as previous one, but less. it's ok to re-test this, but maybe we can reuse script by putting it to resource file and using?

PR #2134, opened 2019-04-03
SC-96 context version test

GIvanovWaves:
Tests for doubled inclusion of InvokeScriptTransaction CaseType in context


lang/jvm/src/test/scala/com/wavesplatform/lang/TypesTest.scala
```diff
@@ -0,0 +1,23 @@
+package com.wavesplatform.lang
+
+import com.wavesplatform.lang.v1.evaluator.ctx.impl.waves.Types
+import org.scalatest.{FreeSpec, Matchers}
+
+class TypesTest extends FreeSpec with Matchers {
```
ismagin: please use a more descriptive test name, like `ContextVersionTest`
```
"InvokeScriptTransaction" - {
"doesn't exist in lib version 2" in {
    val types = Types.buildWavesTypes(true, StdLibVersion.V2)
      types.count(c => c.name == "InvokeScriptTransaction") shouldEqual 0    
  }
"exists in lib version 3" in {
    val types = Types.buildWavesTypes(true, StdLibVersion.V3)
      types.count(c => c.name == "InvokeScriptTransaction") shouldEqual 1
   
  }
}
```

PR #2141, opened 2019-04-05
Merge version 0.16.3 into master

phearnot:


common/shared/src/main/scala/com/wavesplatform/common/state/ByteStr.scala
```diff
@@ -5,14 +5,25 @@ import com.wavesplatform.common.utils.{Base58, Base64}
...
+  override lazy val toString: String = if (arr.length < 1024) {
+    base58
+  } else {
+    base64
+  }
 
   def isEmpty: Boolean =
     arr.length == 0
 
+  def size: Int = arr.length
```
Karasiq: Is it used anywhere?

phearnot: [Decimpiler.scala:68](https://github.com/wavesplatform/Waves/pull/2141/files#diff-d27c15004f08b5dd11f2be22e8984380R68)

PR #2146, opened 2019-04-10
NODE-1550: More understandable error message

disfenoid:
[JIRA: https://wavesplatform.atlassian.net/browse/NODE-1550](https://wavesplatform.atlassian.net/browse/NODE-1550)


src/test/scala/com/wavesplatform/state/diffs/SponsorshipDiffTest.scala
```diff
@@ -208,7 +208,7 @@ class SponsorshipDiffTest extends PropSpec with PropertyChecks with Matchers wit
           blockDiffEi should produce("Asset was issued by other address")
         }
         assertDiffEi(setupBlocks, block(Seq(insufficientFee)), s) { blockDiffEi =>
-          blockDiffEi should produce("does not exceed minimal value of 100000000 WAVES: 99999999")
+          blockDiffEi should produce("(99999999 in WAVES) does not exceed minimal value of 100000000 WAVES")
```
msmolyakov: Please add tests about not enough fee in sponsored asset. Generally, the task was about it.

disfenoid: done

PR #2181, opened 2019-04-19
NODE-1572: Block appender metrics

Karasiq:


waves-devnet.conf
```diff
@@ -137,7 +137,7 @@ waves {
 
 # See application.conf for more information
 kamon {
-  enable = yes
+  enable = no
```
phearnot: Should this really be disabled?

PR #2185, opened 2019-04-22
NODE-1570: Move FunctionalitySettings and generating balance into the blockchain

Karasiq:


disfenoid:
I think preActivatedFeatures and all functions from FunctionalitySettings should be moved to Blockchain class

PR #2195, opened 2019-04-24
SC-190 Correct LazyVal trampolining

xrtm000:


lang/shared/src/main/scala/com/wavesplatform/lang/v1/evaluator/ctx/LazyVal.scala
```diff
@@ -13,13 +13,10 @@ sealed trait LazyVal {
...
 object LazyVal {
   private case class LazyValImpl(v: TrampolinedExecResult[EVALUATED], lc: LogCallback) extends LazyVal {
-    override val value: TrampolinedExecResult[EVALUATED] = EitherT(
-      Coeval.evalOnce(
-        v.value
-          .flatTap(a => Coeval.evalOnce(lc(a)))
-          .apply()
-      )
-    )
+    override val value: TrampolinedExecResult[EVALUATED] = {
```
ismagin: Hm, did anything change here?

PR #2230, opened 2019-05-13
NODE-1607: Fix account data duplicate keys

Karasiq:


node/src/main/scala/com/wavesplatform/database/LevelDBWriter.scala
```diff
@@ -129,12 +129,12 @@ class LevelDBWriter(writableDB: DB, spendableBalanceChanged: Observer[(Address,
...
 
   override def accountDataKeys(address: Address): Seq[String] = readOnly { db =>
-    for {
+    (for {
       addressId <- addressId(address).toVector
       keyChunkCount = db.get(Keys.dataKeyChunkCount(addressId))
       chunkNo <- Range(0, keyChunkCount)
       key     <- db.get(Keys.dataKeyChunk(addressId, chunkNo))
-    } yield key
+    } yield key).distinct
```
phearnot: No need for `distinct` here, because data keys are never added to the list twice (see `LevelDBWriter.scala:354`), although this could be rewritten with `iterateOver`.

PR #2244, opened 2019-05-16
Sc 108 drop support of complex arg types

mrkraft:


node/src/test/scala/com/wavesplatform/transaction/InvokeScriptTransactionSpecification.scala
```diff
@@ -153,6 +155,25 @@ class InvokeScriptTransactionSpecification extends PropSpec with PropertyChecks
...
+  property(s"can't call a func with non native(simple) args") {
+    import com.wavesplatform.common.state.diffs.ProduceError._
+    val pk = PublicKey.fromBase58String(publicKey).explicitGet()
+    InvokeScriptTransaction.create(
+      pk,
+      pk.toAddress,
+      Some(
+        Terms.FUNCTION_CALL(
+          FunctionHeader.User("foo"),
+          List(ARR(IndexedSeq(CONST_LONG(1L), CONST_LONG(2L))), CaseObj(CASETYPEREF("SHA256", List.empty), Map("tmpKey" -> CONST_LONG(42))))
```
msmolyakov: Should be separated into two tests: for array and for caseobj

PR #2252, opened 2019-05-17
SC-209 SC-111: Replace script runs with scripts complexity constraint

Karasiq:


Karasiq:
> Please add a unit test.

https://github.com/wavesplatform/Waves/pull/2252/files#diff-b2ef54149513a687feb7650adcbd94d4R335

PR #2255, opened 2019-05-17
Sc 33 36 add new fields to block info

mrkraft:


lang/shared/src/main/scala/com/wavesplatform/lang/v1/evaluator/ctx/impl/waves/Types.scala
```diff
@@ -26,7 +26,10 @@ object Types {
     List(
       "timestamp"           -> LONG,
       "height"              -> LONG,
-      "generationSignature" -> BYTESTR
+      "baseTarget"          -> LONG,
+      "generationSignature" -> BYTESTR,
+      "generator"           -> BYTESTR,
```
ismagin: generator should be Address I guess, because sender in transaction is

PR #2277, opened 2019-05-23
SC-215 DApp invocation results an internal server error.

mrkraft:


node/src/main/scala/com/wavesplatform/account/Address.scala
```diff
@@ -62,20 +62,27 @@ object Address extends ScorexLogging {
...
-                           (),
-                           s"Wrong addressBytes length: expected: ${Address.AddressLength}, actual: ${addressBytes.length}")
-          checkSum          = addressBytes.takeRight(ChecksumLength)
-          checkSumGenerated = calcCheckSum(addressBytes.dropRight(ChecksumLength))
-          _ <- Either.cond(java.util.Arrays.equals(checkSum, checkSumGenerated), (), s"Bad address checksum")
-        } yield createUnsafe(addressBytes)).left.map(err => InvalidAddress(err))
+        Either
+          .cond(
+            addressBytes.length == Address.AddressLength,
+            (),
```
xrtm000: A bit strange solution

PR #2278, opened 2019-05-23
NODE-1620: UTX packUnconfirmed limit by time

Karasiq:


node/src/test/scala/com/wavesplatform/utx/UtxPoolSpecification.scala
```diff
@@ -548,7 +548,7 @@ class UtxPoolSpecification
           val poolSizeBefore  = utxPool.size
 
           time.advance(maxAge * 2)
-          utxPool.packUnconfirmed(limitByNumber(100))
+          utxPool.packUnconfirmed(limitByNumber(100), FiniteDuration(5, "second"))
```
phearnot:
```suggestion
          utxPool.packUnconfirmed(limitByNumber(100), 5.sconds)
```

PR #2280, opened 2019-05-24
SC-213 Set all scripts' complexity to 4000 in V3

xrtm000:


lang/shared/src/main/scala/com/wavesplatform/lang/v1/ContractLimits.scala
```diff
@@ -1,10 +1,15 @@
 package com.wavesplatform.lang.v1
 
+import com.wavesplatform.lang.directives.values.{StdLibVersion, V1, V2, V3}
+
 object ContractLimits {
-  val MaxExprComplexity  = 20 * 100
+  val MaxExprComplexityByVersion: StdLibVersion => Int = {
+    case V1 | V2 => 2000
+    case V3      => 4000
```
ismagin: why not reuse existing `MaxContractComplexity` and whatever 2000 comes from?

PR #2284, opened 2019-05-24
NODE-1625: Flat fee feature

Karasiq:


phearnot:
Implemented in #2524.

PR #2286, opened 2019-05-27
Fix UTX metrics on microblock reject

Karasiq:


node/src/main/scala/com/wavesplatform/utx/UtxPoolImpl.scala
```diff
@@ -173,6 +156,23 @@ class UtxPoolImpl(time: Time, blockchain: Blockchain, spendableBalanceChanged: O
     Option(transactions.remove(txId))
       .foreach(afterRemove)
 
+  private[this] def addTransaction(tx: Transaction, verify: Boolean): TracedResult[ValidationError, Boolean] = {
+    if (!transactions.containsKey(tx.id())) {
+      val diff = TransactionDiffer(blockchain.lastBlockTimestamp, time.correctedTime(), blockchain.height, verify)(blockchain, tx)
+      diff.resultE.foreach {
+        PoolMetrics.addTransaction(tx)
```
phearnot: Pool metrics should be incremented even when diff is `Left`, but `verify` is `True`.

node/src/main/scala/com/wavesplatform/utx/UtxPoolImpl.scala
```diff
@@ -173,6 +156,23 @@ class UtxPoolImpl(time: Time, blockchain: Blockchain, spendableBalanceChanged: O
...
+        pessimisticPortfolios.add(tx.id(), _)
+      }
+      val isNew = diff.map(_ => true)
```
phearnot: Seems like transactions added with `verify == False` can never be new?

PR #2287, opened 2019-05-27
SC-220 Duplicating func arguments

xrtm000:


lang/shared/src/main/scala/com/wavesplatform/lang/v1/parser/Expressions.scala
```diff
@@ -38,7 +38,15 @@ object Expressions {
 
   sealed trait PART[+T] extends Positioned
   object PART {
-    case class VALID[T](position: Pos, v: T)           extends PART[T]
+    case class VALID[T](position: Pos, v: T)           extends PART[T] {
```
ismagin: why not just compare `v : T` in compiler?

PR #2289, opened 2019-05-27
SC-219 Asset info issuer

xrtm000:


lang/shared/src/main/scala/com/wavesplatform/lang/v1/evaluator/ctx/impl/waves/Types.scala
```diff
@@ -12,12 +12,13 @@ object Types {
...
     List(
-      "totalAmount" -> LONG,
-      "decimals"    -> LONG,
-      "issuer"      -> BYTESTR,
-      "reissuable"  -> BOOLEAN,
-      "scripted"    -> BOOLEAN,
-      "sponsored"   -> BOOLEAN
+      "totalAmount"     -> LONG,
+      "decimals"        -> LONG,
+      "issuer"          -> BYTESTR,
```
ismagin: issuer should be Address, not BYTESTR(as everywhere else)

xrtm000: +

PR #2292, opened 2019-05-27
NODE-1619: NFT API

disfenoid:
[JIRA: https://wavesplatform.atlassian.net/browse/NODE-1619](https://wavesplatform.atlassian.net/browse/NODE-1619)


node/src/main/scala/com/wavesplatform/database/LevelDBWriter.scala
```diff
@@ -172,14 +171,18 @@ class LevelDBWriter(writableDB: DB, spendableBalanceChanged: Observer[(Address,
     Map.empty
   )
 
-  private def loadPortfolio(db: ReadOnlyDB, addressId: BigInt) = loadLposPortfolio(db, addressId).copy(
+  private def loadPortfolio(db: ReadOnlyDB, addressId: BigInt, excludeNFT: Boolean) = loadLposPortfolio(db, addressId).copy(
     assets = (for {
-      asset <- db.get(Keys.assetList(addressId))
+      issuedAsset <- db.get(Keys.assetList(addressId))
+      asset <- transactionInfo(issuedAsset.id).collect {
```
phearnot: This will seriously slow down portfolio retrieval when NFT feature has not been activated yet. I suggest creating two distinct branches, one for when NFT is activated, and another for when it's not.

node/src/main/scala/com/wavesplatform/state/Blockchain.scala
```diff
@@ -50,6 +51,8 @@ trait Blockchain {
   def transactionInfo(id: ByteStr): Option[(Int, Transaction)]
   def transactionHeight(id: ByteStr): Option[Int]
 
+  def portfolioNFT(address: Address, from: Option[IssuedAsset]): CloseableIterator[IssueTransaction]
```
phearnot: `nftList` seems like a better name to me: what it returns isn't actually a portfolio.

node/src/main/scala/com/wavesplatform/state/package.scala
```diff
@@ -20,11 +22,47 @@ import scala.util.Try
 package object state {
   def safeSum(x: Long, y: Long): Long = Try(Math.addExact(x, y)).getOrElse(Long.MinValue)
 
+  def portfolioNFTFromDiff(b: Blockchain, d: Option[Diff])(address: Address, after: Option[IssuedAsset]): CloseableIterator[IssueTransaction] = {
```
phearnot: same here: `nftListFromDiff` seems more appropriate

node/src/main/scala/com/wavesplatform/database/LevelDBWriter.scala
```diff
@@ -172,14 +171,18 @@ class LevelDBWriter(writableDB: DB, spendableBalanceChanged: Observer[(Address,
...
+        case (_, itx: IssueTransaction) if !excludeNFT || !itx.isNFT => IssuedAsset(itx.assetId())
+      }
     } yield asset -> db.fromHistory(Keys.assetBalanceHistory(addressId, asset), Keys.assetBalance(addressId, asset)).getOrElse(0L)).toMap
   )
 
   override protected def loadPortfolio(address: Address): Portfolio = readOnly { db =>
-    addressId(address).fold(Portfolio.empty)(loadPortfolio(db, _))
+    val excludeNFT = activatedFeatures.contains(BlockchainFeatures.ReduceNFTFee.id)
```
phearnot:
```suggestion
    val excludeNFT = activatedFeatures.get(BlockchainFeatures.ReduceNFTFee.id).exists(_ <= height)
```
because `(featureId, height)` pairs are added to the map when voting finishes, but before activation height is reached. Please test this!

node/src/main/scala/com/wavesplatform/database/LevelDBWriter.scala
```diff
@@ -626,7 +629,36 @@ class LevelDBWriter(writableDB: DB, spendableBalanceChanged: Observer[(Address,
...
   }
 
-  override def addressTransactions(address: Address, types: Set[TransactionParser], fromId: Option[ByteStr]): CloseableIterator[(Height, Transaction)] = readStream { db =>
+  override def portfolioNFT(address: Address, from: Option[IssuedAsset]): CloseableIterator[IssueTransaction] = readStream { db =>
+    val assetIdStream: CloseableIterator[IssuedAsset] = db
+      .get(Keys.addressId(address))
+      .fold(CloseableIterator.empty[IssuedAsset]) { id =>
+        val addressId = AddressId @@ id
+
+        val iter = db.get(Keys.assetList(addressId)).iterator
```
phearnot: This will return asset ids from oldest to newest, but we need to have them in the reverse order.

PR #2297, opened 2019-05-28
SC-199 Rework default function

mrkraft:


node/src/main/scala/com/wavesplatform/state/diffs/InvokeScriptTransactionDiff.scala
```diff
@@ -88,17 +89,22 @@ object InvokeScriptTransactionDiff {
...
-            contract.callableFuncs.find(_.u.name == functionName).flatMap(_ => tx.funcCallOpt)
-          else
-            contract.defaultFuncOpt.map(defFunc => FUNCTION_CALL(FunctionHeader.User(defFunc.u.name), List.empty))
+        val contractFuncEi =
+          contract.callableFuncs
+            .find(_.u.name == functionName)
+            .map(_ => tx.funcCallOpt.getOrElse(FUNCTION_CALL(FunctionHeader.User(DEFAULT_FUNC_NAME), List.empty)))
+            .toRight(GenericError(s"No function '$functionName' at address ${tx.dAppAddressOrAlias}"))
+            .flatMap { funcCall =>
+              Either.cond(tx.funcCallOpt.nonEmpty || funcCall.args.isEmpty,
```
ismagin: isn't this a duplication of code in ContractEvaluator? it should handle this

node/src/test/scala/com/wavesplatform/state/diffs/InvokeScriptTransactionDiffTest.scala
```diff
@@ -1022,4 +1014,20 @@ class InvokeScriptTransactionDiffTest extends PropSpec with PropertyChecks with
         }
     }
   }
+
+  property("Default function should not accept arguments") {
```
ismagin: it is a regular callable function, so it can accept whatever amount of arguments it needs. it's only a matter of convention: when no func/args is specified, we try to invoke default()

PR #2299, opened 2019-05-28
NODE-1626: Key regex filtering

disfenoid:
[JIRA: https://wavesplatform.atlassian.net/browse/NODE-1626](https://wavesplatform.atlassian.net/browse/NODE-1626)

### DIFF
--- a/node/src/main/scala/com/wavesplatform/api/http/AddressApiRoute.scala
+++ b/node/src/main/scala/com/wavesplatform/api/http/AddressApiRoute.scala
@@ -268,7 +268,17 @@ case class AddressApiRoute(settings: RestAPISettings, wallet: Wallet, blockchain
 
   @Path("/data/{address}")
   @ApiOperation(value = "Complete Data", notes = "Read all data posted by an account", httpMethod = "GET")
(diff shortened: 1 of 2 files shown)


node/src/main/scala/com/wavesplatform/api/http/AddressApiRoute.scala
```diff
@@ -267,8 +269,19 @@ case class AddressApiRoute(settings: RestAPISettings, wallet: Wallet, blockchain
   @Path("/data/{address}")
   @ApiOperation(value = "Complete Data", notes = "Read all data posted by an account", httpMethod = "GET")
   @ApiImplicitParams(Array(new ApiImplicitParam(name = "address", value = "Address", required = true, dataType = "string", paramType = "path")))
-  def getData: Route = (path("data" / Segment) & get) { address =>
-    complete(accountData(address))
+  def getData: Route = (path("data" / Segment) & parameter('matches.?) & get) { (address, maybeRegex) =>
```
phearnot: Please add some documentation for `matches` param (make sure to mention encoding schemes, etc.)

PR #2306, opened 2019-05-29
NODE-1564: Importer/Exporter/Util command line tools refactoring

Karasiq:


node/src/main/scala/com/wavesplatform/utils/UtilApp.scala
```diff
@@ -38,11 +38,13 @@ object UtilApp {
   case class HashOptions(mode: String = "fast")
   case class SignTxOptions(signerAddress: String = "")
 
-  case class Command(mode: Command.Mode = Command.CompileScript,
+  case class Command(mode: Command.Mode = null,
+                     configFile: Option[String] = None,
                      inputData: Option[String] = None,
                      inputFile: Option[String] = None,
```
phearnot: One can specify _either_ literal input string or an input file name, so `Either[String, String]` would be more appropriate.

node/src/main/scala/com/wavesplatform/utils/UtilApp.scala
```diff
@@ -86,26 +89,40 @@ object UtilApp {
...
-        opt[String]('s', name = "data")
+        opt[String](name = "input-str")
+          .abbr("is")
           .text("Literal input data")
           .action((s, c) => c.copy(inputData = Some(s))),
-        opt[String]('i', "input-file")
-          .action((f, c) => c.copy(inputFile = Some(f)))
-          .text("Input file name")
+        opt[String]("input-file")
+          .abbr("if")
```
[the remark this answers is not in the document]
Karasiq: To be consistent with "is" (input str)

node/src/main/scala/com/wavesplatform/utils/UtilApp.scala
```diff
@@ -86,26 +89,40 @@ object UtilApp {
...
-          .action((f, c) => c.copy(outputFile = Some(f)))
-          .text("Output file name"),
-        opt[String]('f', "format")
-          .action((f, c) => c.copy(format = f))
+          .action((f, c) => c.copy(outputFile = Some(f).filter(s => s != "-" && s.nonEmpty)))
+          .text("Output file name (- for stdout)"),
+        opt[String]("in-format")
+          .abbr("fi")
+          .action((f, c) => c.copy(inFormat = f))
+          .text("Input data format (plain/base58/base64)")
```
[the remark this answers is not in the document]
Karasiq: Raw string or file bytes

node/src/main/scala/com/wavesplatform/settings/WavesSettings.scala
```diff
@@ -60,4 +66,28 @@ object WavesSettings extends CustomValueReaders {
       rootConfig
     )
   }
+
+  def loadRootConfig(external: Option[File] = None): WavesSettings = {
```
phearnot: This is a bad design decision: reading and parsing settings has nothing to do with application component initialization.

node/src/main/scala/com/wavesplatform/settings/WavesSettings.scala
```diff
@@ -60,4 +66,28 @@ object WavesSettings extends CustomValueReaders {
...
+    AddressScheme.current = new AddressScheme {
+      override val chainId: Byte = settings.blockchainSettings.addressSchemeCharacter.toByte
+    }
+
+    if (config.getBoolean("kamon.enable")) {
+      Kamon.addReporter(new InfluxDBReporter())
+      SystemMetrics.startCollecting()
+    }
+
+    // DO NOT LOG BEFORE THIS LINE, THIS PROPERTY IS USED IN logback.xml
```
phearnot: This comment was added for a reason. The changes you propose have broken initialization order, and now logback creates `waves.directory_IS_UNDEFINED` directory, even though it's actually defined in the config.

node/src/main/scala/com/wavesplatform/utils/UtilApp.scala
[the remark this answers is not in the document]
phearnot: Well maybe it should be consistent with `import` command which also reads data from a file?

node/src/main/scala/com/wavesplatform/utils/UtilApp.scala
```diff
@@ -251,14 +258,14 @@ object UtilApp {
   private[this] object IO {
     def readInput(c: Command): Array[Byte] = {
       val inputStream = c.inputData match {
-        case Some(value) =>
-          new ByteArrayInputStream(value.getBytes)
+        case Input.StdIn | Input.File("-") =>
```
phearnot: Dedicated types for input sources is a good idea, but why are bot `File("-")`and `StdIn` used for stdin?

PR #2318, opened 2019-05-31
SC Remove unreachable code

potan:


lang/shared/src/main/scala/com/wavesplatform/lang/v1/compiler/TypeInferrer.scala
```diff
@@ -87,13 +87,13 @@ object TypeInferrer {
     case (NOTHING, t2)        => t2
     case (t1, t2) if t1 == t2 => t1
 
-    case (r @ LIST(it1), a @ LIST(it2)) =>
+    case (LIST(it1), LIST(it2)) => LIST(findCommonType(it1, it2))
```
ismagin: are you sure this works as exprected? the added code is effectively dead

PR #2336, opened 2019-06-05
SC-362 Add String and ByteVector size limits via smart constructor

xrtm000:


lang/shared/src/main/scala/com/wavesplatform/lang/v1/compiler/Terms.scala
```diff
@@ -65,9 +67,33 @@ object Terms {
...
+      )
+  }
+
+  case class CONST_STRING private (s: String) extends EVALUATED {
+    override def toString: String = s
+  }
+  object CONST_STRING {
+    def apply(s: String): Either[ExecutionError, CONST_STRING] =
+      Either.cond(
+        s.length * 2 <= DATA_TX_BYTES_MAX,
```
ismagin: are you sure length string should be capped by the value, not string's bytes length?

lang/shared/src/main/scala/com/wavesplatform/lang/v1/compiler/Terms.scala
```diff
@@ -65,9 +67,33 @@ object Terms {
...
+
+  case class CONST_BYTESTR private (bs: ByteStr) extends EVALUATED {
+    override def toString: String = bs.toString
+  }
+  object CONST_BYTESTR {
+    def apply(bs: ByteStr): Either[ExecutionError, CONST_BYTESTR] =
+      Either.cond(
+        bs.size <= DATA_TX_BYTES_MAX,
+        new CONST_BYTESTR(bs),
+        "ByteStr is too large"
```
ismagin: maybe more detailed message here?

lang/shared/src/main/scala/com/wavesplatform/lang/v1/compiler/Terms.scala
```diff
@@ -65,9 +67,33 @@ object Terms {
...
+
+  case class CONST_STRING private (s: String) extends EVALUATED {
+    override def toString: String = s
+  }
+  object CONST_STRING {
+    def apply(s: String): Either[ExecutionError, CONST_STRING] =
+      Either.cond(
+        s.length * 2 <= DATA_TX_BYTES_MAX,
+        new CONST_STRING(s),
+        "String is too large"
```
ismagin: maybe more detailed message here?

PR #2345, opened 2019-06-07
Node 333 network dependent directory

Karasiq:


node/src/main/scala/com/wavesplatform/settings/package.scala
```diff
@@ -44,35 +44,53 @@ package object settings {
...
+      maybeXdgDir getOrElse defaultDir
+    }
 
-  // No actual interpolation here, `s` to suppress warnings
-  def osxDefaultDirectory: String =
-    s"$${user.home}/Library/Application Support/waves"
+    def withNetwork(config: Config): String = {
+      val bc = config.getString("waves.blockchain.type")
+      val suffix =
+        if (bc == "CUSTOM") s"custom${config.getString("waves.blockchain.custom.address-scheme-character")}"
```
phearnot: This approach has several flaws: some filesystems are case-insensitive, some forbid special charaters in directory names. A safer way to support arbitrary characters would be to append decimal or hexadecimal value as a suffix (e.g. `waves-custom-0xff`)

PR #2359, opened 2019-06-11
DEX-318 Order versions in matcher settings (0.17.x)

DenisSerebryansky:


dex/src/main/scala/com/wavesplatform/matcher/model/OrderValidator.scala
```diff
@@ -147,13 +156,15 @@ object OrderValidator {
       else verifyAssetScript(matcherFeeAsset)
     }
 
+    lazy val isOrderVersionSupported = isOderVersionSupported(blockchain) _
+
     lazy val validateActivatedOrderFeatures =
       lift(order)
         .ensure(MatcherError.OrderVersionUnsupported(order.version, BlockchainFeatures.SmartAccountTrading)) {
-          _.version == 1 || blockchain.isFeatureActivated(BlockchainFeatures.SmartAccountTrading, blockchain.height)
+          _.version == 1 || isOrderVersionSupported(order.version)
```
vsuharnikov: How about:
```scala
private def checkOrderVersion(blockchain: Blockchain)(version: Byte): Result[Unit] =
  version match {
    case 1 => success
    case 2 =>
      if (blockchain.isFeatureActivated(BlockchainFeatures.SmartAccountTrading, blockchain.height)) success
      else Left(MatcherError.OrderVersionUnsupported(version, BlockchainFeatures.SmartAccountTrading))
    case 3 =>
      if (blockchain.isFeatureActivated(BlockchainFeatures.OrderV3, blockchain.height)) success
      else Left(MatcherError.OrderVersionUnsupported(version, BlockchainFeatures.OrderV3))
    case _ => Left(MatcherError.UnknownOrderVersion(version))
  }
```
? With this implementation you doesn't need to write multiple `ensure`.

DenisSerebryansky: Fixed

PR #2363, opened 2019-06-13
NODE-1592 fix for node-generator fees

msmolyakov:


phearnot:
Rushed by @msmolyakov

PR #2370, opened 2019-06-14
NODE-1772: RejectedExecution fix

disfenoid:
[JIRA: https://jira.wavesplatform.com/browse/NODE-1772](https://jira.wavesplatform.com/browse/NODE-1772)

### DIFF
--- a/node/src/main/scala/com/wavesplatform/Application.scala
+++ b/node/src/main/scala/com/wavesplatform/Application.scala
@@ -32,6 +32,7 @@ import com.wavesplatform.settings._
 import com.wavesplatform.state.Blockchain
 import com.wavesplatform.state.appender.{BlockAppender, ExtensionAppender, MicroblockAppender}
 import com.wavesplatform.transaction.{Asset, Transaction}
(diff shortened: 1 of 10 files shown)


node/src/main/scala/com/wavesplatform/Application.scala
```diff
@@ -42,14 +43,13 @@ import kamon.Kamon
...
-import monix.execution.Scheduler
-import monix.execution.Scheduler._
 import monix.execution.schedulers.SchedulerService
+import monix.execution.{Scheduler, UncaughtExceptionReporter}
 import monix.reactive.Observable
 import monix.reactive.subjects.ConcurrentSubject
 import org.influxdb.dto.Point
 import org.slf4j.LoggerFactory
-
+import scala.concurrent.ExecutionContext.Implicits.global
```
Karasiq: Shouldn't `global` be wrapped too?

phearnot: Seems like there's no need to: we never explicitly shut down global execution context.

PR #2371, opened 2019-06-14
DEX-274 New inaccessible functions and fields

vsuharnikov:


dex/src/main/scala/com/wavesplatform/matcher/smart/MatcherContext.scala
```diff
@@ -64,33 +60,31 @@ object MatcherContext {
...
-      wavesBalanceF
-    )
-
-    val matcherContext = CTX(matcherTypes, matcherVars, functions).evaluationContext
+    val nativeFunctions = Array(
+      DATA_LONG_FROM_STATE    -> "getInteger",
+      DATA_BOOLEAN_FROM_STATE -> "getBoolean",
+      DATA_BYTES_FROM_STATE   -> "getBinary",
+      DATA_STRING_FROM_STATE  -> "getString",
+      GETTRANSACTIONBYID      -> "txById",
```
msmolyakov: correct func name is `transactionById`

dex/src/main/scala/com/wavesplatform/matcher/smart/MatcherContext.scala
```diff
@@ -64,33 +60,31 @@ object MatcherContext {
...
+      BLOCKINFOBYHEIGHT       -> "blockInfoByHeight"
+    ).map(Function.tupled(inaccessibleFunction))
+
+    val userFunctions = Array(
+      "getIntegerValue",
+      "getBooleanValue",
+      "getBinaryValue",
+      "getStringValue",
+      "addressFromPublicKey",
+      "addressFromString",
```
msmolyakov: don't ignore `addressFromPublicKey`. May be `addressFromString` too

PR #2379, opened 2019-06-17
SC-385 decompile match

potan:


lang/jvm/src/test/scala/com/wavesplatform/lang/compiler/DecompilerTest.scala
```diff
@@ -464,13 +465,37 @@ class DecompilerTest extends PropSpec with PropertyChecks with Matchers {
...
-        |        then {
-        |            let tx = $match0
-        |            true
-        |            }
-        |        else false""".stripMargin
+        |    case tx: BurnTransaction => 
+        |        true
+        |    case _ => 
+        |        false
+        |}""".stripMargin
```
msmolyakov: - remove 5 spaces between `match     tx`
- add branch `case _: IssueTransaction|ReissueTransaction`

PR #2412, opened 2019-06-27
Show importer time when interrupted

Karasiq:


node/src/main/scala/com/wavesplatform/Importer.scala
```diff
@@ -59,12 +59,18 @@ object Importer extends ScorexLogging {
             val bis           = new BufferedInputStream(inputStream)
             var quit          = false
             val lenBytes      = new Array[Byte](Ints.BYTES)
-            val start         = System.currentTimeMillis()
+            val start         = System.nanoTime()
```
phearnot: Is there really a reason to prefer nanos over millis in this particular case?

node/src/main/scala/com/wavesplatform/Importer.scala
```diff
@@ -59,12 +59,18 @@ object Importer extends ScorexLogging {
...
             var blocksToSkip  = blockchainUpdater.height - 1
             val blocksToApply = importHeight - blockchainUpdater.height + 1
 
-            println(s"Skipping $blocksToSkip block(s)")
+            log.info(s"Skipping $blocksToSkip block(s)")
+
+            sys.addShutdownHook {
+              import scala.concurrent.duration._
+              val millis = (System.nanoTime() - start).nanos.toMillis
+              log.info(s"Imported $counter block(s) from ${blockchainUpdater.height} to $importHeight in ${humanReadableDuration(millis)}")
```
phearnot: Since `log.info` accepts a by-name parameter, upon shutdown `blockchainUpdater.height` will be the final blockchain height, not the height it had before it all started. Also `importHeight` may differ from actual blockchain height.

PR #2429, opened 2019-07-05
NODE-1828: Closeable iterator fixes

Karasiq:


node/src/main/scala/com/wavesplatform/database/LevelDBWriter.scala
```diff
@@ -791,20 +791,25 @@ class LevelDBWriter(writableDB: DB, spendableBalanceChanged: Observer[(Address,
...
-          .collect { case lt: LeaseTransaction => lt }
-      } else None
+          db.get(Keys.transactionAt(height, txNum))
+        } else None
+      }
     }
+
+    iterator
+      .collect { case lt: LeaseTransaction if pf.isDefinedAt(lt) => pf(lt) }
+      .closeAfter(_.toVector)
```
phearnot: Iterator is closed in the same scope it was created => no need for closeable iterator

node/src/main/scala/com/wavesplatform/settings/BlockchainSettings.scala
```diff
@@ -58,7 +58,7 @@ object FunctionalitySettings {
     minimalGeneratingBalanceAfter = 1479168000000L,
     allowTransactionsFromFutureUntil = 1479168000000L,
     allowUnissuedAssetsUntil = 1479416400000L,
-    allowInvalidReissueInSameBlockUntilTimestamp = 1492768800000L,
+    allowInvalidReissueInSameBlockUntilTimestamp = 1530161445559L, // 1492768800000L
```
Karasiq: Maybe should be fixed for testnet too

PR #2432, opened 2019-07-08
NODE-1800: Merkle proofs for balances

alxndrmkd:
[JIRA: https://jira.wavesplatform.com/browse/NODE-1800](https://jira.wavesplatform.com/browse/NODE-1800)


node/src/main/scala/com/wavesplatform/state/Blockchain.scala
```diff
@@ -108,6 +108,8 @@ trait Blockchain {
   // the following methods are used exclusively by patches
   def allActiveLeases: CloseableIterator[LeaseTransaction]
 
+  def proofForBalanceOnHeight(address: Address, height: Height): Option[ProvenBalance]
```
phearnot: To make naming consistent with block (header) at height and other height-related methods:
```suggestion
  def balanceProof(address: Address, height: Height): Option[ProvenBalance]
```

node/src/main/scala/com/wavesplatform/state/BlockchainUpdaterImpl.scala
```diff
@@ -721,6 +721,8 @@ class BlockchainUpdaterImpl(blockchain: LevelDBWriter,
     }
   }
 
+  override def proofForBalanceOnHeight(address: Address, height: Height): Option[ProvenBalance] = blockchain.proofForBalanceOnHeight(address, height)
```
phearnot: This will not work for balance proofs at top height, is this by design?

node/src/main/scala/com/wavesplatform/state/ProvenBalance.scala
```diff
@@ -0,0 +1,21 @@
+package com.wavesplatform.state
+
+import com.wavesplatform.account.Address
+import com.wavesplatform.crypto.Merkle.BalanceProof
+import play.api.libs.json._
+
+case class ProvenBalance(address: Address, proofForMiningBalance: BalanceProof, proofForRegularBalance: BalanceProof, balanceInfo: BalanceInfo)
+
+object ProvenBalance {
+  implicit val format: Writes[ProvenBalance] = Writes { provenBalance =>
```
phearnot: Why not just use the `Json.writes[ProvenBalance]` macro?

node/src/main/scala/com/wavesplatform/api/common/CommonAccountApi.scala
```diff
@@ -78,6 +81,12 @@ class CommonAccountApi(blockchain: Blockchain) {
           (height, leaseTransaction)
       })
   }
+
+  def balanceProof(address: Address, height: Int): Either[ValidationError, ProvenBalance] = {
```
phearnot: I like this method name, it's quite concise!

node/src/main/scala/com/wavesplatform/database/package.scala
```diff
@@ -357,6 +357,33 @@ package object database {
     ndo.toByteArray
   }
 
+  def readBalanceInfo(bs: Array[Byte]): Map[Address, BalanceInfo] = {
+    val ndi        = newDataInput(bs)
+    val entryCount = ndi.readInt()
+
+    (0 until entryCount).map { _ =>
+      val addr          = ndi.readByteStr(Address.AddressLength)
```
phearnot: Please run explorer tool and check how much space is taken by addresses in these list. There's likely no need to store the whole address, address ID will do.

phearnot (changes requested):

node/src/main/scala/com/wavesplatform/state/BlockchainUpdaterImpl.scala
disfenoid: yes, because proof on current height doesn't make sense

widyhafiz03:
No problem

PR #2434, opened 2019-07-08
NODE-1775: Key regex filtering: tests

frekensnork:


node/src/test/scala/com/wavesplatform/http/AddressRouteSpec.scala
```diff
@@ -218,24 +228,42 @@ class AddressRouteSpec
...
+      try {
+        Get(routePath(s"""/data/${allAddresses(1)}?matches=$regex""")) ~> route ~> check {
+          responseAs[JsArray].value.map { json =>
+            ((json \ "key").as[String], (json \ "value").as[String])
+          }
+        }
+        fail("RegexCompiler didn't throw expected error")
+      } catch {
+        case err: Throwable => if (!err.getMessage.contains("Cannot compile regex")) {
+          throw err
```
msmolyakov: if an exception is expected here, then it is probably better to use `assertThrows` of ScalaTest

frekensnork: as i understand assertThrows need concrete type of exception but here need to check only exception message
i guess better solution is to check message in response right after request without throwing added exceptions

PR #2438, opened 2019-07-10
NODE-1802: Blockheader in RIDE

disfenoid:
[JIRA: https://jira.wavesplatform.com/browse/NODE-1802](https://jira.wavesplatform.com/browse/NODE-1802)


lang/shared/src/main/scala/com/wavesplatform/lang/v1/evaluator/ctx/impl/waves/Bindings.scala
```diff
@@ -287,6 +291,23 @@ object Bindings {
...
+      Map[String, EVALUATED](
+        "timestamp"           -> header.timestamp,
+        "version"             -> header.version,
+        "reference"           -> header.reference,
+        "generator"           -> header.generator,
+        "generatorPublicKey"  -> header.generatorPublicKey,
+        "signature"           -> header.signature,
+        "baseTarget"          -> header.baseTarget,
+        "generationSignature" -> header.generationSignature,
+        "transactionCount"    -> header.transactionCount,
```
Karasiq: Is transactionCount field really needed?

node/src/main/scala/com/wavesplatform/block/Block.scala
```diff
@@ -47,6 +48,66 @@ class BlockHeader(val timestamp: Long,
...
+
+    ndo.writeInt(bh.featureVotes.size)
+    bh.featureVotes.foreach(s => ndo.writeShort(s))
+    ndo.write(bh.signerData.generator)
+    ndo.write(bh.signerData.signature)
+
+    ndo.toByteArray
+  }
+
+  def readHeaderOnly(bytes: Array[Byte]): BlockHeader = {
```
Karasiq: Why not reuse `com.wavesplatform.block.BlockHeader#parseBytes`?

disfenoid: Because it parses not only blockheader bytes but also transactions

lang/shared/src/main/scala/com/wavesplatform/lang/v1/evaluator/ctx/impl/waves/Bindings.scala
disfenoid: We can remove it later (now it's not clear do we need this filed or not)

node/src/main/scala/com/wavesplatform/transaction/smart/WavesEnvironment.scala
```diff
@@ -124,4 +126,22 @@ class WavesEnvironment(nByte: Byte, in: Coeval[WavesEnvironment.In], h: Coeval[I
...
+        header.timestamp,
+        header.version,
+        header.reference,
+        header.signerData.generator.toAddress.bytes,
+        header.signerData.generator.bytes,
+        header.signerData.signature,
+        header.consensusData.baseTarget,
+        header.consensusData.generationSignature,
+        header.transactionCount,
+        header.featureVotes.map(_.toLong).toSeq
```
Karasiq: `Set` ordering is not defined, so `.toSeq` can produce different results on same items

Karasiq:
```suggestion
        header.featureVotes.map(_.toLong).toVector.sorted
```

PR #2441, opened 2019-07-12
NODE-1836: NFT balance test

DrBlast:
additional tests for nft balance
https://jira.wavesplatform.com/browse/NODE-1836


disfenoid:
[Fixed](https://github.com/wavesplatform/Waves/blob/cb67fb1883f6661766033bedfdca3af9ed2ef990/node-it/src/test/scala/com/wavesplatform/it/sync/NFTBalanceSuite.scala#L124) in #2449

PR #2451, opened 2019-07-17
SC-406 Stateful REPL

xrtm000:


lang/shared/src/main/scala/com/wavesplatform/lang/v1/Repl.scala
```diff
@@ -0,0 +1,37 @@
...
+import com.wavesplatform.lang.v1.evaluator.ctx.impl.{CryptoContext, PureContext}
+import com.wavesplatform.lang.v1.parser.Parser
+import monix.execution.atomic.Atomic
+import cats.implicits._
+import com.wavesplatform.lang.v1.compiler.Terms.EVALUATED
+import com.wavesplatform.lang.v1.compiler.Types.{FINAL, UNIT}
+
+case class Repl(ver: StdLibVersion = V3) {
+  private val Global: BaseGlobal = com.wavesplatform.lang.Global
+  private val ctx         = CryptoContext.build(Global, ver) |+| PureContext.build(Global, ver)
```
ismagin: please add wavescontext too, mock Environment to throw exceptions ("function not available in repl context")

xrtm000: +

PR #2461, opened 2019-07-24
SC-412 New estimator

xrtm000:


node/src/main/scala/com/wavesplatform/utils/UtilApp.scala
```diff
@@ -202,12 +206,19 @@ object UtilApp {
     type ActionResult = Either[String, Array[Byte]]
 
     //noinspection ScalaDeprecation
-    def doCompile(c: Command, str: Array[Byte]): ActionResult =
-      ScriptCompiler(new String(str), c.compileOptions.assetScript)
+    def doCompile(settings: WavesSettings)(c: Command, str: Array[Byte]): ActionResult = {
+      val estimator =
+        if (settings.featuresSettings.supported.contains(ReduceNFTFee.id))
```
phearnot: Looks like this is the wrong feature ID

node/src/main/scala/com/wavesplatform/features/EstimatorProvider.scala
```diff
@@ -0,0 +1,14 @@
...
+import com.wavesplatform.state.Blockchain
+import BlockchainFeatures.ReduceNFTFee
+import FeatureProvider._
+import com.wavesplatform.lang.v1.estimator.{ScriptEstimator, ScriptEstimatorV1}
+import com.wavesplatform.lang.v2.estimator.ScriptEstimatorV2
+
+object EstimatorProvider {
+  implicit class EstimatorBlockchainExt(b: Blockchain) {
+    def estimator(): ScriptEstimator =
+      if (b.isFeatureActivated(ReduceNFTFee)) ScriptEstimatorV2 else ScriptEstimatorV1
```
phearnot: Once #2524 is merged, this should be changed to `ChangeMinimalFees`

PR #2464, opened 2019-07-24
NODE-1832: Peer connections on startup

chepiov:


node/src/main/scala/com/wavesplatform/network/PeerDatabase.scala
```diff
@@ -49,6 +51,8 @@ object PeerDatabase extends ScorexLogging {
 
     override def randomPeer(excluded: Set[InetSocketAddress]): Option[InetSocketAddress] = None
```
Karasiq: Better remove `randomPeer` method completely and use the `randomPeers(1, ...)` as a replacement

chepiov: Done

node/src/main/scala/com/wavesplatform/network/NetworkServer.scala
```diff
@@ -202,37 +206,47 @@ object NetworkServer extends ScorexLogging {
         }
       )
 
-    val connectTask = workerGroup.scheduleWithFixedDelay(1.second, 5.seconds) {
-      import scala.collection.JavaConverters._
-      val outgoing = outgoingChannels.keySet.iterator().asScala.toVector
+    def scheduleConnectTask(): Unit = {
```
phearnot: Please avoid `return` at all costs!
```suggestion
    def scheduleConnectTask(): Unit = if (!shutdownInitiated) {
```

node/src/main/scala/com/wavesplatform/network/HandshakeHandler.scala
```diff
@@ -31,12 +31,12 @@ class HandshakeDecoder(peerDatabase: PeerDatabase) extends ReplayingDecoder[Void
 
 case object HandshakeTimeoutExpired
 
-class HandshakeTimeoutHandler(handshakeTimeout: FiniteDuration) extends ChannelInboundHandlerAdapter with ScorexLogging {
+class HandshakeTimeoutHandler(handshakeTimeout: => FiniteDuration) extends ChannelInboundHandlerAdapter with ScorexLogging {
```
Karasiq: Better do `handshakeTimeout` an explicit function 
```suggestion
class HandshakeTimeoutHandler(handshakeTimeout: () => FiniteDuration) extends ChannelInboundHandlerAdapter with ScorexLogging {
```

phearnot: Actually, looks like there's no need to use a by-name parameter at all here.

PR #2475, opened 2019-07-31
NODE-1859: gRPC endpoint for alias resolution

Karasiq:


grpc-server/src/main/scala/com/wavesplatform/api/grpc/AccountsApiGrpcImpl.scala
```diff
@@ -65,4 +68,18 @@ class AccountsApiGrpcImpl(blockchain: Blockchain)(implicit sc: Scheduler)
...
+        alias   <- Alias.create(request.value)
+        address <- blockchain.resolveAlias(alias)
+      } yield BytesValue(address.bytes)
+
+      Future.fromTry {
+        addressEither
+          .leftMap(_.toException)
+          .toTry
+      }
+    }.flatten
```
phearnot: 1. Please remove `ValidationErrorException`
1. Replace `Future.fromTry` with `addressEither.explicitGet()`

grpc-server/src/main/protobuf/accounts_api.proto
```diff
@@ -5,12 +5,14 @@ option csharp_namespace = "Waves.Grpc";
...
 import "transactions_api.proto";
 import "transaction.proto";
+import "google/protobuf/wrappers.proto";
 
 service AccountsApi {
     rpc GetBalances (BalancesRequest) returns (stream BalanceResponse);
     rpc GetScript (AccountRequest) returns (ScriptData);
     rpc GetActiveLeases (AccountRequest) returns (stream TransactionResponse);
     rpc GetDataEntries (DataRequest) returns (stream DataEntryResponse);
+    rpc ResolveAlias (google.protobuf.StringValue) returns (google.protobuf.BytesValue);
```
phearnot: How about returning the whole `CreateAliasTransaction`?

grpc-server/src/main/scala/com/wavesplatform/api/grpc/AccountsApiGrpcImpl.scala
Karasiq: `ValidationErrorException` required for a gRPC errors logic, because plain Exceptions from `explicitGet()` omits the corresponding ValidationError, so we cant embed it conveniently into a gRPC status response.

PR #2488, opened 2019-08-05
NODE-1806: Test for transferById optimization

Karasiq:


node/src/test/scala/com/wavesplatform/database/LevelDBWriterSpec.scala
```diff
@@ -408,5 +409,18 @@ class LevelDBWriterSpec extends FreeSpec with Matchers with TransactionGen with
       }
     }
 
+    "dont parse irrelevant transactions in transferById" in {
+      baseTest(time => preconditions(time.correctedTime())) { (writer, _) =>
```
phearnot: There's actually no need to use `baseTest` here. You could just use `db.put` and then instantiate `LevelDBWriter` manually. This will also allow to keep `readWrite` method private.
It may also be a good idea to use a generator to test random transaction IDs.

node/src/main/scala/com/wavesplatform/database/LevelDBWriter.scala
```diff
@@ -639,12 +639,11 @@ class LevelDBWriter(writableDB: DB, spendableBalanceChanged: Observer[(Address,
         }.getOrElse(false)
 
         if (isTransfer)
-          TransactionParsers
-            .parseBytes(txBytes)
-            .collect {
-              case ttx: TransferTransaction => (height, ttx)
-            }
-            .toOption
+          TransactionParsers.parseBytes(txBytes) match {
```
phearnot:
```suggestion
          Some(height -> TransactionParsers.parseBytes(txBytes).get)
```

PR #2497, opened 2019-08-08
NODE-1868: Blockchain refactoring

Karasiq:


node/src/main/scala/com/wavesplatform/database/LevelDBWriter.scala
```diff
@@ -737,25 +733,18 @@ class LevelDBWriter(private[database] val writableDB: DB,
...
+    db.iterateOver(Keys.LeaseStatusPrefix) { kv =>
+      val height = Height(Ints.fromByteArray(kv.getKey.slice(2, 6)))
+      val txId   = TransactionId(ByteStr(kv.getKey.drop(2 + 4)))
+      if (kv.getValue.head == 1) results += txId -> height
+      else results -= txId
     }
-    results.result()
+
+    for {
+      (txId, _)                 <- results.toVector.sortBy(_._2)
```
phearnot: Do we really need to sort these?

phearnot (changes requested):
- [ ] Refactor/rename `BlockchainUpdaterImplSpec` so that it's obvious what it tests (hint: it's address transactions)
- [ ] Remove address transactions related tests from `LevelDBWriterSpec`
- [ ] Remove `com.wavesplatform.utils.Implicits`:
    - replace SubjectOps.empty with `PublishSubject` or whatever
    - remove `blockchainToApiExtensions` (see above)
- [ ] Remove `BlockchainAddressTransactionsList`
- [ ] Remove `accountDataKeys` and `accountDataInfo` from `Blockchain`
- [ ] Is there really a reason to have `LockedCompositeBlockchain` (or any other locks, for that matter)?
- [ ] Make `LevelDBDistributions` explicitly depend only on `DB` and pass a `FeatureProvider` (or just a boolean flag) into `portfolio`
- [ ] Remove all implicit schedulers from generic interfaces: it's the duty of consumers (REST API or gRPC endpoints) to provide appropriate execution environment.
- [ ] Remove other classes/traits which are no longer needed

PR #2603, opened 2019-09-18
Test for reward block headers

finnoio:


finnoio:
Test is already merged here:
https://github.com/wavesplatform/Waves/pull/2607/files

PR #2611, opened 2019-09-18
gRPC events stream

Karasiq:
https://github.com/wavesplatform/protobuf-schemas/pull/5/files


dvshur:
We've already started to make a proper implementation of this feature with keeping historical data, consumer disconnects etc. handling. So I suggest we wait for now.

phearnot:
This has been implemented in #3226

PR #2613, opened 2019-09-19
NODE-1812: Error message fixing

chepiov:


node/src/main/scala/com/wavesplatform/state/diffs/InvokeScriptTransactionDiff.scala
```diff
@@ -160,13 +160,14 @@ object InvokeScriptTransactionDiff {
                 .count(blockchain.hasAssetScript) +
                 ps.count(_._3.fold(false)(id => blockchain.hasAssetScript(IssuedAsset(id)))) +
                 (if (blockchain.hasScript(tx.sender)) 1 else 0)
-            val minWaves = totalScriptsInvoked * ScriptExtraFee + FeeConstants(InvokeScriptTransaction.typeId) * FeeUnit
-            val txName   = Constants.TransactionNames(InvokeScriptTransaction.typeId)
+            val minWaves  = totalScriptsInvoked * ScriptExtraFee + FeeConstants(InvokeScriptTransaction.typeId) * FeeUnit
+            val txName    = Constants.TransactionNames(InvokeScriptTransaction.typeId)
+            val assetName = s"${tx.assetFee._1.fold("WAVES")(_.id.base58)}"
```
phearnot:
```suggestion
            val assetName = tx.assetFee._1.fold("WAVES")(_.id.toString)
```
(also in line 102)

PR #2637, opened 2019-09-26
Fix appender execution contexts

Karasiq:


phearnot:
Having looked deeper into the original issue, I find these changes unnecessary, because callback execution does not really affect the way appender thread changes blockchain.

PR #2668, opened 2019-10-16
NODE-1911 add JENKINS_JOB_ID env value to names of test containers

msmolyakov:


mtuktarov:
I've added the removal part in Jenkins - everything works just fine! Logic:

Before starting IT I check if there any sbt processes with `;update;clean;it/test` arguments. If no, there are no parallel Integration Test jobs running so i remove all containers that contain 'waves' and 'node'.
Then I start IT:
```
SBT_THREAD_NUMBER=7 JENKINS_JOB_ID=jenkins-${BUILD_ID} sbt ";update;clean;it/test"
```
After it is finished or failed I kill and remove all containers containing jenkins-${BUILD_ID}

PR #2685, opened 2019-10-24
NODE-1928: CreateAlias transaction new format

Karasiq:


node/src/main/scala/com/wavesplatform/transaction/serialization/TxSerializer.scala
```diff
@@ -0,0 +1,17 @@
+package com.wavesplatform.transaction.serialization
+
+import com.wavesplatform.transaction.Transaction
+import play.api.libs.json.JsObject
+
+import scala.util.Try
+
+trait TxSerializer[T <: Transaction] {
```
phearnot: Remove this type

node/src/main/scala/com/wavesplatform/transaction/sign/TxSigner.scala
```diff
@@ -0,0 +1,8 @@
+package com.wavesplatform.transaction.sign
+
+import com.wavesplatform.account.PrivateKey
+import com.wavesplatform.transaction.Transaction
+
+trait TxSigner[T <: Transaction] {
```
phearnot: Remove this type

PR #2708, opened 2019-11-05
NODE-1939: CreateAliasTransaction gRPC tests

finnoio:


node-it/src/test/scala/com/wavesplatform/it/sync/grpc/CreateAliasTransactionGrpcSuite.scala
```diff
@@ -0,0 +1,122 @@
...
+    val balance = sender.grpc.wavesBalance(aliasCreatorAddr).available
+    val alias   = randomAlias()
+    assertGrpcError(sender.grpc.broadcastCreateAlias(aliasCreator, alias, balance + minFee), "negative waves balance", Code.INVALID_ARGUMENT)
+  }
+
+  private def randomAlias(): String = {
+    s"testalias.${Random.alphanumeric.take(9).mkString}".toLowerCase
+  }
+
+}
```
DrBlast: remove test "Able to resolve alias via gRPC method" from basic alias suite

PR #2751, opened 2019-11-20
SC-486 Correct callable V4 results

xrtm000:


node/src/main/scala/com/wavesplatform/transaction/smart/script/trace/TraceStep.scala
```diff
@@ -72,26 +73,62 @@ case class InvokeScriptTrace(
...
+      "description"  -> issue.description,
+      "isReissuable" -> issue.isReissuable,
+      "quantity"     -> issue.quantity,
+      "name"         -> issue.name
+    )
+
+  private def reissueJson(reissue: Reissue) =
+    Json.obj(
+      "assetId"      -> reissue.assetId.toString,
+      "isReissuable" -> reissue.isReissuable,
```
msmolyakov: rename to "reissuable" as field Asset.reissuable and as in json

node/src/main/scala/com/wavesplatform/transaction/smart/script/trace/TraceStep.scala
```diff
@@ -72,26 +73,62 @@ case class InvokeScriptTrace(
...
+      "decimals"     -> issue.decimals,
+      "description"  -> issue.description,
+      "isReissuable" -> issue.isReissuable,
+      "quantity"     -> issue.quantity,
+      "name"         -> issue.name
+    )
+
+  private def reissueJson(reissue: Reissue) =
+    Json.obj(
+      "assetId"      -> reissue.assetId.toString,
```
msmolyakov: is needed to rename to "id" as field `Asset.id`? How do you think?

node/src/main/scala/com/wavesplatform/transaction/smart/script/trace/TraceStep.scala
```diff
@@ -72,26 +73,62 @@ case class InvokeScriptTrace(
...
+      "isReissuable" -> reissue.isReissuable,
+      "quantity"     -> reissue.quantity
+    )
+
+  private def burnJson(burn: Burn) =
+    Json.obj(
+      "assetId"  -> burn.assetId.toString,
```
msmolyakov: is needed to rename to "id" as field `Asset.id`? How do you think?

node/src/main/scala/com/wavesplatform/transaction/smart/script/trace/TraceStep.scala
xrtm000: I would not change to make it clearer

PR #2759, opened 2019-11-25
SC-501 Store complexity by callable in db

xrtm000:


node/src/main/scala/com/wavesplatform/account/Address.scala
```diff
@@ -14,6 +14,8 @@ import play.api.libs.json._
 sealed trait Address extends AddressOrAlias {
   val bytes: ByteStr
   lazy val stringRepr: String = bytes.toString
+
+  override val toString: String = stringRepr
```
phearnot: `toString` is already defined in base trait

PR #2811, opened 2019-12-11
SC-520-minSponsoredFee-in-AssetInfo

potan:
Add minSponsoredAssetFee to the AssetInfo.
Reorder Issue arguments.


phearnot:
Fixed in #2846

PR #2816, opened 2019-12-13
NODE-1974-api-balances-on-height

potan:
API method


node/src/main/scala/com/wavesplatform/api/http/AddressApiRoute.scala
```diff
@@ -382,6 +394,23 @@ case class AddressApiRoute(settings: RestAPISettings, wallet: Wallet, blockchain
...
+  def balancesJson(height: Int, addreses: Seq[String]) = {
+    import com.wavesplatform.common.state.ByteStr
+    import com.wavesplatform.state.BalanceSnapshot
+    implicit val balancesWrites: Writes[(String, BalanceSnapshot)] = Writes[(String, BalanceSnapshot)] { b =>
+      Json.obj("id" -> b._1, "balance" -> b._2.regularBalance)
+    }
+
+    val balances = addreses.flatMap { address =>
+      Address.fromString(address).right.toOption.flatMap { a =>
+        blockchain.balanceSnapshots(a, height, ByteStr(Array[Byte]())).headOption.map { balance =>
```
phearnot: When no block id is provided, all snapshots up to current height are loaded. This process can take too long. Instead of an empty id, please specify the ID of the block at height `height`.

phearnot (changes requested):
Please also handle `POST` requests, as per spec.

potan:
> Please also handle `POST` requests, as per spec.

Done

phearnot:
This will be merged from 0.17.x

PR #2824, opened 2019-12-18
NODE-1947: Reject txs from UTX when blockchain is stale

Karasiq:


node/src/main/resources/application.conf
```diff
@@ -340,6 +340,8 @@ waves {
     allow-transactions-from-smart-accounts = true
     # Allow skipping checks with highest fee
     allow-skip-checks = true
+    # Reject all txs when blockchain is stale
+    reject-txs-when-blockchain-is-stale = true
```
phearnot:
```suggestion
    reject-transactions-when-blockchain-is-stale = true
```
to make it consistent with other settings

PR #2827, opened 2019-12-18
NODE-1977 Bump Kamon dependencies versions

xrtm000:


xrtm000:
API has changed largely, so this PR was deferred until it would become necessary.

PR #2836, opened 2019-12-19
NODE-1973: Amount Serialization Tests

finnoio:


finnoio:
moved to https://github.com/wavesplatform/Waves/pull/2876

PR #2849, opened 2019-12-25
SC-526 Allow to interrupt ScriptEstimatorV3

xrtm000:


lang/shared/src/main/scala/com/wavesplatform/lang/v1/estimator/v3/ScriptEstimatorV3.scala
```diff
@@ -10,8 +10,14 @@ import com.wavesplatform.lang.v1.task.imports._
 import com.wavesplatform.lang.v1.estimator.v3.EstimatorContext.EvalM
 import com.wavesplatform.lang.v1.estimator.v3.EstimatorContext.Lenses._
 import monix.eval.Coeval
+import monix.execution.misc.ThreadLocal
 
 object ScriptEstimatorV3 extends ScriptEstimator {
+  private val isInterrupted: ThreadLocal[Boolean] = ThreadLocal(false)
+
+  def interrupt(): Unit =
+    isInterrupted.set(true)
```
Karasiq: How it can be used, when a thread is already blocked?

xrtm000: Yes, I will provide other solution.

PR #2852, opened 2019-12-25
GZIP support in exporter and importer

Karasiq:


phearnot:
Once #2851 is merged, users will be able to pipe `gzip`'s output into importer.

PR #2860, opened 2020-01-03
Set default log level to INFO for systemd service

vlddm:


phearnot:
As described in the [docs](https://docs.wavesplatform.com/en/waves-node/logging-configuration.html), `logback.file.level` controls logging level for `waves.log` file, not the stdout (which is already set to `INFO` by default). It is intended behaviour that `waves.log` is more verbose that standard output.

vlddm:
> As described in the [docs](https://docs.wavesplatform.com/en/waves-node/logging-configuration.html), `logback.file.level` controls logging level for `waves.log` file, not the stdout (which is already set to `INFO` by default). It is intended behaviour that `waves.log` is more verbose that standard output.

waves.log with current default settings consume about 150Gb in a few days. 
Do you believe its what someone running fullnode wants?

PR #2902, opened 2020-01-27
NODE-2036: VRF in block header integration test

chepiov:
https://jira.wavesplatform.com/browse/NODE-2036


chepiov:
> add vrf length checks where defined

Done

PR #2925, opened 2020-02-10
NODE-2050: Proto block serialization

chepiov:
https://jira.wavesplatform.com/browse/NODE-2050


chepiov:
https://github.com/wavesplatform/Waves/pull/2935

PR #2940, opened 2020-02-18
NODE-2066-refactoring-InvokeScriptTransactionDiff

potan:


node/src/main/scala/com/wavesplatform/transaction/TxValidationError.scala
```diff
@@ -36,6 +36,7 @@ object TxValidationError {
   case class BlockAppendError(err: String, b: Block)           extends ValidationError
   case class ActivationError(err: String)                      extends ValidationError
   case class UnsupportedVersion(version: Int)                  extends ValidationError
+  case class ValueLimitsError(err: String)                     extends ValidationError
```
phearnot: There's no need for a special class if this error is just [converted](#diff-ad2c3659d051a49cdab23c5b1b627332R58) to `CustomValidationError` in the end.

PR #2967, opened 2020-03-04
SC-556-function-families

potan:
SigVerify_NKb


lang/tests/src/test/scala/com/wavesplatform/lang/v1/estimator/ScriptEstimatorV3Test.scala
```diff
@@ -113,4 +113,14 @@ class ScriptEstimatorV3Test extends ScriptEstimatorTestBase(ScriptEstimatorV3) {
     val script = "lastBlock.height"
     estimate(functionCosts(V3), compile(script)) shouldBe Right(2) /* ref eval and field access */
   }
+
+  property("groth16Verify_1inputs") {
```
ismagin: can we make it `1input`? :)

PR #2970, opened 2020-03-05
NODE-2070: Use block header hash as a reference

Karasiq:


node/src/main/scala/com/wavesplatform/network/messages.scala
```diff
@@ -51,19 +52,26 @@ case class BlockForged(block: Block) extends Message
...
 
-case class MicroBlockInv(sender: PublicKey, totalBlockSig: ByteStr, prevBlockSig: ByteStr, signature: ByteStr) extends Message with Signed {
+object MicroBlockResponse {
+  def apply(mb: MicroBlock): MicroBlockResponse = {
+    require(mb.version < Block.ProtoBlockVersion)
+    MicroBlockResponse(mb, mb.totalResBlockSig)
+  }
+}
+
+case class MicroBlockInv(sender: PublicKey, totalBlockRef: ByteStr, prevBlockRef: ByteStr, signature: ByteStr) extends Message with Signed {
```
phearnot:
```suggestion
case class MicroBlockInv(sender: PublicKey, totalBlockId: ByteStr, reference: ByteStr, signature: ByteStr) extends Message with Signed {
```

node/src/main/scala/com/wavesplatform/block/Block.scala
```diff
@@ -38,7 +38,10 @@ case class Block(
 ) extends Signed {
   import Block._
 
-  val uniqueId: ByteStr = signature
+  lazy val uniqueId: ByteStr =
```
phearnot: How about
```suggestion
  val id: Coeval[ByteStr] =
```
for consistency with transactions and JSON representation?

PR #3014, opened 2020-03-20
NODE-2070: Fix PB microblocks network format

Karasiq:


node/src/main/scala/com/wavesplatform/protobuf/block/PBMicroBlocks.scala
```diff
@@ -10,7 +11,7 @@ object PBMicroBlocks {
   import com.wavesplatform.common.state.ByteStr
   import com.wavesplatform.protobuf.utils.PBImplicitConversions._
 
-  def vanilla(signedMicro: PBSignedMicroBlock, unsafe: Boolean = false): Try[VanillaMicroBlock] = Try {
+  def vanilla(signedMicro: PBSignedMicroBlock, unsafe: Boolean = false): Try[(VanillaMicroBlock, BlockId)] = Try {
```
phearnot: The only place this code is used is in `PBMicroBlockSpec`, where a tuple is immediately converted to `MicroBlockResponse`, so why not change return type to `Try[MicroBlockResponse]`?

PR #3023, opened 2020-03-25
NODE-1881: Storage optimisations

phearnot:


node/src/main/scala/com/wavesplatform/database/Caches.scala
```diff
@@ -122,18 +122,23 @@ abstract class Caches(spendableBalanceChanged: Observer[(Address, Asset)]) exten
   override def assetScript(asset: IssuedAsset): Option[(Script, Long)] = assetScriptCache.get(asset)
 
   private var lastAddressId = loadMaxAddressId()
-  protected def loadMaxAddressId(): BigInt
+  protected def loadMaxAddressId(): Long
```
Karasiq:
```suggestion
  protected def loadMaxAddressId(): AddressId
```

project/plugins.sbt
```diff
@@ -1,8 +1,7 @@
 resolvers ++= Seq(
-  "Typesafe repository" at "http://repo.typesafe.com/typesafe/releases/",
-  "Artima Maven Repository" at "http://repo.artima.com/releases",
-  "JBoss" at "https://repository.jboss.org",
-  Resolver.sbtPluginRepo("releases"),
+  Resolver.typesafeRepo("releases"),
+//  "JBoss" at "https://repository.jboss.org",
```
Karasiq: Is jboss repo used?

node/src/main/scala/com/wavesplatform/database/Caches.scala
phearnot: This is actually a number, not a real address ID. It's only used in arithmetic operations when new address IDs are calculated, so I thought I'd leave it as is.

PR #3020, opened 2020-03-23
NODE-2066: Accept transactions with failed script result

chepiov:
https://jira.wavesplatform.com/browse/NODE-2066
https://jira.wavesplatform.com/browse/SC-581

PR #3030, opened 2020-03-30
NODE-2082: Fix importer no-verify

Karasiq:


chepiov:
@Karasiq fixed in #3020

PR #3035, opened 2020-04-02
NODE-2108: Asset balances by address fix

chepiov:
https://jira.wavesplatform.com/browse/NODE-2108


chepiov:
Fixed in https://github.com/wavesplatform/Waves/pull/3023

PR #3050, opened 2020-04-10
SC-585 API callable complexities

xrtm000:


node/src/test/scala/com/wavesplatform/http/UtilsRouteSpec.scala
```diff
@@ -174,6 +184,31 @@ class UtilsRouteSpec extends RouteSpec("/utils") with RestAPISettingsHelper with
...
 
+  val dApp =
```
msmolyakov, line 187: please add Verifier function to this script

xrtm000: +

node/src/test/scala/com/wavesplatform/http/UtilsRouteSpec.scala
```diff
@@ -174,6 +184,34 @@ class UtilsRouteSpec extends RouteSpec("/utils") with RestAPISettingsHelper with
...
+      |
+      |@Callable(i)
+      |func writeAndSendWaves(value: Int, recipient: String, amount: Int) = {
+      |    ScriptResult(
+      |        WriteSet([DataEntry("result", value)]),
+      |        TransferSet([ScriptTransfer(Address(recipient.fromBase58String()), amount, unit)])
+      |    )
+      |}
+      |
+      |@Verifier(tx)
```
msmolyakov, line 211: also check explicitly that without the verifier function the `complexity` field is 0

msmolyakov:
what if DApp script is empty (contains only directives)?

node/src/test/scala/com/wavesplatform/http/UtilsRouteSpec.scala
xrtm000: Ok

PR #3052, opened 2020-04-13
SC-587 Fix base16 register

xrtm000:


lang/jvm/src/main/scala/com/wavesplatform/lang/Global.scala
```diff
@@ -38,7 +38,7 @@ object Global extends BaseGlobal {
     } yield result
 
   override def base16EncodeImpl(input: Array[Byte]): Either[String, String] =
-    toEither(BaseEncoding.base16().encode(input))
+    toEither(BaseEncoding.base16().encode(input)).map(_.toLowerCase)
```
phearnot, line 41: I suggest caching a pre-configured instance of an encoding instead of calling `toLowerCase`:
```
val encoding = BaseEnconding.base16().lowerCase()
...
Right(encoding.encode(input))
```

lang/tests/src/test/scala/com/wavesplatform/lang/IntegrationTest.scala
```diff
@@ -1380,25 +1380,25 @@ class IntegrationTest extends PropSpec with PropertyChecks with ScriptGen with M
   }
 
   property("toBase16String limit 8Kb from V4") {
-    val base16String8Kb = "FEDCBA9876543210" * 1024
+    val base16String8Kb = "fedcba9876543210" * 1024
```
msmolyakov, line 1383: what if `"FEDCBA9876543210"`?
Should there be an error? Or should a function successfully handle uppercase letters?
Add a test.

xrtm000: ok

lang/jvm/src/main/scala/com/wavesplatform/lang/Global.scala
xrtm000: ok

PR #3053, opened 2020-04-13
NODE-2110: Change PoS min block time to 30s

Karasiq:


node/src/main/scala/com/wavesplatform/consensus/PoSCalculator.scala
```diff
@@ -82,19 +82,17 @@ object NxtPoSCalculator extends PoSCalculator {
...
 
-  private val MaxSignature: Array[Byte] = Array.fill[Byte](HitSize)(-1)
-  private val MaxHit: BigDecimal        = BigDecimal(BigInt(1, MaxSignature))
-  private val C1                        = 70000
-  private val C2                        = 5e17
-  private val TMin                      = 5000
+  private[this] val MaxSignature: Array[Byte] = Array.fill[Byte](HitSize)(-1)
+  private[this] val MaxHit: BigDecimal        = BigDecimal(BigInt(1, MaxSignature))
+  private[this] val C1                        = 70000
+  private[this] val C2                        = 5e17
```
phearnot, line 91: Please move all constants to `FairPoSCalculator` object.

node/src/main/scala/com/wavesplatform/consensus/PoSCalculator.scala
```diff
@@ -120,3 +118,7 @@ object FairPoSCalculator extends PoSCalculator {
     }
   }
 }
+
+object FairPoSCalculator extends FairPoSCalculator(30000) {
+  lazy val old = new FairPoSCalculator(5000)
```
phearnot, line 123: l
```suggestion
  lazy val V1 = new FairPoSCalculator(5000)
  lazy val V2 = new FairPoSCalculator(30000)
```

node/src/main/scala/com/wavesplatform/consensus/PoSSelector.scala
```diff
@@ -20,7 +20,11 @@ import scala.concurrent.duration.FiniteDuration
 class PoSSelector(blockchain: Blockchain, blockchainSettings: BlockchainSettings, syncSettings: SynchronizationSettings) extends ScorexLogging {
   import PoSCalculator._
 
-  protected def pos(height: Int): PoSCalculator = if (fairPosActivated(height)) FairPoSCalculator else NxtPoSCalculator
+  protected def posCalculator(height: Int): PoSCalculator =
```
phearnot, line 23:
```suggestion
  protected def posCalculator(height: Int): PoSCalculator =
    if (vrfActivated(height)) FairPoSCalculator.V2
    else if (fairPoSActivated(height)) FairPoSCalculator.V1
    else NxtPoSCalculator
```

node/src/main/scala/com/wavesplatform/consensus/PoSCalculator.scala
```diff
@@ -82,19 +83,22 @@ object NxtPoSCalculator extends PoSCalculator {
 
 }
 
-object FairPoSCalculator extends PoSCalculator {
+object FairPoSCalculator {
+  lazy val V1 = new FairPoSCalculator(5000)
+  lazy val V2 = new FairPoSCalculator(30000)
```
msmolyakov, line 88: change to 15 seconds

PR #3067, opened 2020-04-17
SC Ride V4 for REPL

xrtm000:


xrtm000:
RIDE V4 has been set already in REPL.

PR #3084, opened 2020-05-06
Sc 558 fix rollback tests

DrBlast:


msmolyakov:
These changes are moved to #3093

PR #3089, opened 2020-05-08
SC-601 Sponsor fee action

xrtm000:


lang/shared/src/main/scala/com/wavesplatform/lang/v1/evaluator/ScriptResult.scala
```diff
@@ -194,32 +204,47 @@ object ScriptResult {
         err(other, V4, FieldNames.Burn)
     }
 
+  private def processSponsorFee(fields: Map[String, EVALUATED]): Either[String, SponsorFee] =
+    (fields.get(FieldNames.SponsorFeeAssetId), fields.get(FieldNames.SponsorFeeMinFee)) match {
+      case (Some(CONST_BYTESTR(assetId)), Some(minFeeOpt)) =>
+        val minFeeValueOpt = minFeeOpt match {
+          case CONST_LONG(minFee)         => Right(Some(minFee))
+          case CaseObj(_, m) if m.isEmpty => Right(None)
```
potan, line 212: Not only Unit is allowed. Is this correct?

xrtm000: Fixed

xrtm000:
> Can we make scripted asset sponsored?
@potan
No, such invokes should be failed.

PR #3092, opened 2020-05-12
SC-545 EvaluatorV2 and refactoring

xrtm000:


lang/shared/src/main/scala/com/wavesplatform/lang/v1/compiler/Decompiler.scala
```diff
@@ -181,8 +181,8 @@ object Decompiler {
             val name = extractFunctionName(ctx, header)
             argsCoeval.map(as => out(s"$name(${as.mkString(", ")})", i))
         }
-      case _: Terms.ARR     => ??? // never happens
-      case _: Terms.CaseObj => ??? // never happens
+      case _: Terms.ARR       => ??? // never happens
```
potan, line 184: How about processing ARR too?

lang/tests/src/test/scala/com/wavesplatform/lang/IntegrationTest.scala
```diff
@@ -14,20 +15,19 @@ import com.wavesplatform.lang.v1.compiler.Terms._
 import com.wavesplatform.lang.v1.compiler.Types.{BYTESTR, FINAL, LONG, STRING}
 import com.wavesplatform.lang.v1.compiler.{ExpressionCompiler, Terms}
 import com.wavesplatform.lang.v1.evaluator.Contextful.NoContext
-import com.wavesplatform.lang.v1.evaluator.EvaluatorV1._
 import com.wavesplatform.lang.v1.evaluator.ctx._
 import com.wavesplatform.lang.v1.evaluator.ctx.impl.PureContext.MaxListLengthV4
 import com.wavesplatform.lang.v1.evaluator.ctx.impl.waves.WavesContext
 import com.wavesplatform.lang.v1.evaluator.ctx.impl.{PureContext, _}
-import com.wavesplatform.lang.v1.evaluator.{Contextful, ContextfulVal, EvaluatorV1}
+import com.wavesplatform.lang.v1.evaluator.{Contextful, ContextfulVal, EvaluatorV2}
```
potan, line 22: I mean we should parametrize tests by evaluator to test both with same code.

xrtm000: Does it make sense?
IntegrationTest is not a specification for evaluator, and further EvaluatorV1 will be removed.

lang/shared/src/main/scala/com/wavesplatform/lang/v1/compiler/Decompiler.scala
xrtm000: Maybe. But there is no ARR in compiled user source code.

PR #3094, opened 2020-05-13
SC-567 Test RIDE Block info

xrtm000:


node-it/src/test/scala/com/wavesplatform/it/sync/smartcontract/RideBlockInfoSuite.scala
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  test("able to retrieve genesis block info") {
+    val height = 1
+    sender.invokeScript(caller, dApp, func = Some("blockInfo"), args = List(CONST_LONG(height)), waitForTx = true)
+
+    checkCommonFields(sender.blockAt(height))
+  }
+
+  ignore("liquid blocks affect vrf field") {
```
msmolyakov, line 115: maybe just remove this test or check that vrf is not changed?

PR #3097, opened 2020-05-14
SC-591 Overloaded Issue action constructor

xrtm000:


lang/doc/v4/funcs/internal-functions.hjson
```diff
@@ -5,6 +5,18 @@
...
       paramsDoc: [ "The value", "The type name" ]
+    },
+    {
+      name: "Issue"
+      params: [ "String", "String", "Int", "Int", "Boolean", "Script|Unit", "Int" ]
+      doc: "Issue action detailed constructor"
+      paramsDoc: [ "Token name", "Token description", "Amount of the token", "Number of digits in decimal part", "Reissue ability flag", "Smart asset script (optional)", "Sequential number"]
+    },
+    {
+      name: "Issue"
```
msmolyakov, line 16: are they will be shown in the surfboard help, for example?

lang/shared/src/main/scala/com/wavesplatform/lang/v1/evaluator/ctx/impl/waves/Functions.scala
```diff
@@ -534,4 +534,34 @@ object Functions {
...
+      FieldNames.IssueQuantity -> LONG,
+      FieldNames.IssueDecimals -> LONG,
+      FieldNames.IssueIsReissuable -> BOOLEAN,
+    ) {
+      args =>
+        val typedArgs = (issueActionType.fields.map(_._1) zip (args ::: List(unit, CONST_LONG(0)))).toMap
+        Right(CaseObj(issueActionType, typedArgs))
+    }
+
+  val detailedIssueActionConstructor: BaseFunction[Environment] =
```
msmolyakov, line 552: Is the origin constructor of the Issue structure removed from the context?

xrtm000: Not removed but overlapped

lang/doc/v4/funcs/internal-functions.hjson
xrtm000: should be

PR #3101, opened 2020-05-15
SC-584-WebEnvironment

potan:


lang/shared/src/main/scala/com/wavesplatform/lang/v1/repl/package.scala
```diff
@@ -16,10 +16,12 @@ package object repl {
   val internalVarPrefixes: Set[Char] = Set('@', '$')
   val internalFuncPrefix: String = "_"
 
+  val version = V4
```
xrtm000, line 19: Is it needed?

lang/shared/src/main/scala/com/wavesplatform/lang/v1/evaluator/ctx/impl/waves/Functions.scala
```diff
@@ -350,7 +350,7 @@ object Functions {
       balanceDetailsType,
       ("addressOrAlias", addressOrAliasType)
     ) {
-      new ContextfulNativeFunction[Environment]("assetBalance", LONG, Seq(("addressOrAlias", addressOrAliasType),("assetId", UNION(UNIT, BYTESTR)))) {
+      new ContextfulNativeFunction[Environment]("wavesBalance", LONG, Seq(("addressOrAlias", addressOrAliasType))) {
```
xrtm000, line 353: Could we remove this function definition?

potan: Removing this definition requires refactor ContextfulNativeFunction class.

lang/shared/src/main/scala/com/wavesplatform/lang/v1/repl/node/http/WebEnvironment.scala
```diff
@@ -80,7 +84,11 @@ private[repl] case class WebEnvironment(settings: NodeConnectionSettings) extend
   override def transactionById(id: Array[Byte]): Future[Option[Tx]]           = ???
   override def multiPaymentAllowed: Boolean                                   = ???
   override def txId: ByteStr                                                  = ???
-  override def transferTransactionFromProto(b: Array[Byte]): Option[Tx.Transfer] = ???
+
+  override def transferTransactionFromProto(b: Array[Byte]): Future[Option[Tx.Transfer]] = {
+    getEntity[Option, TransferTransaction, Tx.Transfer](s"/transactions/byProtoBytes/${Base64.encode(b).replace("/", "%2f").replace("+", "%2b")}")
```
msmolyakov, line 89: What is this endpoint? Can't find it

PR #3107, opened 2020-05-18
SC-559 MakeString function

xrtm000:


lang/shared/src/main/scala/com/wavesplatform/lang/v1/evaluator/ctx/impl/PureContext.scala
```diff
@@ -514,6 +514,21 @@ object PureContext {
       )
   }
 
+  lazy val makeString: BaseFunction[NoContext] =
+    NativeFunction("makeString", 30, MAKESTRING, listString, ("list", LIST(STRING)), ("separator", STRING)) {
+      case ARR(list) :: CONST_STRING(separator) :: Nil =>
+        val expectedStringSize =
```
potan, line 520: Could we check limits by array weight without calculation utf8 bytes?

xrtm000: Fixed

lang/shared/src/main/scala/com/wavesplatform/lang/v1/evaluator/ctx/impl/PureContext.scala
```diff
@@ -514,6 +514,18 @@ object PureContext {
       )
   }
 
+  lazy val makeString: BaseFunction[NoContext] =
+    NativeFunction("makeString", 30, MAKESTRING, listString, ("list", LIST(STRING)), ("separator", STRING)) {
+      case ARR(list) :: CONST_STRING(separator) :: Nil =>
+        val expectedStringSize = list.map(_.asInstanceOf[CONST_STRING].weight).sum
```
potan, line 520: Could we use weight of ARR object?

xrtm000: Yes

PR #3113, opened 2020-05-19
NODE-2125: Extensions initializing during Import

chepiov:
https://jira.wavesplatform.com/browse/NODE-2125

### DIFF
--- a/node/src/main/scala/com/wavesplatform/Importer.scala
+++ b/node/src/main/scala/com/wavesplatform/Importer.scala
@@ -33,7 +33,7 @@ import scopt.OParser
 
 import scala.concurrent.duration._
 import scala.concurrent.{Await, Future}
-import scala.util.Success
+import scala.util.{Failure, Success, Try}
 
 object Importer extends ScorexLogging {
   import monix.execution.Scheduler.Implicits.global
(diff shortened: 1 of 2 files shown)


node/src/main/scala/com/wavesplatform/Importer.scala
```diff
@@ -109,18 +112,18 @@ object Importer extends ScorexLogging {
...
-          override def blocksApi                                             = ???
-          override def accountsApi                                           = ???
-          override def assetsApi                                             = ???
+          override def broadcastTransaction(tx: Transaction): TracedResult[ValidationError, Boolean] = notImplementedDuringImport
+          override def spendableBalanceChanged: Observable[(Address, Asset)]                         = notImplementedDuringImport
+          override def actorSystem: ActorSystem                                                      = notImplementedDuringImport
+          override def blockchainUpdated: Observable[BlockchainUpdated]                              = blockchainUpdatedObservable
+          override def utxEvents: Observable[UtxEvent]                                               = Observable.empty
+          override def transactionsApi: CommonTransactionsApi                                        = notImplementedDuringImport
+          override def blocksApi: CommonBlocksApi                                                    = notImplementedDuringImport
```
phearnot, line 124: Instead of throwing an exception, please supply a usable instance, just in case the extension needs to actually do something during import.

PR #3120, opened 2020-05-21
SC-578-allow-FOLD-single-expr

potan:


potan:
> What about ParserV2?

It works ok.

xrtm000:
Add test for ParserV2, please :)

PR #3131, opened 2020-05-26
NODE-2144: Failed transactions complexity

chepiov:
https://jira.wavesplatform.com/browse/NODE-2144


node/src/main/scala/com/wavesplatform/transaction/TxValidationError.scala
```diff
@@ -47,94 +48,87 @@ object TxValidationError {
...
+    sealed trait Cause extends Product with Serializable {
+      def code: Int
     }
-
-    private val target: String = this match {
-      case _: ByAssetScript         => "Asset"
-      case _: ByAssetScriptInAction => "Asset"
-      case _: ByAccountScript       => "Account"
+    object Cause {
+      case object DAppExecution extends Cause {
```
DrBlast, line 103: what is the difference with AssetScriptInAction and AssetScript?
And DAppExecution and FeeForActions
may be rename error codes?

chepiov: AssetScriptInAction is for asset scripts executed in dApp actions, AssetScript is for remaining asset scripts.
DAppExecution is for errors throwed by dApp script, FeeForActions is for case when there is not enough fee for all actions in dApp.

I added .toString representation of FailedTransactionError: https://github.com/wavesplatform/Waves/blob/125626afcd91216ad5969719673fee0969966ba3/node/src/main/scala/com/wavesplatform/transaction/TxValidationError.scala#L79

node/src/main/scala/com/wavesplatform/transaction/TxValidationError.scala
```diff
@@ -47,94 +48,89 @@ object TxValidationError {
     override def toString: String = s"InvalidSignature(${s.toString + " reason: " + details})"
   }
 
+  sealed trait WithLog extends Product with Serializable {
+    def log: Log[Id]
+  }
+
   /** Errors which can produce failed transaction */
-  sealed trait FailedTransactionError extends Product with Serializable with ValidationError {
+  case class FailedTransactionError private (
```
phearnot, line 56: Marking case class constructor as private does not prevent others from creating new instances by calling companion object's `apply` method. Besides, Is there really a reason to make the constructor private?

PR #3146, opened 2020-06-08
Scala 2.13

phearnot:


lang/shared/src/main/scala/com/wavesplatform/lang/v1/repl/node/http/response/model/Transaction.scala
```diff
@@ -92,7 +92,7 @@ object Transaction {
       version.toByte,
       senderPublicKey,
       proofs,
-      bodyBytes,
+      ByteString(),
```
msmolyakov, line 95: how to get real bodyBytes?

phearnot: https://jira.wavesplatform.com/browse/SC-631

lang/tests/src/test/scala/com/wavesplatform/lang/DirectiveParserTest.scala
```diff
@@ -68,8 +68,8 @@ class DirectiveParserTest extends PropSpec with PropertyChecks with Matchers {
 
   property("directive with illegal format") {
     parse("{-# ILLEGAL #-}") shouldBe Left("Directive {-# ILLEGAL #-} has illegal format")
-    parse("{-# A%@#%$^$ A%#%%$ #-}") shouldBe 'left
-    parse("{-# AAA BBB CCC DDD #-}") shouldBe 'left
+    parse("{-# A%@#%$^$ A%#%%$ #-}") shouldBe Symbol("left")
```
msmolyakov, line 71: why can't we use `'left` now?

phearnot: Symbol literals are deprecated in 2.13

PR #3149, opened 2020-06-09
SC-618-this-alias

potan:
Allow use this while calling by alias by fork parameter BlockV5.


node-it/src/test/scala/com/wavesplatform/it/sync/smartcontract/InvokeScriptTransactionSuite.scala
```diff
@@ -110,6 +156,36 @@ class InvokeScriptTransactionSuite extends BaseTransactionSuite with CancelAfter
...
+      caller,
+      firstContract,
+      func = Some("baz"),
+      args = List(),
+      payment = Seq(),
+      fee = 1.waves,
+      waitForTx = true
+     )
+    sender.getDataByKey(firstContract, "test") shouldBe BinaryDataEntry("test", ByteStr.decodeBase58(firstContract).get)
+  }
```
finnoio, line 187: Add test where dApp is called by alias and call "baz" function, which uses "this" expression.

PR #3156, opened 2020-06-19
NODE-2150: Native curve25519 usage

chepiov:
https://jira.wavesplatform.com/browse/NODE-2150


node/src/main/scala/com/wavesplatform/crypto/package.scala
```diff
@@ -24,7 +24,9 @@ package object crypto {
     val constructor = classOf[OpportunisticCurve25519Provider].getDeclaredConstructors.head
       .asInstanceOf[Constructor[OpportunisticCurve25519Provider]]
     constructor.setAccessible(true)
-    constructor.newInstance()
+    val p = constructor.newInstance()
+    log.info(s"Native provider used: ${p.isNative}")
```
msmolyakov, line 28: why is this log message needed?

phearnot (approved):

PR #3164, opened 2020-06-25
NODE-2158: Public /debug/validate method

Karasiq:


node/src/test/scala/com/wavesplatform/http/AssetsBroadcastRouteSpec.scala
```diff
@@ -191,9 +191,9 @@ class AssetsBroadcastRouteSpec
           senderPrivateKey,
           receiverPrivateKey.toAddress,
           Asset.Waves,
-          1 * Waves,
+          TestValues.OneWaves,
```
msmolyakov, line 194: In integration tests we use `1.waves`. Maybe use this in unit tests too?

PR #3166, opened 2020-06-26
Integration tests fixes

Frozen:


node-it/src/test/scala/com/wavesplatform/it/api/SyncGrpcApi.scala
```diff
@@ -73,7 +73,7 @@ object SyncGrpcApi extends Assertions {
 
     def resolveAlias(alias: String): Addr = {
       val addr = accounts.resolveAlias(StringValue.of(alias))
-      Addr.fromBytes(addr.value.toByteArray).explicitGet()
+      PBRecipients.toAddress(addr.value.toByteArray, AddressScheme.current.chainId).explicitGet()
```
phearnot, line 76: In fact, accounts api returns the complete address, so the original implementation seems to be correct.

phearnot:
Changing error messages, although appropriate, breaks backward compatibility, so I'll close this for now.

PR #3174, opened 2020-07-03
NODE-2172: Support extensions in Docker

Karasiq:


docker/entrypoint.sh
```diff
@@ -36,11 +39,12 @@ echo "WAVES_WALLET_SEED='${WAVES_WALLET_SEED}'" | tee -a /var/log/waves/waves.lo
...
-	${JAVA_OPTS} \
-	-jar /usr/share/waves/lib/waves-all-${WAVES_VERSION}.jar /etc/waves/waves.conf
-
+exec java -Dlogback.stdout.level=${WAVES_LOG_LEVEL} \
+  -XX:+ExitOnOutOfMemoryError \
+  -Xmx${WAVES_HEAP_SIZE} \
+  -Dlogback.file.directory=/var/log/waves \
+  -Dconfig.override_with_env_vars=true \
+  ${JAVA_OPTS} \
+  -cp "/usr/share/waves/lib/*:/usr/share/waves/lib/plugins/*" \
```
DenisSerebryansky, line 48: In our case we had to swap source directories in -cp arguments. Otherwise, line

```waves.extensions += "com.wavesplatform.dex.grpc.integration.DEXExtension"```

 in [our extension's](https://github.com/wavesplatform/matcher/blob/master/waves-ext/src/main/resources/application.conf#L1) `application.conf` is overwritten by [default node configuration](https://github.com/wavesplatform/Waves/blob/master/node/src/main/resources/application.conf#L348). Do you face this problem?

DenisSerebryansky: I just tried to start Node with our extension in your way. As I said, its required to swap class paths of Node and plugins

Karasiq: You mean /usr/share/waves/lib/plugins/*:/usr/share/waves/lib/*?

DenisSerebryansky: Yep

PR #3175, opened 2020-07-06
SC-644 Test complexity of every function

xrtm000:


lang/tests/src/test/scala/com/wavesplatform/lang/v1/estimator/FunctionComplexityTest.scala
```diff
@@ -0,0 +1,63 @@
...
+
+class FunctionComplexityTest extends PropSpec with PropertyChecks with Matchers with NoShrink {
+  val directives: Iterable[DirectiveSet] =
+  DirectiveDictionary[StdLibVersion].all
+      .flatMap(
+        version =>
+          Seq(
+            DirectiveSet(version, Account, Expression).explicitGet(),
+            DirectiveSet(version, Asset, Expression).explicitGet()
+          ) ++ (if (version >= V3) Seq(DirectiveSet(version, Account, Expression).explicitGet())
```
finnoio, line 25: if (version >= V3) than merge Seq of DirectiveSet with DApp content type like this `Seq(DirectiveSet(version, Account, DApp)`

xrtm000: Done

PR #3201, opened 2020-07-17
SC-458 Update Java adapter with new methods

xrtm000:


lang/jvm/src/main/java/com/wavesplatform/lang/Script.java
```diff
@@ -0,0 +1,27 @@
+package com.wavesplatform.lang;
+
+import com.wavesplatform.common.utils.Base64;
+
+public interface Script {
+    int version();
+
+    boolean isExpression();
```
msmolyakov, line 8: maybe single enum field?

lang/jvm/src/main/scala/com/wavesplatform/lang/JavaAdapter.scala
```diff
@@ -1,29 +1,171 @@
...
+        directives   <- DirectiveParser(input)
+        directiveSet <- Directive.extractDirectives(directives, defaultStdLib = V4)
+        compile = if (directiveSet.contentType == DApp)
+          compileDApp _
+        else
+          compileExpression _
+        compiled <- compile(input, ctxCache(directiveSet).compilerContext, directiveSet)
+      } yield compiled
+
+    script.fold(error => throw new RideException(error), identity)
```
msmolyakov, line 52: maybe handled exception?

msmolyakov (changes requested):

lang/jvm/src/main/java/com/wavesplatform/lang/Script.java
xrtm000: Why?
There are few options.

lang/jvm/src/main/scala/com/wavesplatform/lang/JavaAdapter.scala
xrtm000: Checked exception?
It is not so useful.

potan (approved):

mrkraft (approved):

PR #3204, opened 2020-07-20
NODE-2182: Custom min block interval

Karasiq:


node/src/main/resources/application.conf
```diff
@@ -143,61 +143,9 @@ waves {
...
-    #        average-block-delay = 60s
-    #
-    #        # List of genesis transactions
-    #        transactions = [
-    #          {recipient = "BASE58ADDRESS1", amount = 50000000000000},
-    #          {recipient = "BASE58ADDRESS2", amount = 50000000000000}
-    #        ]
-    #      }
-    #    }
+    custom {
```
DrBlast, line 146: custom config should be commented by default

PR #3206, opened 2020-07-21
SC-457 Improve parse error message in REPL

xrtm000:


lang/shared/src/main/scala/com/wavesplatform/lang/v1/repl/ReplEngine.scala
```diff
@@ -36,7 +36,7 @@ class ReplEngine[F[_] : Monad] {
   private def parse(expr: String): Either[String, EXPR] =
     Parser.parseExprOrDecl(expr)
       .fold(
-        { case (_, _, err) => Left(err.trace().toString) },
+        { case _           => Left(s"Can't parse '$expr'") },
```
potan, line 39: Is there enough error information in this message?

xrtm000: Currently Parser doesn't produce any useful error messages.

PR #3212, opened 2020-07-24
SC-657 Fix /address/scriptInfo complexity field

xrtm000:


node/src/test/scala/com/wavesplatform/http/AddressRouteSpec.scala
```diff
@@ -184,7 +184,7 @@ class AddressRouteSpec
       (response \ "address").as[String] shouldBe allAddresses(1).toString
       (response \ "script").as[String] shouldBe "base64:AQa3b8tH"
       (response \ "scriptText").as[String] shouldBe "true"
-      (response \ "complexity").as[Long] shouldBe 1
+      (response \ "complexity").as[Long] shouldBe 0
```
msmolyakov, line 187: why is the complexity of the expression `true` not considered?

xrtm000: Fixed

PR #3236, opened 2020-08-13
set stopsignal to sigint

ghost:
Setting `STOPSIGNAL` to `SIGINT` is required otherwise stopping containers will cause the node not to save any bloom filters.


phearnot:
By default `docker stop` [sends](https://docs.docker.com/engine/reference/commandline/stop/) `SIGTERM` to the main process inside container. When java process receives `SIGTERM`, it initializes an orderly shutdown of the virtual machine, which includes running shutdown hooks (which in turn close LevelDB and save bloom filters). Could your issue be related to `SIGKIL` which is sent after a grace period?

PR #3254, opened 2020-08-31
Minimize concurrent leveldb usage

phearnot:


Karasiq (approved):

node/src/main/scala/com/wavesplatform/api/common/AddressTransactions.scala
```diff
@@ -72,36 +68,35 @@ object AddressTransactions {
   ): Iterable[(Height, Transaction, Boolean)] =
```
Karasiq, line 68: I recommend to replace it with Iterator
```suggestion
  ): Iterator[(Height, Transaction, Boolean)] =
```

DrBlast (approved):

PR #3268, opened 2020-09-15
SC-658-decompiller-add-brackets

potan:
Add brackets when IF operator place in expression.


xrtm000 (changes requested):
It is a special case.
Decompile of `(if(condition) then obj1 else obj2).field` also returns `if(condition) then obj1 else obj2.field`

xrtm000 (approved):

mrkraft (approved):

msmolyakov (approved):

PR #3283, opened 2020-10-08
NODE-2247: Move BlockchainUpdates extension to grpc-server module

Karasiq:


grpc-server/src/main/scala/com/wavesplatform/events/BlockchainUpdates.scala
```diff
@@ -24,7 +24,7 @@ import scala.util.{Failure, Success}
 class BlockchainUpdates(private val context: Context) extends Extension with ScorexLogging with BlockchainUpdateTriggers {
   implicit val scheduler: Scheduler = Scheduler(context.actorSystem.dispatcher)
 
-  private[this] val settings = context.settings.config.as[BlockchainUpdatesSettings]("blockchain-updates")
+  private[this] val settings = context.settings.config.as[BlockchainUpdatesSettings]("waves.blockchain-updates")
```
Karasiq, line 27: Maybe should be fixed in docs also

phearnot (approved):

grpc-server/src/main/resources/application.conf
```diff
@@ -1,4 +1,10 @@
-waves.grpc {
-  host = localhost
-  port = 6870
+waves {
+  grpc {
+    host = localhost
+    port = 6870
+  }
+
+  blockchain-updates {
```
msmolyakov, line 7: Shouldn't this setting be inside `grpc` because it's the same extension?
```
waves {
  grpc {
    host = localhost
    port = 6870
    blockchain-updates {
      port = 6881
...
```

msmolyakov (approved):

grpc-server/src/main/resources/application.conf
phearnot: Actually they're not. Both are packaged in the same tarball/package, but are quite independent.

PR #3285, opened 2020-10-11
Update application.conf

G1zm0nl:
In order for /debug/state and /debug/stateWaves to work again and not create either errors:

The server was not able to produce a timely response to your request.
* Connection #0 to host localhost left intact
Please try again in a short while!

Connection reset by peer

### DIFF
--- a/node/src/main/resources/application.conf
+++ b/node/src/main/resources/application.conf
@@ -421,11 +421,16 @@ akka {
 
   http.server {
     max-connections = 128
-    request-timeout = 20s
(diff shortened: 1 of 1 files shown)


phearnot:
20 seconds seem like a sensible default for most cases, so I don't think we'll change the default value at this time.
We've added a dedicated page documenting workarounds for slow requests: https://docs.waves.tech/en/waves-node/node-api/slow-requests, which I hope will help those who run into such issues.

PR #3289, opened 2020-10-13
NODE-2249: BlockchainUpdates fixes

Karasiq:
~~Depends from https://github.com/wavesplatform/Waves/pull/3283~~
Solves NODE-2249, NODE-2250, and NODE-2251


grpc-server/src/main/scala/com/wavesplatform/events/repo/UpdatesRepoImpl.scala
```diff
@@ -173,114 +152,97 @@ class UpdatesRepoImpl(directory: String)(implicit val scheduler: Scheduler)
...
 
 object UpdatesRepoImpl {
-  private val LEVELDB_READ_BATCH_SIZE = 1024
-
-  private val RANGE_REQUEST_MAX_SIZE = 1000
-  private val RANGE_REQUEST_TIMEOUT  = 1.minute
+  private val LevelDBReadBatchSize = 100
 
   private def key(height: Int): Array[Byte] =
     ByteBuffer.allocate(8).order(ByteOrder.BIG_ENDIAN).putInt(height).array()
```
Karasiq, line 248: Should it be 8 bytes?

phearnot: I suppose 4 bytes would've sufficed, but oh well. I don't think now is good time to change storage structure, maybe we'll fix this in one of the future releases.

phearnot (approved):

msmolyakov (approved):

PR #3296, opened 2020-10-22
SC-665 Pass estimator as lambda to Utils route

xrtm000:


phearnot (changes requested):
There should probably be a test which checks that estimator changes upon feature activation.

phearnot (approved):

msmolyakov (approved):

PR #3305, opened 2020-11-16
NODE-2273: /utils/script/evaluate API route

Karasiq:


phearnot (approved):

DrBlast (changes requested):
endpoint utils/script/evaluate/{address} not availiable and not added to swagger

node/src/test/scala/com/wavesplatform/http/UtilsRouteSpec.scala
```diff
@@ -674,6 +680,107 @@ class UtilsRouteSpec extends RouteSpec("/utils") with RestAPISettingsHelper with
     }
   }
 
+  routePath("/script/evaluate/{address}") in {
```
DrBlast, line 683: add tests for different types of function args: String, ByteVector, List[String|ByteVector|Boolean|Int]

DrBlast (approved):

node/src/test/scala/com/wavesplatform/http/UtilsRouteSpec.scala
```diff
@@ -674,6 +680,107 @@ class UtilsRouteSpec extends RouteSpec("/utils") with RestAPISettingsHelper with
...
+    val testScript = {
```
DrBlast, line 684: add tests for different types of function args: String, ByteVector, List[String|ByteVector|Boolean|Int]

DrBlast (changes requested):

DrBlast (approved):

PR #3307, opened 2020-11-24
NODE-2269: Check address seq range (REST API)

Karasiq:


phearnot (approved):

node/src/test/scala/com/wavesplatform/http/AddressRouteSpec.scala
```diff
@@ -88,6 +88,14 @@ class AddressRouteSpec
     }
 
     r1 shouldNot contain allElementsOf r2
+
+    Get(routePath("/seq/1/9000")) ~> route ~> check {
+      responseAs[JsObject] shouldBe Json.obj("error" -> 10, "message" -> "Too big sequences requested")
```
msmolyakov, line 93: 1. `sequences` -> `sequence`
2. Let's tell user in the error text what is the limit

node/src/test/scala/com/wavesplatform/http/AddressRouteSpec.scala
```diff
@@ -88,6 +88,14 @@ class AddressRouteSpec
...
+    }
+
+    Get(routePath("/seq/10/1")) ~> route ~> check {
+      responseAs[JsObject] shouldBe Json.obj("error" -> 10, "message" -> "Too big sequences requested")
```
msmolyakov, line 97: Why "too big sequence"? Еhe order of values is incorrect, but the range is not exceeded

msmolyakov (changes requested):

msmolyakov (approved):

PR #3308, opened 2020-11-25
NODE-2268: Fix assets distribution limit and depth checks (REST API)

Karasiq:


phearnot (approved):

node/src/test/scala/com/wavesplatform/http/RestAPISettingsHelper.scala
```diff
@@ -22,7 +22,7 @@ trait RestAPISettingsHelper {
         s"""waves.rest-api {
            |  api-key-hash = $keyHash
            |  transactions-by-address-limit = $MaxTransactionsPerRequest
-           |  distribution-by-address-limit = $MaxAddressesPerRequest
+           |  distribution-address-limit = $MaxAddressesPerRequest
```
msmolyakov, line 25: is this necessary to rename existing config parameter? What about backward compatibility?

msmolyakov (changes requested):

node/src/test/scala/com/wavesplatform/http/RestAPISettingsHelper.scala
msmolyakov: this is typo in test config

msmolyakov (approved):

PR #3364, opened 2021-02-09
SC-716 increase payments limit

potan:


msmolyakov (approved):

node/src/main/scala/com/wavesplatform/state/diffs/invoke/InvokeScriptDiff.scala
```diff
@@ -68,6 +69,7 @@ object InvokeScriptDiff {
 
           directives <- TracedResult.wrapE(DirectiveSet(version, Account, DAppType).leftMap(GenericError.apply))
           payments   <- TracedResult.wrapE(AttachedPaymentExtractor.extractPayments(tx, version, blockchain, DAppTarget).leftMap(GenericError.apply))
+          _          <- TracedResult(Either.cond(payments.payments.size <= version.maxPayments, (), GenericError(s"Script payment amount=${payments.payments.size} should not exceed ${version.maxPayments}")))
```
xrtm000: Why is it needed? `AttachedPaymentExtractor.extractPayments` contains this check

xrtm000 (changes requested):
Where is the check that total complexity doesn't exceed 52000?

xrtm000 (approved):

PR #3367, opened 2021-02-15
SC-715-use-unicode

potan:


msmolyakov (approved):

xrtm000 (changes requested):
Add tests for all functions including empty string test, check that behavior of split function has not been changed, see SplitFunctionTest

lang/tests/src/test/scala/com/wavesplatform/lang/IntegrationTest.scala
```diff
@@ -2174,4 +2174,88 @@ class IntegrationTest extends PropSpec with PropertyChecks with ScriptGen with M
...
+
+    val script5 = s"""dropRight("x冬x", 2)"""
+    eval(script5, version = ver) shouldBe
+      Right(CONST_STRING("x\ud87e").explicitGet())
+  }
+
+  property("unicode indexOf") {
+    val src =
+      """ "x冬xqweqwe".indexOf("we") """
+    genericEval[Environment, EVALUATED](src, ctxt = v5Ctx, version = V5, env = utils.environment) shouldBe Right(CONST_LONG(4L))
```
alexeykiselev, line 2325: What this function returns in V4 context? 
Also the funny script that may be tested
```
indexOf(take("世界x冬x", 4), take("冬", 1))
```
What it returns for V4 and V5?

potan: All characters are 2-bytes. It returns 3 in both cases.

xrtm000 (approved):

mrkraft (approved):

PR #3391, opened 2021-03-18
SC-730-hashScriptAtAddress

potan:


xrtm000 (changes requested):
Add test checking `hashScriptAtAddress` at non-scripted account

xrtm000 (approved):

mrkraft (approved):

PR #3451, opened 2021-04-23
SC-757 BigInt message fixes

xrtm000:


lang/shared/src/main/scala/com/wavesplatform/lang/v1/evaluator/ctx/impl/PureContext.scala
```diff
@@ -414,7 +414,7 @@ object PureContext {
           _ <- Either.cond(result < BigIntMax, (), s"Long overflow: value `$result` greater than 2^511-1")
           _ <- Either.cond(result > Long.MinValue, (), s"Long overflow: value `$result` less than -2^511-1")
         } yield CONST_BIGINT(result)
-      case xs => notImplemented[Id, EVALUATED]("fractionRounds(value: BigInt, numerator: BigInt, denominator: BigInt, round: rounds)", xs)
+      case xs => notImplemented[Id, EVALUATED]("fraction(value: BigInt, numerator: BigInt, denominator: BigInt, round: rounds)", xs)
```
msmolyakov: I think `round: rounds` this also does not apply to Ride. Maybe this should be `round: DOWN|CEILING|...`.

msmolyakov (changes requested):

msmolyakov (approved):

mrkraft (approved):

PR #3496, opened 2021-05-31
SC-781 Forbid sync call V3 and V4 DApps

xrtm000:


DrBlast (approved):

node/src/main/scala/com/wavesplatform/state/diffs/invoke/InvokeScriptDiff.scala
```diff
@@ -57,6 +57,13 @@ object InvokeScriptDiff {
     val result = blockchain.accountScript(dAppAddress) match {
       case Some(AccountScriptInfo(pk, ContractScriptImpl(version, contract), _, callableComplexities)) =>
         for {
+          _ <- traced(
+            Either.cond(
+              version >= V5,
+              (),
+              GenericError(s"Calling DApp is available only from V5, but DApp $version at address $dAppAddress was called from $invoker")
```
mrkraft: I suggest to replace error message with this: "DApp $invoker invoked DApp $dAppAddress that uses Ride version $version, but dApp-to-dApp invocation requires version 5 or higher."

mrkraft (changes requested):

node/src/main/scala/com/wavesplatform/state/diffs/invoke/InvokeScriptDiff.scala
xrtm000: Done

mrkraft (approved):

PR #3500, opened 2021-06-04
SC Generic blockchain context method

xrtm000:


vsuharnikov (approved):

phearnot (changes requested):
Please add a test which uses matcher's Blockchain implementation, which will check a script handling both orders and transactions. Transaction validation branch should access blockchain function, while order branch should not. Make sure exception is thrown for transaction and is not thrown for order.

phearnot (approved):

phearnot (approved):

PR #3511, opened 2021-06-18
NODE-2334: Fixed testnet/stagenet pool port

phearnot:


DrBlast (changes requested):
add fix for stagenet too
set peers port 6868

DrBlast (approved):

PR #3552, opened 2021-09-27
SC-799 InvokeDataEntriesBytesSuite

xrtm000:


Karasiq:
Doesnt work:
"Error while executing account-script: Storing data size should not exceed 15360, actual: 20456 bytes" did not include substring that matched regex Error while executing account-script: Storing data size should not exceed 15360, actual: 20476 bytes

PR #3588, opened 2021-11-18
NODE-2362: New data entries limit calculation

Karasiq:


node/src/test/scala/com/wavesplatform/transaction/data/DataTransactionNewLimitsSpec.scala
```diff
@@ -0,0 +1,69 @@
...
+
+  it should "handle 400 bytes key" in withDomain(DomainPresets.RideV6) { d =>
+    val dataEntry = StringDataEntry("a" * 400, "test")
+    val dataTransaction = TxHelpers.dataWithMultipleEntries(TxHelpers.defaultSigner, Seq(dataEntry))
+
+    d.helpers.creditWavesToDefaultSigner()
+    d.appendAndAssertSucceed(dataTransaction)
+    d.blockchain.accountData(TxHelpers.defaultAddress, "a" * 400) shouldBe Some(dataEntry)
+  }
+
```
yvaskov, line 56: Add negative tests (just in case): 
- key > 400 bytes
- value > 32767 bytes
- overall data > 165835 bytes (with keys <= 400 and data <= 32767)

yvaskov (changes requested):

yvaskov (approved):

phearnot (approved):

PR #3619, opened 2022-01-13
NODE-2416: Support underscores and dashes for directive values

xrtm000:


lang/shared/src/main/scala/com/wavesplatform/lang/directives/DirectiveParser.scala
```diff
@@ -20,7 +20,7 @@ object DirectiveParser {
       .!
 
   private def directiveValueP[_: P]: P[String] =
-    P(CharIn("a-zA-Z0-9/\\., "))
+    P(CharIn("a-zA-Z0-9_\\-/\\., "))
```
Karasiq, line 23: Add unit tests for it

Karasiq (changes requested):

lang/shared/src/main/scala/com/wavesplatform/lang/directives/DirectiveParser.scala
xrtm000: Done

yvaskov (approved):

phearnot (approved):

PR #3633, opened 2022-01-23
NODE-2427: Increase SetScript limit after V6

Karasiq:


yvaskov (approved):

node/src/test/scala/com/wavesplatform/state/diffs/SetScriptTransactionDiffTest.scala
```diff
@@ -62,12 +53,97 @@ class SetScriptTransactionDiffTest extends PropSpec with WithDomain {
...
+                                       |func test() = {
+                                       |  []
+                                       |}
+                                       |""".stripMargin)
+
+    withDomain(DomainPresets.RideV6) { d =>
+      val setScriptTransaction = TxHelpers.setScript(TxHelpers.defaultSigner, script, version = TxVersion.V2)
+      d.commonApi.calculateWavesFee(setScriptTransaction) shouldBe 0.001.waves
+    }
+  }
```
yvaskov, line 162: Add test for dynamic fee calculation:
- 1024 bytes should be 0.001 waves
- 1025 bytes should be 0.002 waves
- 32/160kb before/after V6 (max) - 0.032 and 0.16 waves respectively

yvaskov (changes requested):

yvaskov (approved):

phearnot (approved):

PR #3597, opened 2021-12-02
NODE-2438: Refactor error type system to avoid exceptions

xrtm000:

PR #3647, opened 2022-02-22
Portfolio errors as Either

xrtm000:


node/src/main/scala/com/wavesplatform/api/common/BalanceDistribution.scala
```diff
@@ -70,11 +70,17 @@ object BalanceDistribution {
...
-          if (currentHeight <= height && adjustedBalance > 0) Some(address -> adjustedBalance)
-          else findNextBalance()
+          if (adjustedBalanceE.isRight) {
+            val adjustedBalance = adjustedBalanceE.explicitGet()
+            if (currentHeight <= height && adjustedBalance > 0)
+              Some(address -> adjustedBalance)
+            else
+              findNextBalance()
+          } else
+            findNextBalance()
```
vsuharnikov, line 83: Minor, how about a pattern matching? :)
```scala
adjustedBalanceE match {
  case Right(adjustedBalance) if currentHeight <= height && adjustedBalance > 0 => Some(address -> adjustedBalance)
  case _ => findNextBalance()
}
```

node/src/main/scala/com/wavesplatform/api/http/DebugApiRoute.scala
```diff
@@ -221,33 +220,32 @@ case class DebugApiRoute(
...
       val transactionJson = parsedTransaction.fold(_ => jsv, _.json())
 
       val serializer = tracedDiff.resultE
-        .fold(_ => this.serializer, {
-          case (_, diff) =>
-            val compositeBlockchain = CompositeBlockchain(blockchain, diff)
-            this.serializer.copy(blockchain = compositeBlockchain)
-        })
+        .flatMap { case (_, diff, _) => blockchain.flatMap(CompositeBlockchain(_, diff)) }
+        .fold(_ => this.serializer, blockchain => this.serializer.copy(blockchain = blockchain))
```
vsuharnikov, line 237: This's interesting. Why can we ignore the fact that we aren't able to construct `CompositeBlockchain` (the first part of `fold`). This means we can't apply the `diff`, thus, the blockchain won't accept this transaction too.

It looks like we should return the error from this API in this case. Or add a comment, why this is not possible.

node/src/main/scala/com/wavesplatform/state/diffs/BlockDiffer.scala
```diff
@@ -53,34 +52,43 @@ object BlockDiffer extends ScorexLogging {
...
-        Portfolio.empty
+        Right(Portfolio.empty)
 
+    def wrap[A](r: Either[String, A]) = TracedResult(r).leftMap(GenericError(_))
     for {
-      _ <- TracedResult(Either.cond(!verify || block.signatureValid(), (), GenericError(s"Block $block has invalid signature")))
-      blockchainWithNewBlock = CompositeBlockchain(blockchain, Diff.empty, block, hitSource, 0, None)
+      _                       <- wrap(Either.cond(!verify || block.signatureValid(), (), s"Block $block has invalid signature"))
+      feeFromPreviousBlock    <- wrap(feeFromPreviousBlockE)
+      initialFeeFromThisBlock <- wrap(initialFeeFromThisBlockE)
```
vsuharnikov, line 82: Probably, `feeFromPreviousBlockE` and `initialFeeFromThisBlockE` should be lazy.

node/src/main/scala/com/wavesplatform/state/diffs/ExchangeTransactionDiff.scala
```diff
@@ -143,9 +142,11 @@ object ExchangeTransactionDiff {
       buyAmountAssetChange  <- getReceiveAmount(tx.buyOrder, amountDecimals, priceDecimals, tx.amount, tx.price)
       sellPriceAssetChange  <- getReceiveAmount(tx.sellOrder, amountDecimals, priceDecimals, tx.amount, tx.price)
       sellAmountAssetChange <- getSpendAmount(tx.sellOrder, amountDecimals, priceDecimals, tx.amount, tx.price).map(-_)
-      priceDiff  = getAssetDiff(tx.buyOrder.assetPair.priceAsset, buyPriceAssetChange, sellPriceAssetChange)
-      amountDiff = getAssetDiff(tx.buyOrder.assetPair.amountAsset, buyAmountAssetChange, sellAmountAssetChange)
-    } yield Monoid.combineAll(Seq(feeDiff, priceDiff, amountDiff))
+      priceDiff             <- getAssetDiff(tx.buyOrder.assetPair.priceAsset, buyPriceAssetChange, sellPriceAssetChange).leftMap(GenericError(_))
+      amountDiff            <- getAssetDiff(tx.buyOrder.assetPair.amountAsset, buyAmountAssetChange, sellAmountAssetChange).leftMap(GenericError(_))
+      feeDiff               <- feeDiffE.leftMap(GenericError(_))
```
vsuharnikov, line 147: Minor: here too. Probably, `matcherPortfolioE` and `feeDiffE` should be lazy.

node/src/main/scala/com/wavesplatform/state/diffs/MassTransferTransactionDiff.scala
```diff
@@ -29,28 +28,33 @@ object MassTransferTransactionDiff {
...
-      val completePortfolio =
-        recipientPortfolios
-          .combine(
-            tx.assetId
-              .fold(Map(sender -> Portfolio(-totalAmount, LeaseBalance.empty, Map.empty))) { asset =>
-                Map(sender -> Portfolio(0, LeaseBalance.empty, Map(asset -> -totalAmount)))
-              }
+      list
+        .foldM(foldInit) { (u, v) =>
+          Diff.combine(u._1, v._1).map((_, u._2 + v._2))
```
vsuharnikov, line 33: It'd good to see what we combine here instead of numbers. I know, it is an old code here, but it is hard to read and understand :)

node/src/main/scala/com/wavesplatform/state/diffs/invoke/InvokeScriptDiff.scala
```diff
@@ -197,39 +199,35 @@ object InvokeScriptDiff {
...
             }
             _ = invocationRoot.setLog(log)
 
-            _ = if (blockchain.height >= blockchain.settings.functionalitySettings.syncDAppCheckTransfersHeight)
-              checkDiffBalances(diff, blockchain)
+            blockchainAfterEvaluation <- traced(CompositeBlockchain(blockchain, diff))
 
-            newBlockchain = CompositeBlockchain(blockchain, diff)
+            _ = if (blockchain.height >= blockchain.settings.functionalitySettings.syncDAppCheckTransfersHeight)
+              checkDiffBalances(diff, blockchainAfterEvaluation)
```
vsuharnikov, line 223: I see it throws an exception. I think, this is not good to mix exceptions and Either. But let it be for now.

node/src/main/scala/com/wavesplatform/state/reader/CompositeBlockchain.scala
```diff
@@ -12,35 +12,48 @@ import com.wavesplatform.lang.ValidationError
...
   override def balance(address: Address, assetId: Asset): Long =
-    inner.balance(address, assetId) + diff.portfolios.getOrElse(address, Portfolio.empty).balanceOf(assetId)
+    compositePortfolios
+      .get(address)
+      .flatMap(
+        p =>
+          assetId match {
+            case asset: IssuedAsset => p.assets.get(asset)
+            case Waves              => Some(p.balance)
+        }
```
vsuharnikov, line 40: I saw this patter multiple times. I think, we need a special method in `Portfolio` (if it hasn't)

vsuharnikov (changes requested):

node/src/main/scala/com/wavesplatform/api/common/BalanceDistribution.scala
xrtm000: Done

node/src/main/scala/com/wavesplatform/api/http/DebugApiRoute.scala
xrtm000: Not actual currently

node/src/main/scala/com/wavesplatform/state/diffs/BlockDiffer.scala
xrtm000: Why? It seems be used almost always.

node/src/main/scala/com/wavesplatform/state/diffs/ExchangeTransactionDiff.scala
xrtm000: Done

node/src/main/scala/com/wavesplatform/state/diffs/MassTransferTransactionDiff.scala
xrtm000: Done

node/src/main/scala/com/wavesplatform/state/diffs/invoke/InvokeScriptDiff.scala
xrtm000: Refactored in #3597

node/src/main/scala/com/wavesplatform/state/reader/CompositeBlockchain.scala
xrtm000: Is it actual at the moment?)

vsuharnikov: We could add a task to backlog

PR #3704, opened 2022-05-17
NODE-2329 Reestabilish input stream on unexpected read bytes size

xrtm000:


yvaskov (approved):

ivan-mashonskiy (approved):

yvaskov (approved):

node/src/main/scala/com/wavesplatform/Importer.scala
```diff
@@ -188,58 +189,68 @@ object Importer extends ScorexLogging {
...
+                }
+              } else {
+                log.warn(s"Block $block is not a child of the last block ${blockchain.lastBlockId.get}")
+              }
             }
+          } else {
+            log.info(s"$factReadSize != expected $blockSize")
+            log.info(s"reestablishing input stream")
+            inputStream.close()
+            inputStream = getInputStream()
```
vsuharnikov, line 244: Just a notice: actually we have `blockSize` for each block, so we don't need to read whole `BlockInfoAtHeight` sequence to find the offset.

vsuharnikov (approved):

PR #3714, opened 2022-05-27
NODE-2474 Improve generator base target calculation

xrtm000:


ivan-mashonskiy (approved):

yvaskov (changes requested):
~~Specifying `initial-base-target` for `GenesisBlockGenerator` doesn't work. Keeps calculating its own base target~~
Docs bug

yvaskov (approved):

yvaskov (changes requested):
Needs a fix to account for `miner` flag in ``DistributionItem(seedText: String, nonce: Int = 0, amount: Share, miner: Boolean = true)`` while calculating initial base target

yvaskov (approved):

phearnot (approved):

PR #3726, opened 2022-06-22
NODE-2291 Add request span metrics

ivan-mashonskiy:


node/build.sbt
```diff
@@ -1,9 +1,26 @@
...
+  GitVersioning,
+  VersionObject,
+  JavaAgent
+)
 
 libraryDependencies ++= Dependencies.node.value
 
+javaAgents ++= {
+  if (instrumentation.value) {
+    Seq("io.kamon" % "kanela-agent" % "1.0.14")
```
vsuharnikov: Probably, `Dependencies` is a better place for this

node/src/main/scala/com/wavesplatform/Application.scala
```diff
@@ -327,19 +336,45 @@ class Application(val actorSystem: ActorSystem, val settings: WavesSettings, con
           "rest-time-limited",
           reporter = log.trace("Uncaught exception in time limited pool", _)
         )
+      val heavyRequestProcessorPoolThreads =
+        settings.restAPISettings.heavyRequestProcessorPoolThreads.getOrElse((Runtime.getRuntime.availableProcessors() * 2).min(4))
```
vsuharnikov, line 340: Could you add a comment, why do we need ~at least~ maximum 4 threads?

node/src/main/scala/com/wavesplatform/Application.scala
```diff
@@ -327,19 +336,45 @@ class Application(val actorSystem: ActorSystem, val settings: WavesSettings, con
...
+      val heavyRequestExecutor = new ThreadPoolExecutor(
+        heavyRequestProcessorPoolThreads,
+        heavyRequestProcessorPoolThreads,
+        60,
+        TimeUnit.SECONDS,
```
vsuharnikov: I know, these values doesn't affect a number of threads in the pool in this case, but probably it'd be better to specify 0 millis as `Executors.newFixedThreadPool` does:
```java
public static ExecutorService newFixedThreadPool(int nThreads, ThreadFactory threadFactory) {
    return new ThreadPoolExecutor(nThreads, nThreads,
                                  0L, TimeUnit.MILLISECONDS,
                                  new LinkedBlockingQueue<Runnable>(),
                                  threadFactory);
}
```

node/src/main/scala/com/wavesplatform/api/http/AddressApiRoute.scala
```diff
@@ -287,19 +286,22 @@ case class AddressApiRoute(
...
+    )(_.runAsyncLogErr(_))
+  }
 
-  private def accountData(addr: Address, regex: String)(implicit sc: Scheduler) =
-    commonAccountsApi
-      .dataStream(addr, Some(regex))
-      .toListL
-      .runAsyncLogErr
-      .map(data => Source.fromIterator(() => data.sortBy(_.key).iterator.map(Json.toJson[DataEntry[?]])))
+  private def accountData(addr: Address, regex: String)(implicit m: ToResponseMarshaller[Source[JsValue, NotUsed]]) =
```
vsuharnikov: The only difference of these two `accountData` methods is `regex: String`. How about:
```scala
private def accountData(addr: Address, regex: Option[String] = None)(implicit m: ToResponseMarshaller[Source[JsValue, NotUsed]])
```
?

node/src/main/scala/com/wavesplatform/api/http/BlocksApiRoute.scala
```diff
@@ -17,8 +17,9 @@ import com.wavesplatform.transaction.TxValidationError.GenericError
 import com.wavesplatform.utils.Time
 import play.api.libs.json.*
 
-case class BlocksApiRoute(settings: RestAPISettings, commonApi: CommonBlocksApi, time: Time) extends ApiRoute {
+case class BlocksApiRoute(settings: RestAPISettings, commonApi: CommonBlocksApi, time: Time, routeTimeout: RouteTimeout) extends ApiRoute {
   import BlocksApiRoute.*
+  private[this] val MaxBlocksPerRequest = 100 // todo: make this configurable and fix integration tests
```
vsuharnikov: I remind you about this todo :)

node/src/main/scala/com/wavesplatform/Application.scala
xrtm000: And why doubled available processors?

node/src/main/scala/com/wavesplatform/Application.scala
```diff
@@ -390,19 +428,36 @@ class Application(val actorSystem: ActorSystem, val settings: WavesSettings, con
...
+        ),
         RewardApiRoute(blockchainUpdater)
       )
 
       val httpService = CompositeHttpService(apiRoutes, settings.restAPISettings)
       val httpFuture =
         Http().newServerAt(settings.restAPISettings.bindAddress, settings.restAPISettings.port).bindFlow(httpService.loggingCompositeRoute)
       serverBinding = Await.result(httpFuture, 20.seconds)
-      serverBinding.whenTerminated.foreach(_ => httpService.scheduler.shutdown())
+      serverBinding.whenTerminated.foreach(_ => heavyRequestScheduler.shutdown())
```
xrtm000, line 460: Why `httpService.scheduler` should not be shutdowned?

node/src/main/scala/com/wavesplatform/api/common/CommonAccountsApi.scala
```diff
@@ -57,9 +57,8 @@ object CommonAccountsApi {
 
   def apply(diff: () => Diff, db: DB, blockchain: Blockchain): CommonAccountsApi = new CommonAccountsApi {
 
-    override def balance(address: Address, confirmations: Int = 0): Long = {
-      blockchain.balance(address, blockchain.height, confirmations)
-    }
+    override def balance(address: Address, confirmations: Int = 0): Long =
+      blockchain.balance(address)
```
xrtm000: Why confirmations should be unused?

DrBlast (approved):

xrtm000 (approved):

vsuharnikov (approved):

phearnot (approved):

PR #3738, opened 2022-07-25
Fix eth_call response

phearnot:


yvaskov (approved):

yvaskov (approved):

xrtm000 (approved):

vsuharnikov (approved):

node/src/main/scala/com/wavesplatform/api/http/eth/EthRpcRoute.scala
```diff
@@ -160,10 +158,10 @@ class EthRpcRoute(blockchain: Blockchain, transactionsApi: CommonTransactionsApi
                   )
                 )
               )
-            case "01ffc9a7" => // supportsInterface()
+            case "01ffc9a7" => // supportsInterface() https://eips.ethereum.org/EIPS/eip-165
```
nickeskov, line 161: Btw, correct signature which gives selector `0x01ffc9a7` is `supportsInterface(bytes4)`.

phearnot: Fair enough!

PR #3748, opened 2022-08-17
NODE-2348 Consistent API liquid state

xrtm000:


node/src/main/scala/com/wavesplatform/state/reader/CompositeBlockchain.scala
```diff
@@ -95,7 +95,7 @@ final class CompositeBlockchain private (
     }
 
   override def balanceSnapshots(address: Address, from: Int, to: Option[BlockId]): Seq[BalanceSnapshot] =
-    if (maybeDiff.isEmpty || to.exists(id => inner.heightOf(id).isDefined)) {
+    if (maybeDiff.isEmpty || to.exists(id => inner.heightOf(id).exists(_ != height))) {
```
vsuharnikov, line 95: Could you add a comment here, why do we need to check the `height`?

vsuharnikov (approved):

node/src/main/scala/com/wavesplatform/state/reader/CompositeBlockchain.scala
xrtm000: The height of the composite blockchain can match the height of the nested blockchain if there is no block in the composite one.

yvaskov (approved):

ivan-mashonskiy (approved):

PR #3757, opened 2022-09-05
NODE-2402 Invocation structure for /utils/evaluate/ 

xrtm000:


node/src/main/scala/com/wavesplatform/api/http/utils/UtilsInvocationRequest.scala
```diff
@@ -0,0 +1,47 @@
...
+case class UtilsInvocationRequest(
+    call: Option[FunctionCallPart],
+    id: Option[String],
+    fee: Option[Long],
+    feeAssetId: Option[String],
+    sender: Option[String],
+    senderPublicKey: Option[String],
+    payment: Option[Seq[Payment]]
+) {
+  def toInvocation: Either[ValidationError, Invocation] =
```
vsuharnikov, line 29: You can use default values for some fields, see https://scastie.scala-lang.org/Ck1UwdqATkqBfF0Kq2YQow

node/src/main/scala/com/wavesplatform/api/http/utils/UtilsInvocationRequest.scala
```diff
@@ -0,0 +1,47 @@
...
+    for {
+      senderPK <- senderPublicKey.map(PublicKey.fromBase58String).getOrElse(Right(ByteStr(new Array[Byte](32))))
+      id       <- id.map(decodeBase58).getOrElse(Right(ByteStr(new Array[Byte](32))))
+      functionCall = call.map(InvokeScriptRequest.buildFunctionCall).getOrElse(InvokeTransaction.DefaultCall)
+      feeAssetId <- feeAssetId.traverse(decodeBase58)
+      sender <- sender
+        .map(AddressOrAlias.fromString(_, checkChainId = false).map(a => Address(ByteStr(a.bytes))))
```
vsuharnikov: We're trying to parse an address or alias, but we convert the result to an address, even if we get an alias :thinking: This doesn't look like it will work. Did you want `Address.fromString`?

vsuharnikov (changes requested):

node/src/main/scala/com/wavesplatform/api/http/utils/UtilsEvaluator.scala
```diff
@@ -0,0 +1,103 @@
...
+
+  def executeExpression(blockchain: Blockchain, script: Script, address: Address, pk: PublicKey, limit: Int)(
+      expr: EXPR
+  ): Either[ValidationError, (EVALUATED, Int, Log[Id])] =
+    for {
+      ds <- DirectiveSet(script.stdLibVersion, Account, DAppType).leftMap(GenericError(_))
+      invoke = new InvokeScriptTransactionLike {
+        override def dApp: AddressOrAlias              = address
+        override def funcCall: Terms.FUNCTION_CALL     = Terms.FUNCTION_CALL(FunctionHeader.User(""), Nil)
+        override def payments: Seq[Payment]            = Seq.empty
```
vsuharnikov, line 54: Then how do payments work?

node/src/main/scala/com/wavesplatform/api/http/utils/UtilsEvaluator.scala
```diff
@@ -0,0 +1,103 @@
...
+        override def root: InvokeScriptTransactionLike = this
+        override val sender: PublicKey                 = PublicKey(ByteStr(new Array[Byte](32)))
+        override def assetFee: (Asset, Long)           = Asset.Waves -> 0L
+        override def timestamp: Long                   = System.currentTimeMillis()
+        override def chainId: Byte                     = AddressScheme.current.chainId
+        override def id: Coeval[ByteStr]               = Coeval.evalOnce(ByteStr.empty)
+        override def checkedAssets: Seq[IssuedAsset]   = Seq.empty
```
vsuharnikov, line 61: This actually is a sequence of scripted assets. So, their scripts won't run.
Am I right?

vsuharnikov: Will be fixed in another task

node/src/main/scala/com/wavesplatform/api/http/utils/UtilsEvaluator.scala
vsuharnikov, line 54: Could you add a comment, that payments are propagated in `Invocation`?

ivan-mashonskiy (approved):

yvaskov (approved):

node/src/main/scala/com/wavesplatform/api/http/utils/UtilsEvaluator.scala
xrtm000: Payments, that are mapped to RIDE structure, is taken from [Invocation](https://github.com/wavesplatform/Waves/blob/version-1.4.x/lang/shared/src/main/scala/com/wavesplatform/lang/v1/evaluator/ctx/impl/waves/Bindings.scala#L131), while field, that you noticed, used for validation inside [InvokeScriptTransactionDiff](https://github.com/wavesplatform/Waves/blob/version-1.4.x/node/src/main/scala/com/wavesplatform/state/diffs/invoke/InvokeScriptTransactionDiff.scala), that unused in the current implementation.

vsuharnikov: I meant to add this comment to the code.

yvaskov (approved):

vsuharnikov (approved):

phearnot (approved):

PR #3759, opened 2022-09-07
NODE-2501 Extended log info for nested dApp invocations

ivan-mashonskiy:


lang/shared/src/main/scala/com/wavesplatform/lang/v1/compiler/Terms.scala
```diff
@@ -147,8 +147,8 @@ object Terms {
   }
 
   sealed trait EVALUATED extends EXPR {
-    def prettyString(level: Int): String = toString
-    def toStr: Coeval[String]            = Coeval.now(toString)
+    def prettyString(level: Int, fixArrIndentation: Boolean = false): String = toString
```
vsuharnikov: This isn't a good idea to have in a base class things, those are specific to a subset of descendants. I suggest to move all `toString` and `prettyString`-related code to `TermPrinter`.

lang/shared/src/main/scala/com/wavesplatform/lang/v1/evaluator/EvaluatorV2.scala
```diff
@@ -369,17 +383,44 @@ object EvaluatorV2 {
...
       expr,
+      logExtraInfo,
       stdLibVersion,
       Int.MaxValue,
       correctFunctionCallScope,
       newMode,
       expr => Left(s"Unexpected incomplete evaluation result $expr")
     )
+
+  private def enrichLog(loggedCtx: LoggedEvaluationContext[Environment, Id], logExtraInfo: LogExtraInfo, exprCopy: EXPR) = {
```
vsuharnikov: Probably, `logCalls` describes this function better.

yvaskov (approved):

xrtm000 (approved):

vsuharnikov (approved):

phearnot (approved):

PR #3765, opened 2022-09-23
Bloom filter in Diff

phearnot:


node/src/main/scala/com/wavesplatform/state/Diff.scala
```diff
@@ -145,34 +146,52 @@ case class NewTransactionInfo(transaction: Transaction, affected: Set[Address],
...
+
   def combineF(newer: Diff): Either[String, Diff] =
     Diff
       .combine(portfolios, newer.portfolios)
-      .map(portfolios =>
+      .map { portfolios =>
+        val newTransactions = if (transactions.isEmpty) newer.transactions else transactions ++ newer.transactions
+        val newFilter = transactionFilter match {
+          case Some(bf) =>
+            newer.transactions.foreach(nti => bf.put(nti.transaction.id().arr))
```
xrtm000, line 190: Isn't it worth to compare sizes of `this.transactions` and `newer.transactions` to choose bigger bloom filter as the base?

phearnot: We mostly append a diff for a single transaction to an accumulator (e.g. in TransactionDiffer and UtxPoolImpl). Can't think of a place where `newer` has more transactions than `this`.

DrBlast (approved):

PR #3793, opened 2022-12-09
NODE-2425 Optimization of JS compile methods

xrtm000:


yvaskov (approved):

lang/js/src/main/scala/com/wavesplatform/JsApiUtils.scala
```diff
@@ -110,65 +109,56 @@ object JsApiUtils {
 
       case x: Expressions.REF =>
         val additionalDataObj = jObj("name" -> Expressions.PART.toOption[String](x.key).getOrElse("").toString)
-        mergeJSObjects(commonDataObj, additionalDataObj)
+        js.Object.assign(commonDataObj, additionalDataObj)
```
vsuharnikov: If I got it right, you've preserved the arguments order to make less operations: copy only `sizeOf(additionalDataObj)==1` property instead of `sizeOf(commonDataObj)==5` ?
But this doesn't explain, why you changed the order in other places. Also I see that there is no key intersection of two this objects  :thinking:

ivan-mashonskiy (approved):

lang/js/src/main/scala/com/wavesplatform/JsApiUtils.scala
xrtm000: `js.Object.assign(a, b)` mutates the first argument `a`, so reusable `commonDataObj` should always be the second.

xrtm000: Fix has been uploaded

yvaskov (approved):

phearnot (approved):

PR #3824, opened 2023-03-21
NODE-2567 State snapshot hash computation

ivan-mashonskiy:


node/src/main/scala/com/wavesplatform/serialization/package.scala
```diff
@@ -61,6 +61,14 @@ package object serialization {
       result
     }
 
+    def getByteArrayOpt(size: Int): Option[Array[Byte]] = {
+      if (buf.limit() - buf.position() >= size) {
```
vsuharnikov: `buf.remaining() >= size` ?

vsuharnikov (approved):

yvaskov (changes requested):
Add test for genesis block state hash computation
Add asset decimals to existing tests
Check feature activation block state hash

yvaskov (approved):

xrtm000 (approved):

phearnot (approved):

PR #3829, opened 2023-03-28
Common isWellFormed() implementation

xrtm000:


yvaskov (approved):

ivan-mashonskiy (approved):

lang/shared/src/main/scala/com/wavesplatform/lang/package.scala
```diff
@@ -20,4 +20,19 @@ package object lang {
...
+      var i = 0
+      while (i < s.length) {
+        val c = s.charAt(i)
+        if (Character.isSurrogate(c)) {
+          if (s.codePointAt(i) == c) return false
+          i += 1
+        }
+        i += 1
+      }
+      true
```
vsuharnikov: How about not using `return` (because it is generally weird in Scala), e.g.
```scala
  def isWellFormed: Boolean = {
    var i = 0
    var wellFormed = true
    while (i < s.length && wellFormed) {
      val c = s.charAt(i)
      if (Character.isSurrogate(c)) {
        if (s.codePointAt(i) == c) wellFormed = false
        else i += 1
      }
      i += 1
    }
    wellFormed
  }
```
?

xrtm000: +

vsuharnikov (approved):

lang/shared/src/main/scala/com/wavesplatform/lang/package.scala
vsuharnikov: UP: it works via exceptions in a case of non local returns, but generally it's better to avoid `return`

phearnot (approved):
Import up to height 3735314 revealed no issues.

PR #3840, opened 2023-04-25
NODE-2585 Distribute block reward between several addresses

ivan-mashonskiy:


node/src/main/scala/com/wavesplatform/settings/package.scala
```diff
@@ -49,6 +50,14 @@ package object settings {
 
   implicit val privateKeyReader: ValueReader[PrivateKey] = byteStrReader.map(PrivateKey(_))
 
+  implicit val addressReader: ValueReader[Address] =
```
vsuharnikov: Not sure this works for custom networks. In `Application` we load a config, parse settings and only then set a network byte:
```scala
// Application.scala
    val settings = WavesSettings.fromRootConfig(config) // (1)

    // Initialize global var with actual address scheme
    AddressScheme.current = new AddressScheme {
      override val chainId: Byte = settings.blockchainSettings.addressSchemeCharacter.toByte
    }
```
So we should get a network byte and setup the address scheme before we parse settings (1).

vsuharnikov (changes requested):

node/src/main/scala/com/wavesplatform/settings/package.scala
ivan-mashonskiy: Fixed

xrtm000 (approved):

DrBlast (approved):

yvaskov (approved):

phearnot (approved):

PR #3859, opened 2023-06-28
NODE-2594 Block state hash challenge

ivan-mashonskiy:


xrtm000 (approved):

grpc-server/src/main/scala/com/wavesplatform/api/grpc/TransactionsApiGrpcImpl.scala
```diff
@@ -102,7 +101,12 @@ class TransactionsApiGrpcImpl(blockchain: Blockchain, commonApi: CommonTransacti
           .map(_ => TransactionStatus(txId, TransactionStatus.Status.UNCONFIRMED))
           .orElse {
             commonApi.transactionById(txId.toByteStr).map { m =>
-              val status = if (m.succeeded) ApplicationStatus.SUCCEEDED else ApplicationStatus.SCRIPT_EXECUTION_FAILED
+              val status = m.status match {
+                case TxMeta.Status.Succeeded => ApplicationStatus.SUCCEEDED
+                case TxMeta.Status.Failed    => ApplicationStatus.SCRIPT_EXECUTION_FAILED
+                case TxMeta.Status.Elided    => ApplicationStatus.ELIDED
+              }
```
vsuharnikov: This appears 5 times. Let's create a function for this.

node/src/main/protobuf/waves/database.proto
```diff
@@ -34,7 +40,7 @@ message TransactionMeta {
   int32 height = 1;
   int32 num = 2;
   int32 type = 3;
-  bool failed = 4;
+  Status status = 4;
```
vsuharnikov, line 43: Interesting. Can we have here some kind of backward compatibility so probably we don't need to rebuild a state? I mean:
1. Can we encode `FAILED` as 1
```scala
  @SerialVersionUID(0L)
  case object FAILED extends Status(0) with Status.Recognized {
    val index = 0
```
`SUCCEEDED` as 0 and `ELIDED` as 2?
2. Will we have same bytes in protobuf for `FAILED` and `SUCCEEDED` as it was `bool failed` (I hope so)?

node/src/main/scala/com/wavesplatform/mining/BlockChallenger.scala
```diff
@@ -0,0 +1,277 @@
...
+    (for {
+      discarded <- EitherT(Task(blockchainUpdater.removeAfter(blockchainUpdater.lastBlockHeader.get.header.reference)))
+      block     <- EitherT(Task(discarded.headOption.map(_._1).toRight(GenericError("Liquid block wasn't discarded"))))
+      _ <- EitherT.liftF(Task(processingTxs.putAll((block.transactionData ++ md.microBlock.transactionData).map(tx => tx.id() -> tx).toMap.asJava)))
+      challengingBlock <- EitherT(
+        createChallengingBlock(
+          block,
+          md.microBlock.stateHash,
+          md.microBlock.totalResBlockSig,
+          block.transactionData ++ md.microBlock.transactionData,
```
vsuharnikov: Probably, we should save `block.transactionData ++ md.microBlock.transactionData` in a variable `txs` and use it twice

node/src/main/scala/com/wavesplatform/mining/BlockChallenger.scala
```diff
@@ -0,0 +1,277 @@
...
+        createChallengingBlock(block, block.header.stateHash, block.signature, block.transactionData, prevStateHash, blockReward)
+      )
+      _ <- EitherT(appendBlock(challengingBlock).asyncBoundary)
+      _ <- EitherT.liftF[Task, ValidationError, Unit](Task(processingTxs.clear()))
+    } yield {
+      log.debug(s"Successfully challenged $block with $challengingBlock")
+      allChannels.broadcast(BlockForged(challengingBlock), Some(ch))
+    }).fold(
+      err => {
+        processingTxs.clear()
```
vsuharnikov: There (and in `challengeMicroblock` too) could be `Task.bracket` to guarantee removing transactions from `processingTxs` even an error happened, e.g.:
```scala
  def challengeBlock(block: Block, ch: Channel, prevStateHash: ByteStr, blockReward: Option[Long]): Task[Unit] = {
    log.debug(s"Challenging block $block")

    Task(processingTxs.putAll(block.transactionData.map(tx => tx.id() -> tx).toMap.asJava)).bracket { _ =>
      (for {
        challengingBlock <- EitherT(
          createChallengingBlock(block, block.header.stateHash, block.signature, block.transactionData, prevStateHash, blockReward)
        )
        _ <- EitherT(appendBlock(challengingBlock).asyncBoundary)
      } yield {
        log.debug(s"Successfully challenged $block with $challengingBlock")
        allChannels.broadcast(BlockForged(challengingBlock), Some(ch))
      }).fold(
        err => log.debug(s"Could not challenge $block: $err"),
        identity
      )
    }(_ => Task(processingTxs.clear()))
  }
```

The next improvement is to save transactions and remove only them. But I'm not sure are concurrent challenges possible or not.

What do you think?

node/src/main/scala/com/wavesplatform/mining/BlockChallenger.scala
```diff
@@ -0,0 +1,277 @@
...
+        log.debug(s"Could not challenge microblock $idStr: $err")
+      },
+      identity
+    )
+  }
+
+  def pickBestAccount(accounts: Seq[(SeedKeyPair, Long)]): Either[GenericError, (SeedKeyPair, Long)] =
+    accounts.minByOption(_._2).toRight(GenericError("No suitable account in wallet"))
+
+  def getChallengingAccounts(challengedMiner: Address): Either[ValidationError, Seq[(SeedKeyPair, Long)]] =
```
vsuharnikov, line 127: It seems, `pickBestAccount` and `getChallengingAccounts` always are used both. Probably, it'd better to make `def pickBestChallengingAccount(challengedMiner: Address): Either[GenericError, (SeedKeyPair, Long)]` ?
Also, I'm not sure we should make `pickBestAccount` and `getChallengingAccounts` public (or even this new method). We use them only in `BlockChallenger`.

node/src/main/scala/com/wavesplatform/state/BlockchainUpdaterImpl.scala
```diff
@@ -674,9 +689,17 @@ class BlockchainUpdaterImpl(
...
+      ngState.flatMap(_.totalDiffOf(id))
+    }.fold[Blockchain](rocksdb) { case (block, diff, _, _, _) =>
+      CompositeBlockchain(rocksdb, diff, block, ByteStr.empty, 0L, None)
+    }.balanceSnapshots(address, from, to)
+
+//    to.flatMap(id => ngState.flatMap(_.totalDiffOf(id)))
+//      .fold[Blockchain](rocksdb) { case (block, diff, _, _, _) =>
+//        CompositeBlockchain(rocksdb, diff, block, ByteStr.empty, 0L, None)
+//      }
+//      .balanceSnapshots(address, from, to)
```
vsuharnikov: ?

node/src/main/scala/com/wavesplatform/state/diffs/BlockDiffer.scala
```diff
@@ -104,26 +135,34 @@ object BlockDiffer {
         totalReward             <- minerReward.combine(initialFeeFromThisBlock).flatMap(_.combine(feeFromPreviousBlock))
         patches                 <- patchesDiff(blockchainWithNewBlock)
         resultDiff              <- Diff(portfolios = Map(block.sender.toAddress -> totalReward)).combineF(patches)
-      } yield resultDiff
+      } yield resultDiff.withPortfolios(resultDiff.portfolios.filter(!_._2.isEmpty))
 
     for {
       _          <- TracedResult(Either.cond(!verify || block.signatureValid(), (), GenericError(s"Block $block has invalid signature")))
       resultDiff <- TracedResult(initDiffE.leftMap(GenericError(_)))
+      // TODO: correctly obtain previous state hash on feature activation height
```
vsuharnikov, line 143: TODO. Do we need a separate task or implementation in this PR? If we need a task, please create it.

node/src/main/scala/com/wavesplatform/mining/BlockChallenger.scala
ivan-mashonskiy: It was done for testing purposes: https://github.com/wavesplatform/Waves/pull/3859/files#diff-0696e3ee17bb75c4e1bc89d88cc58f915014d9af7d321678d6eb7ed3c0cd1f18R178-R180

node/src/main/protobuf/waves/database.proto
ivan-mashonskiy: This changes appears only in 1.5 version, which anyway requires state rebuilding at least due to a change in used storage and some scheme improvements

vsuharnikov (commented):
The main question is `processingTxs.clear`, the rest issues are cosmetic

grpc-server/src/main/scala/com/wavesplatform/api/grpc/TransactionsApiGrpcImpl.scala
ivan-mashonskiy: Fixed

node/src/main/scala/com/wavesplatform/mining/BlockChallenger.scala
ivan-mashonskiy: Fixed

node/src/main/scala/com/wavesplatform/mining/BlockChallenger.scala
ivan-mashonskiy: Fixed
Concurrent challenges are not possible

node/src/main/scala/com/wavesplatform/state/BlockchainUpdaterImpl.scala
ivan-mashonskiy: Fixed

vsuharnikov (approved):

yvaskov (approved):

phearnot (approved):

PR #3874, opened 2023-09-01
NODE-2609 Light node implementation

ivan-mashonskiy:


xrtm000 (approved):

node/src/main/scala/com/wavesplatform/api/http/TransactionsApiRoute.scala
```diff
@@ -172,9 +174,13 @@ case class TransactionsApiRoute(
...
 
-  def signedBroadcast: Route = path("broadcast")(
-    broadcast[JsValue](TransactionFactory.fromSignedRequest)
-  )
+  def signedBroadcast: Route = path("broadcast") {
+    if (isLightMode) {
+      complete(StatusCodes.NotImplemented, CustomValidationError("Transaction broadcast in not supported for light node").json)
+    } else {
+      broadcast[JsValue](TransactionFactory.fromSignedRequest)
+    }
```
vsuharnikov: Let's enable the transactions broadcasting for light nodes, it is useful and nothing prevents from doing this

node/src/main/scala/com/wavesplatform/database/RocksDBWriter.scala
```diff
@@ -622,14 +629,14 @@ class RocksDBWriter(
     log.trace(s"Finished persisting block ${blockMeta.id} at height $height")
   }
 
-  override protected def doRollback(targetHeight: Int): Seq[(Block, ByteStr)] = {
+  override protected def doRollback(targetHeight: Int): Seq[(Block, ByteStr, Option[BlockSnapshot])] = {
```
vsuharnikov: There is a type `DiscardedBlocks = (Block, ByteStr, Option[BlockSnapshot])`. 
I still prefer case classes for public methods, because IDE always tries to expand type aliases.
But type aliases are good too. They help searching usages in IDE.

node/src/main/scala/com/wavesplatform/database/RocksDBWriter.scala
```diff
@@ -1078,4 +1090,8 @@ class RocksDBWriter(
 
   override def resolveERC20Address(address: ERC20Address): Option[IssuedAsset] =
     readOnly(_.get(Keys.assetStaticInfo(address)).map(assetInfo => IssuedAsset(assetInfo.id.toByteStr)))
+
+  override def prevStateHash(refId: Option[ByteStr]): ByteStr = {
```
vsuharnikov: Actually, it looks like `lastStateHash`, because we look on the current height or on a specified liquid block. Isn't it?

node/src/main/scala/com/wavesplatform/Application.scala
```diff
@@ -121,26 +121,16 @@ class Application(val actorSystem: ActorSystem, val settings: WavesSettings, con
 
     val establishedConnections = new ConcurrentHashMap[Channel, PeerInfo]
     val allChannels            = new DefaultChannelGroup(GlobalEventExecutor.INSTANCE)
-    val utxStorage =
+    val utxStorage = if (!settings.enableLightMode) {
       new UtxPoolImpl(time, blockchainUpdater, settings.utxSettings, settings.maxTxErrorLogSize, settings.minerSettings.enable, utxEvents.onNext)
+    } else UtxPool.NoOp
```
vsuharnikov: Just wondering why don't write instead:
```
val utxStorage = if (settings.enableLightMode) UtxPool.NoOp else {
  new UtxPoolImpl(time, blockchainUpdater, settings.utxSettings, settings.maxTxErrorLogSize, settings.minerSettings.enable, utxEvents.onNext)
}
```
? When "if" branch has a negation, thinking about the "else" branch breaks a brain, because you have to think about a double negation (`!!settings.enableLightMode`).

There must be the reason, and I would like to know it.

node/src/main/scala/com/wavesplatform/network/DiscardingHandler.scala
```diff
@@ -7,15 +7,20 @@ import monix.execution.schedulers.SchedulerService
...
   private implicit val scheduler: SchedulerService = Schedulers.fixedPool(2, "discarding-handler")
   private val lastReadiness                        = lastObserved(blockchainReadiness)
 
   override def channelRead(ctx: ChannelHandlerContext, msg: AnyRef): Unit = msg match {
-    case RawBytes(code @ (TransactionSpec.messageCode | PBTransactionSpec.messageCode), _) if !lastReadiness().contains(true) =>
-      log.trace(s"${id(ctx)} Discarding incoming message $code")
+    case RawBytes(code @ (TransactionSpec.messageCode | PBTransactionSpec.messageCode), _) if !lastReadiness().contains(true) || isLightMode =>
+      logDiscarding(ctx, code)
+    case RawBytes(code @ (BlockSnapshotResponseSpec.messageCode | MicroBlockSnapshotResponseSpec.messageCode), _) if !isLightMode =>
+      logDiscarding(ctx, code)
```
vsuharnikov, line 19: Do we really need to log this? I mean this is normal behavior of a full node.

node/src/main/scala/com/wavesplatform/network/DiscardingHandler.scala
```diff
@@ -7,15 +7,20 @@ import monix.execution.schedulers.SchedulerService
...
-class DiscardingHandler(blockchainReadiness: Observable[Boolean]) extends ChannelDuplexHandler with ScorexLogging {
+class DiscardingHandler(blockchainReadiness: Observable[Boolean], isLightMode: Boolean) extends ChannelDuplexHandler with ScorexLogging {
 
   private implicit val scheduler: SchedulerService = Schedulers.fixedPool(2, "discarding-handler")
   private val lastReadiness                        = lastObserved(blockchainReadiness)
 
   override def channelRead(ctx: ChannelHandlerContext, msg: AnyRef): Unit = msg match {
-    case RawBytes(code @ (TransactionSpec.messageCode | PBTransactionSpec.messageCode), _) if !lastReadiness().contains(true) =>
-      log.trace(s"${id(ctx)} Discarding incoming message $code")
+    case RawBytes(code @ (TransactionSpec.messageCode | PBTransactionSpec.messageCode), _) if !lastReadiness().contains(true) || isLightMode =>
```
vsuharnikov: I think we need to re-broadcast transactions even this is a light node, because imagine, you are a miner and all your connections are light nodes. This could happen, because the light node is an appropriate choice for most projects on blockchain (hope there will be much more projects 🤞).

node/src/main/scala/com/wavesplatform/network/MessageCodec.scala
```diff
@@ -22,11 +22,16 @@ class MessageCodec(peerDatabase: PeerDatabase) extends MessageToMessageCodec[Raw
     case BlockForged(b)           => out.add(RawBytes.fromBlock(b))
 
     // With a spec
-    case GetPeers             => out.add(RawBytes(GetPeersSpec.messageCode, Array[Byte]()))
-    case k: KnownPeers        => out.add(RawBytes(PeersSpec.messageCode, PeersSpec.serializeData(k)))
-    case g: GetBlock          => out.add(RawBytes(GetBlockSpec.messageCode, GetBlockSpec.serializeData(g)))
-    case m: MicroBlockInv     => out.add(RawBytes(MicroBlockInvSpec.messageCode, MicroBlockInvSpec.serializeData(m)))
-    case m: MicroBlockRequest => out.add(RawBytes(MicroBlockRequestSpec.messageCode, MicroBlockRequestSpec.serializeData(m)))
+    case GetPeers                 => out.add(RawBytes(GetPeersSpec.messageCode, Array[Byte]()))
+    case k: KnownPeers            => out.add(RawBytes(PeersSpec.messageCode, PeersSpec.serializeData(k)))
```
vsuharnikov: How about:
```scala
object RawBytes {
  def from[T <: AnyRef](spec: MessageSpec[T], message: T): RawBytes = RawBytes(spec.messageCode, spec.serializeData(message))
  // ...
}

RawBytes.from(PeersSpec, k)
``` 
? 🤔

vsuharnikov: Also you might want to move `out.add` outside `match`, this will reduce duplication.

node/src/main/scala/com/wavesplatform/network/RxExtensionLoader.scala
```diff
@@ -147,29 +170,143 @@ object RxExtensionLoader extends ScorexLogging {
...
+            if (Option(receivedSnapshots.getIfPresent(blockId)).isEmpty) {
+              pendingBlocks.put((ch, blockId), block)
+              ch.writeAndFlush(GetSnapshot(blockId))
+            }
+          }
           state
+      }
+    }
 
+    def onSnapshot(state: State, ch: Channel, snapshot: BlockSnapshot): State = {
```
vsuharnikov, line 218: This function shares a functionality with `onBlock`. I think we could make them simpler, if extract a common part (errors and changing the state) to a separate function. How can it be:
1. Create an updated instance of `LoaderState.ExpectingBlocksWithSnapshots` based on received data (block or snapshot), update structures and write all required data to channels
2. A common part: look into the updated `LoaderState.ExpectingBlocksWithSnapshots` and:
   1. If we received all blocks and snapshots - finish the loading of extension.
   2. If not: 
      1. Scheduler an error, that we didn't receive blocks and snapshots. We could use a common error message format, e.g.: `"Timeout loading blocks={} and snapshots={}"`. We also can limit the shown number of blocks and snapshots by 1 and add "..." if there are more blocks or snapshots. Actually, I think, just 1 not received block and 1 snapshot are enought: `"Timeout loading block={} and snapshot={}"`
      2. Use `state.withLoaderState` to return a new state.

node/src/main/scala/com/wavesplatform/network/TransactionPublisher.scala
```diff
@@ -20,6 +20,8 @@ object TransactionPublisher extends ScorexLogging {
 
   import Scheduler.Implicits.global
 
+  val NoOp: TransactionPublisher = { (_, _) => Future(TracedResult.wrapValue(true)) }
```
vsuharnikov:
```suggestion
  val NoOp: TransactionPublisher = { (_, _) => Future.successful(TracedResult.wrapValue(true)) }
```
😉

vsuharnikov (changes requested):

vsuharnikov (approved):

DrBlast (approved):

phearnot (approved):

xrtm000 (approved):

PR #3884, opened 2023-09-29
NODE-2596 Optimised active leases

xrtm000:


ivan-mashonskiy (approved):

vsuharnikov (approved):

node/src/main/scala/com/wavesplatform/api/common/CommonAccountsApi.scala
```diff
@@ -60,7 +53,8 @@ object CommonAccountsApi {
   def apply(
       compositeBlockchain: () => SnapshotBlockchain,
       rdb: RDB,
-      blockchain: Blockchain
+      blockchain: Blockchain,
+      leaseStatesAreStoredByAddress: Boolean
```
phearnot: unused?

phearnot (changes requested):
- Instead of copying lease details to sender and recipient accounts, store only lease IDs (just like for address transactions).
- Add a test which checks that cancelled leases are excluded from the response.
- Add a test to check if rollback works (for both active and cancelled leases).

node/src/main/scala/com/wavesplatform/api/common/CommonAccountsApi.scala
xrtm000: Removed

DrBlast (approved):

phearnot (approved):

PR #3890, opened 2023-10-09
NODE-2503 Better transaction asset payment errors

xrtm000:


node/src/test/scala/com/wavesplatform/state/diffs/ci/TransactionAssetChecksTest.scala
```diff
@@ -0,0 +1,61 @@
...
+      )
+      d.appendBlockE(transfer(asset = IssuedAsset(ByteStr.fill(32)(1)))) should produce(
+        "asset '4vJ9JU1bJJE96FWSJKvHsmmFADCg4gpZQff4P3bkLKi' is not found on the blockchain"
+      )
+      val transferIssued = transfer(asset = asset)
+      d.appendBlockE(transferIssued) should produce(s"leads to negative asset '$asset' balance")
+      d.appendBlock(transfer(secondSigner, defaultAddress, asset = asset))
+      d.appendAndAssertSucceed(transferIssued)
+    }
+  }
```
DrBlast, line 91: add the same tests for ethereum invoke and transfer

DrBlast (changes requested):

node/src/main/scala/com/wavesplatform/state/diffs/CommonValidation.scala
```diff
@@ -95,6 +100,13 @@ object CommonValidation {
 
           for {
             address <- blockchain.resolveAlias(citx.dApp)
+            _ <- citx.payments
+              .collectFirstSome {
+                case Payment(_, IssuedAsset(id)) => InvokeDiffsCommon.checkAsset(blockchain, id).swap.toOption
+                case Payment(_, Waves)           => None
+              }
```
vsuharnikov: Why just don't use `.collectFirst` which works the same way but with less code?
```suggestion
            _ <- citx.payments
              .collectFirst { 
                case Payment(_, IssuedAsset(id)) => InvokeDiffsCommon.checkAsset(blockchain, id).swap
              }
```

ivan-mashonskiy (approved):

node/src/main/scala/com/wavesplatform/state/diffs/CommonValidation.scala
xrtm000: Because it's not the same. 
`.collectFirstSome` allows here to collect error from `Left`, avoiding `Either` structure.

node/src/test/scala/com/wavesplatform/state/diffs/ci/TransactionAssetChecksTest.scala
xrtm000: Done

DrBlast (approved):

vsuharnikov (approved):

phearnot (approved):

PR #3896, opened 2023-10-16
NODE-2528 Typed errors replacing exceptions in RIDE sources

xrtm000:


ivan-mashonskiy (approved):

lang/shared/src/main/scala/com/wavesplatform/lang/v1/evaluator/ctx/impl/waves/Functions.scala
```diff
@@ -434,16 +441,24 @@ object Functions {
...
+                  r => env.accountBalanceOf(r, Some(assetId.arr)).map(_.map(CONST_LONG).leftMap(CommonError(_)))
+                )
             case xs =>
               notImplemented[F, EVALUATED](s"assetBalance(a: Address|Alias, u: ByteVector|Unit)", xs)
           }
       }
     }
 
-  val assetBalanceV4F: BaseFunction[Environment] =
+  def assetBalanceV4F(typedError: Boolean): BaseFunction[Environment] =
```
vsuharnikov: We couldn't add `typedError` argument everywhere, if we've had the `ThrownError` subclass of `ExecutionError`. E.g.:
1. Return `ThrownError` if we expect an exception. The class may contain this exception, so we don't lose a stacktrace.
2. Check in the end the `typedError` flag and throw the exception if it is needed.

lang/shared/src/main/scala/com/wavesplatform/lang/v1/evaluator/ctx/impl/waves/Functions.scala
```diff
@@ -434,16 +441,24 @@ object Functions {
         override def evaluate[F[_]: Monad](env: Environment[F], args: List[EVALUATED]): F[Either[ExecutionError, EVALUATED]] =
           args match {
             case (c: CaseObj) :: u :: Nil if u == unit =>
-              env.accountBalanceOf(caseObjToRecipient(c), None).map(_.map(CONST_LONG).leftMap(CommonError(_)))
+              caseObjToRecipient(c, typedError)
+                .fold(
+                  _.asLeft[EVALUATED].pure[F],
```
vsuharnikov: Should we make `caseObjToRecipient` return `F[Either[ExecutionError, Recipient]]` and handle this error? We convert the result to `F` anyway now.

xrtm000: `.fold` will become unavailable for `F[Either[*, *]]`

lang/shared/src/main/scala/com/wavesplatform/lang/v1/evaluator/ctx/impl/waves/Functions.scala
xrtm000: Done

vsuharnikov (approved):

DrBlast (approved):

phearnot (approved):

PR #3904, opened 2023-10-23
NODE-2618 Fixed GRPC getNFTList

xrtm000:


grpc-server/src/main/scala/com/wavesplatform/api/grpc/AssetsApiGrpcImpl.scala
```diff
@@ -35,7 +35,7 @@ class AssetsApiGrpcImpl(assetsApi: CommonAssetsApi, accountsApi: CommonAccountsA
           .concatMapIterable(_.map { case (a, d) =>
             NFTResponse(a.id.toByteString, Some(assetInfoResponse(d)))
           })
-          .take(request.limit)
+          .takeLast(request.limit)
```
phearnot: This does not seem right: it would always return the last `request.limit` entries.

phearnot (changes requested):
Please add the following test: given the list [N1, N2, N3, N4, N5], limit=2 and after=N2, the method should return [N3, N4]. It should also work across the liquid block boundaries (e.g. when N2 is in the DB, and N3..N5 are in the liquid block).

grpc-server/src/main/scala/com/wavesplatform/api/grpc/AssetsApiGrpcImpl.scala
xrtm000: Done

DrBlast (approved):

node/src/main/scala/com/wavesplatform/api/common/AddressPortfolio.scala
```diff
@@ -100,18 +99,20 @@ class BalanceIterator(
     address: Address,
     underlying: Iterator[Seq[(IssuedAsset, Long)]],
     includeAsset: IssuedAsset => Boolean,
-    private var pendingOverrides: Map[(Address, Asset), Long]
+    private var pendingOverrides: Map[(Address, Asset), Long],
+    maybeAfter: Option[IssuedAsset]
```
vsuharnikov: We don't use `includeAsset` and `maybeAfter` for solid blocks here, but we use them indirectly near `BalanceIterator`instances. So, two places for same filtering. We could do better.

There are two possible solutions those eliminate any questions here and make the code simpler:
a) Implement the filtering inside `BalanceIterator` for all values and eliminate a current external filtering. This could require `include: (Boolean, IssuedAsset) => Boolean`, where first `Boolean` is `IsInLiquidBlock`.
b)  Remove the filtering here and use only the external filtering.

I think, `b` is better, because it is simpler and we load asset descriptions twice in `nftIterator` now, but will load once after implementation of `b`.

node/src/main/scala/com/wavesplatform/api/common/AddressPortfolio.scala
```diff
@@ -100,18 +99,20 @@ class BalanceIterator(
     address: Address,
     underlying: Iterator[Seq[(IssuedAsset, Long)]],
     includeAsset: IssuedAsset => Boolean,
-    private var pendingOverrides: Map[(Address, Asset), Long]
+    private var pendingOverrides: Map[(Address, Asset), Long],
```
vsuharnikov: I suggest using `VectorMap` here, because when I look at this function, I see:
1) Strict ordering is not required.
2) We are iterating by entries of this map <-- looks like a bug.
3) The result returned to a user.

Also we are passing `VectorMap` anyway. So this is only a "documentation" moment.

vsuharnikov (changes requested):
I'm ready to discuss this:

node/src/main/scala/com/wavesplatform/api/common/AddressPortfolio.scala
xrtm000: Done

node/src/main/scala/com/wavesplatform/api/common/AddressPortfolio.scala
xrtm000: Done

vsuharnikov (approved):

phearnot (approved):

PR #3906, opened 2023-10-24
NODE-2616 Refactored validation of transaction version

xrtm000:


node/src/main/scala/com/wavesplatform/transaction/VersionedTransaction.scala
```diff
@@ -1,11 +1,19 @@
 package com.wavesplatform.transaction
 
 trait VersionedTransaction {
-  def version: TxVersion
+  val version: TxVersion
+  val maxVersion: TxVersion
```
phearnot: Moving `maxVersion` to `TransactionType` would better reflect the reality, since max version is a property of transaction type, rather than some particular transaction

phearnot (changes requested):

node/src/main/scala/com/wavesplatform/transaction/VersionedTransaction.scala
xrtm000: Refactored

node/src/main/scala/com/wavesplatform/database/package.scala
```diff
@@ -645,11 +637,11 @@ package object database {
...
-      case lps: PBSince if !lps.isProtobufVersion => TD.LegacyBytes(ByteString.copyFrom(tx.bytes()))
-      case _: GenesisTransaction                  => TD.LegacyBytes(ByteString.copyFrom(tx.bytes()))
-      case _: PaymentTransaction                  => TD.LegacyBytes(ByteString.copyFrom(tx.bytes()))
-      case et: EthereumTransaction                => TD.EthereumTransaction(ByteString.copyFrom(et.bytes()))
-      case _                                      => TD.WavesTransaction(PBTransactions.protobuf(tx))
+      case lps: PBSince with VersionedTransaction if !PBSince.affects(lps) => TD.LegacyBytes(ByteString.copyFrom(tx.bytes()))
+      case _: GenesisTransaction                                           => TD.LegacyBytes(ByteString.copyFrom(tx.bytes()))
+      case _: PaymentTransaction                                           => TD.LegacyBytes(ByteString.copyFrom(tx.bytes()))
+      case et: EthereumTransaction                                         => TD.EthereumTransaction(ByteString.copyFrom(et.bytes()))
+      case _                                                               => TD.WavesTransaction(PBTransactions.protobuf(tx))
```
vsuharnikov: Wondering, why not:
```suggestion
      case lps: PBSince with VersionedTransaction if PBSince.affects(lps) => TD.WavesTransaction(PBTransactions.protobuf(tx))
      case et: EthereumTransaction                                        => TD.EthereumTransaction(ByteString.copyFrom(et.bytes()))
      case _                                                              => TD.LegacyBytes(ByteString.copyFrom(tx.bytes()))
```
It looks simpler

vsuharnikov (approved):

node/src/main/scala/com/wavesplatform/database/package.scala
xrtm000: Done

ivan-mashonskiy (approved):

DrBlast (approved):

node/src/main/scala/com/wavesplatform/transaction/VersionedTransaction.scala
```diff
@@ -1,11 +1,13 @@
 package com.wavesplatform.transaction
 
-trait VersionedTransaction {
-  def version: TxVersion
+sealed trait VersionedTransaction {
```
phearnot:
```suggestion
sealed trait VersionedTransaction {
  def version: TxVersion
  protected def maxVersion: TxVersion
  def isVersionValid: Boolean = 0 < version && version <= maxVersion
}

object VersionedTransaction {
  trait OnlyV1 extends VersionedTransaction {
    override val version: TxVersion = TxVersion.V1
    override val maxVersion: TxVersion = TxVersion.V1
  }
  trait ToV2 extends VersionedTransaction {
    override val maxVersion: TxVersion = TxVersion.V2
  }
  trait ToV3 extends VersionedTransaction {
    override val maxVersion: TxVersion = TxVersion.V3
  }
}
```

phearnot (changes requested):
- `EthereumTransaction` should be neither `VersionedTransaction` nor `PBSince`

phearnot (approved):

PR #3909, opened 2023-10-27
NODE-2626 Corrected snapshot hash of LeaseState

xrtm000:


node/src/main/scala/com/wavesplatform/state/TxStateSnapshotHashBuilder.scala
```diff
@@ -75,7 +75,17 @@ object TxStateSnapshotHashBuilder {
     } addEntry(KeyType.AssetScript, asset.id.arr)(scriptInfo.script.bytes().arr)
 
     snapshot.leaseStates.foreach { case (leaseId, details) =>
-      addEntry(KeyType.LeaseStatus, leaseId.arr)(booleanToBytes(details.isActive))
+      if (details.isActive)
+        addEntry(KeyType.LeaseStatus, leaseId.arr)(
+          booleanToBytes(true),
+          details.sender.arr,
+          details.recipient.bytes,
+          Longs.toByteArray(details.amount)
```
vsuharnikov, line 84: I see the [different](https://github.com/wavesplatform/protobuf-schemas/pull/84/files#diff-6cdd8c2ddb6ac7af1d937fa548cfcb0bd558c151ce202aa3895deb78a85d1cffR42-R45) order of fields: amount, sender, recipient.
Should we maintain the order of fields during the hashing? It looks like we haven't. Personally, the consistency is a good thing. What do you think?

vsuharnikov (approved):

node/src/main/scala/com/wavesplatform/state/StateSnapshot.scala
```diff
@@ -172,23 +163,28 @@ object StateSnapshot {
...
+              height = 0
+            )
+          case _: TransactionStateSnapshot.LeaseState.Status.Cancelled | TransactionStateSnapshot.LeaseState.Status.Empty =>
+            ls.leaseId.toByteStr -> LeaseDetails(
+              PublicKey(ByteStr.fill(KeyLength)(0)),
+              Address(Array.fill(Address.HashLength)(0)),
+              0,
+              LeaseDetails.Status.Cancelled(0, None),
+              sourceId = ByteStr.empty,
+              height = 0
```
vsuharnikov: As @phearnot said, we can fill the data here from the snapshot

node/src/main/scala/com/wavesplatform/state/TxStateSnapshotHashBuilder.scala
xrtm000: I think these cases are from different contexts

goodmai (approved):

ivan-mashonskiy (approved):

node/src/main/scala/com/wavesplatform/state/StateSnapshot.scala
xrtm000: Done

phearnot (approved):

PR #3910, opened 2023-11-01
NODE-2625 Fixed GRPC Reflection API

xrtm000:


node-it/src/test/scala/com/wavesplatform/it/Docker.scala
```diff
@@ -263,7 +263,7 @@ class Docker(
       val containerConfig = ContainerConfig
         .builder()
         .image(imageName)
-        .exposedPorts(s"$internalDebuggerPort")
+        .exposedPorts(if (Try(nodeConfig.getStringList("waves.extensions").contains("com.wavesplatform.events.BlockchainUpdates")).getOrElse(false)) "6881" else s"$internalDebuggerPort")
```
phearnot: This completely replaces port mapping when BU is enabled, which means one will not be able to debug the node inside the container. Please change this to always expose debug port, and expose BU port as needed.

phearnot (changes requested):

node-it/src/test/scala/com/wavesplatform/it/Docker.scala
xrtm000: Done

DrBlast (approved):

phearnot (approved):

PR #3953, opened 2024-06-27
Boosted block reward

phearnot:


vsuharnikov (commented):
It seems, we can write:
```scala
blockchainUpdater.blockRewardBoost(height) // Returns BlockRewardCalculator.RewardBoost or 1
```
instead of:
```scala
if (blockchainUpdater.isBlockRewardBoostActive(height)) BlockRewardCalculator.RewardBoost else 1
```
because `isBlockRewardBoostActive` always comes with `if ... else ...`?

ivan-mashonskiy (approved):

DrBlast (approved):

PR #3966, opened 2024-09-16
Use seekForPrev in balanceAtHeight

phearnot:


node/src/main/scala/com/wavesplatform/database/RocksDBWriter.scala
```diff
@@ -1315,28 +1315,16 @@ class RocksDBWriter(
...
-        balance.height -> balance.balance
+    db.get(Keys.addressId(address)).flatMap { aid =>
+      val key = assetId match {
+        case Waves              => Keys.wavesBalanceAt(aid, Height(height))
+        case asset: IssuedAsset => Keys.assetBalanceAt(aid, asset, Height(height))
       }
+      Using(db.newIterator) { iter =>
+        iter.seekForPrev(key.keyBytes)
+        require(iter.isValid && iter.key().startsWith(key.keyBytes.dropRight(4)))
+        Ints.fromByteArray(iter.key().takeRight(4)) -> key.parse(iter.value()).balance
```
vladimirlogachev: Maybe `4` could be declared as `somethingLength`?
That would be easier to follow when reading the code.

vsuharnikov (approved):

ivan-mashonskiy (approved):

node/src/main/scala/com/wavesplatform/database/RocksDBWriter.scala
vladimirlogachev: And `dropRight` as well

vladimirlogachev (approved):

PR #3977, opened 2024-11-20
Added Units registry

phearnot:


DrBlast (approved):

node/src/main/scala/com/wavesplatform/settings/BlockchainSettings.scala
```diff
@@ -78,7 +78,8 @@ case class FunctionalitySettings(
     xtnBuybackRewardPeriod: Int = Int.MaxValue,
     lightNodeBlockFieldsAbsenceInterval: Int = 1000,
     blockRewardBoostPeriod: Int = 1000,
-    paymentsCheckHeight: Int = 0
+    paymentsCheckHeight: Int = 0,
+    unitsRegistryAddress: Option[String] = None
```
vladimirlogachev, line 82: I suggest to add `unitsRegistryAddress = None` to `val STAGENET` to make it more explicit.

vsuharnikov (approved):

PR #3978, opened 2024-11-28
Ability to disable time synchronization

vsuharnikov:
This is useful during testing in a local environment with multiple Waves nodes in Docker. The system time will be used with disabled NTP setting: `waves.ntp-server = null`


vladimirlogachev (approved):

phearnot:
Can be achieved with `waves.ntp-server = ""`

PR #4010, opened 2025-08-06
More central release related fixes

phearnot:
https://www.scala-sbt.org/1.x/docs/Using-Sonatype.html


.github/workflows/check-pr.yaml
```diff
@@ -82,7 +102,7 @@ jobs:
         run: sbt --batch ";node-it/docker;node-it/test"
         env:
           RUN_ID: ${{ github.ref_name }}-${{ github.run_number }}-${{ github.run_attempt }}
-      - uses: dorny/test-reporter@v2
+      - uses: dorny/test-reporter@dc3a92680fcc15842eef52e8c4606ea7ce6bd3f3
```
vsuharnikov, line 105: We use this multiple times. Can we add a global variable:
```yaml
env:
  TEST_REPORTER: "dorny/test-reporter@dc3a92680fcc15842eef52e8c4606ea7ce6bd3f3"
```
And then use it:
```yaml
      - uses: ${{ env.TEST_REPORTER }}
```
?

.github/workflows/check-pr.yaml
```diff
@@ -95,4 +115,27 @@ jobs:
...
+      - uses: actions/checkout@v4
+      - uses: actions/setup-java@v4
+        with:
+          distribution: 'temurin'
+          java-version: '11'
+          cache: 'sbt'
+          gpg-private-key: ${{ secrets.OSSRH_GPG_KEY }}
+          gpg-passphrase: ${{ secrets.OSSRH_GPG_PASSPHRASE }}
+      - uses: sbt/setup-sbt@v1
+      - run: sbt -Dproject.version=$(awk -F '[".]' '{print $3"."$4"."$5+1}' version.sbt)-$PR_NUMBER-SNAPSHOT --batch ';publish'
```
vsuharnikov:
```suggestion
      - run: sbt -Dproject.version=$(awk -F '[".]' '{print $3"."$4"."$5+1}' version.sbt)-$PR_NUMBER-SNAPSHOT --batch 'publish'
```
The heading ";" is not required in ";even;there;are;multiple;commands"

.github/workflows/publish-to-npmjs.yml
```diff
@@ -0,0 +1,61 @@
...
+      - uses: actions/checkout@v4
+      - uses: proudust/gh-describe@70f72d4f6304ea053cf5a3d71c36211d5acc0c73
+        id: ghd
+      - uses: actions/setup-java@v4
+        with:
+          distribution: 'temurin'
+          java-version: '11'
+          cache: 'sbt'
+      - uses: sbt/setup-sbt@v1
+      - run: sbt -Dproject.version=${{ steps.ghd.outputs.describe }} --batch ';lang-tests/test;lang-tests-js/test;repl/test;langJS/fullOptJS;replJS/fullOptJS'
```
vsuharnikov, line 40:
```suggestion
      - run: sbt -Dproject.version=${{ steps.ghd.outputs.describe }} --batch 'lang-tests/test;lang-tests-js/test;repl/test;langJS/fullOptJS;replJS/fullOptJS'
```

.github/workflows/check-pr.yaml
phearnot: Environment variables are evaluated when actions are already cloned, so you can't use environment variables in `uses:`. You could probably use yaml-specific tricks, but I don't think it's worth it.

vsuharnikov (approved):

PR #4014, opened 2025-10-27
Removed supertagged

phearnot:


node/src/main/scala/com/wavesplatform/database/RDB.scala
```diff
@@ -43,12 +43,15 @@ object RDB extends StrictLogging {
...
 
-    val handles             = new util.ArrayList[ColumnFamilyHandle]()
-    val defaultCfOptions    = newColumnFamilyOptions(12.0, 16 << 10, settings.rocksdb.mainCacheSize, 0.6, settings.rocksdb.writeBufferSize)
-    val txMetaCfOptions     = newColumnFamilyOptions(10.0, 2 << 10, settings.rocksdb.txMetaCacheSize, 0.9, settings.rocksdb.writeBufferSize)
-    val txCfOptions         = newColumnFamilyOptions(10.0, 2 << 10, settings.rocksdb.txCacheSize, 0.9, settings.rocksdb.writeBufferSize)
-    val txSnapshotCfOptions = newColumnFamilyOptions(10.0, 2 << 10, settings.rocksdb.txSnapshotCacheSize, 0.9, settings.rocksdb.writeBufferSize)
-    val apiCfOptions        = newColumnFamilyOptions(10.0, 2 << 10, settings.rocksdb.apiCacheSize, 0.9, settings.rocksdb.writeBufferSize)
+    val handles = new util.ArrayList[ColumnFamilyHandle]()
+    val defaultCfOptions =
+      newColumnFamilyOptions(12.0, 16 << 10, settings.rocksdb.mainCacheSize.longValue, 0.6, settings.rocksdb.writeBufferSize.longValue)
```
vsuharnikov: What if `newColumnFamilyOptions` accepts `SizeInBytes` instead of `Long`?

node/src/main/scala/com/wavesplatform/database/KeyHelpers.scala
```diff
@@ -8,16 +8,20 @@ import org.rocksdb.ColumnFamilyHandle
...
+    ByteBuffer.allocate(4 + bytes.length).put(bytes).putInt(n).array()
 
-  def hNum(height: Int, num: TxNum): Array[Byte] = Bytes.concat(Ints.toByteArray(height), Shorts.toByteArray(num))
+  def hAddr(height: Height, addressId: AddressId): Array[Byte] = hBytes(addressId.toByteArray, height)
 
-  def historyKey(keyTag: KeyTag, suffix: Array[Byte]): Key[Seq[Int]] = Key(keyTag, suffix, readIntSeq, writeIntSeq)
+  def hNum(height: Height, num: TxNum): Array[Byte] = Bytes.concat(Ints.toByteArray(height.toInt), Shorts.toByteArray(num.toShort))
+
+  def historyKey(keyTag: KeyTag, suffix: Array[Byte]): Key[Seq[Height]] =
+    Key(keyTag, suffix, bs => readIntSeq(bs).map(Height.apply), hs => writeIntSeq(hs.map(_.toInt)))
```
vsuharnikov: If we would have:
```scala
object Height {
   def apply(heights: Seq[Int]): Seq[Height] = heights
```
then we don't need an iteration.

vladimirlogachev (approved):

vsuharnikov (approved):

vsuharnikov (approved):

PR #4016, opened 2025-11-17
Add finalization voting to the microblock signature

vladimirlogachev:


node/src/main/scala/com/wavesplatform/block/serialization/MicroBlockSerializer.scala
```diff
@@ -39,7 +46,13 @@ object MicroBlockSerializer {
...
-
-      MicroBlock(version, generator, transactionData, reference, totalResBlockSig, signature, stateHash, finalizationVoting = None)
+      val finalizationVoting =
+        if (buf.hasRemaining) {
+          val serializedVoting = buf.getByteArray(buf.remaining())
+          val pbVoting =
+            PBUtils.decode(serializedVoting, PBFinalizationVoting).fold(throw _, identity)
+          Some(PBFinalizationVotings.vanilla(pbVoting).get)
+        } else None
+      MicroBlock(version, generator, transactionData, reference, totalResBlockSig, signature, stateHash, finalizationVoting)
```
vsuharnikov: It seems we don't need this, because this parsing is a legacy, we use protobuf now
```suggestion
      // We don't use this parsing since protobuf, no new bytes parsing is needed

      MicroBlock(version, generator, transactionData, reference, totalResBlockSig, signature, stateHash, finalizationVoting = None)
```

vsuharnikov (changes requested):

vsuharnikov (approved):

PR #4018, opened 2025-11-21
Fix statehash api

vladimirlogachev:


node/src/main/scala/com/wavesplatform/database/Caches.scala
```diff
@@ -458,7 +460,7 @@ abstract class Caches extends Blockchain, Storage {
       snapshot.accountScriptsByAddress.map { case (address, s) => addressIdWithFallback(address, newAddressIds) -> s },
       newFinalizedHeight,
       updatedCurrentGeneratorBalances,
-      nextCommittedGenerators,
+      if deterministicFinalityActivated then Some(nextCommittedGenerators) else None,
```
vsuharnikov: We don't need Option here, because `deterministicFinalityActivated` only if `this.generationPeriodOf(h)` non empty. See the next comment.

node/src/main/scala/com/wavesplatform/database/RocksDBWriter.scala
```diff
@@ -744,13 +744,16 @@ class RocksDBWriter(
       }
 
       this.generationPeriodOf(h).foreach { currPeriod => // None checked in Caches
```
vsuharnikov, line 746: Here we checking that feature is activated. There is no "generation period" before the feature activation. See also `GenerationPeriod.from`

node/src/main/scala/com/wavesplatform/api/http/DebugApiRoute.scala
```diff
@@ -258,13 +259,21 @@ case class DebugApiRoute(
...
-    } yield Json.toJson(sh).as[JsObject] ++ Json.obj(
-      "snapshotHash" -> db.snapshotStateHash(height),
-      "blockId"      -> h.id().toString,
-      "baseTarget"   -> h.header.baseTarget,
-      "height"       -> height,
-      "version"      -> Version.VersionString
-    )
+    } yield {
+      val stateHashJson = Json.toJson(sh).as[JsObject]
+      println(s"stateHashJson: ${sh}")
```
vsuharnikov: Debugging stuff

node/src/main/scala/com/wavesplatform/database/Caches.scala
```diff
@@ -442,7 +443,8 @@ abstract class Caches extends Blockchain, Storage {
     for (leaseId <- snapshot.cancelledLeases.keys) stateHash.addLeaseStatus(leaseId, isActive = false)
     for ((assetId, sponsorship) <- snapshot.sponsorships) stateHash.addSponsorship(assetId, sponsorship.minFee)
     for ((alias, address) <- snapshot.aliases) stateHash.addAlias(address, alias.name)
-    snapshot.nextCommittedGenerators.foreach(stateHash.addNextCommittedGenerator)
+    if (deterministicFinalityActivated)
+      snapshot.nextCommittedGenerators.foreach(stateHash.addNextCommittedGenerator)
```
vsuharnikov: This is empty if the feature is not activated, because `CommitToGeneration` transaction is not allowed. No need for additional checks.

vsuharnikov (changes requested):

node/src/main/scala/com/wavesplatform/api/http/DebugApiRoute.scala
vladimirlogachev: Thank you, fixed

node/src/main/scala/com/wavesplatform/database/RocksDBWriter.scala
vladimirlogachev: Thank you, fixed

node/src/main/scala/com/wavesplatform/database/Caches.scala
vladimirlogachev: Thank you, fixed

node/src/main/scala/com/wavesplatform/database/Caches.scala
vladimirlogachev: Thank you, fixed

PR #4022, opened 2025-11-28
Add committed generator balances to the StateHash

vladimirlogachev:


node/src/main/scala/com/wavesplatform/state/StateHashBuilder.scala
```diff
@@ -40,7 +40,7 @@ class StateHashBuilder {
   private val maps = Vector.fill(SectionId.maxId)(mutable.TreeMap.empty[ByteStr, Array[Byte]])
 
   private def addEntry(section: SectionId.Value, key: Array[Byte]*)(value: Array[Byte]*): Unit = {
-    val solidKey   = ByteStr(key.reduce(_ ++ _))
+    val solidKey   = ByteStr(if (key.isEmpty) Array.emptyByteArray else key.reduce(_ ++ _))
```
vsuharnikov:
```suggestion
    val solidKey   = ByteStr(key.foldLeft(Array.emptyByteArray)(_ ++ _))
```

PR #4045, opened 2026-06-04
build: add waves-ext project dependency to core build

D-H-O-R-A:
This change integrates the `waves-ext` project dependency directly into the root Waves platform SBT build. 

Key improvements:
- Declared the `waves-ext` lazy project, establishing explicit dependencies on the `node` (compile/runtime) and `grpc-server` modules.
- Added `waves-ext` to the root project's aggregate list, ensuring it compiles as part of the main Waves build and packaging flow.
- Configured proper source generators for Protobuf (`scalapb.gen`) under `waves-ext` to resolve the required schemas dynamically.
- Fixes classloader and linkage issues when loading custom dex extension modules alongside the node JVM in custom network environments.

Developer Contact:
- Email: diegoantunes2301@gmail.com
- WhatsApp: +5511974289097

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -104,6 +104,22 @@ lazy val `node-tests` = project
 lazy val `grpc-server` =
   project.dependsOn(node % "compile;runtime->provided", `node-testkit` % "test")
 
+lazy val `waves-ext` = project
+  .in(file("waves-ext"))
+  .dependsOn(node % "compile;runtime->provided", `grpc-server` % "compile->compile", `node-testkit` % "test")
+  .settings(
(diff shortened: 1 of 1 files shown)


phearnot:
Thanks for your contribution.
What would be the point of a placeholder project?

D-H-O-R-A:
It's not a placeholder. waves-ext is the DEX extension module — the full implementation is in wavesplatform/matcher#761.

Adding it to the core build ensures it compiles against the same node internals and gets included in the packaging flow, avoiding binary incompatibilities at runtime.

Check the matcher PR for context.

